PR #6, opened 2018-04-03
Pin requirements.txt

tunnell:
If you pin versions, it will make things in the long run easier for reproducibility.  Also, we make sure that we are testing the same versions that we are deploying.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
-numpy
-pandas
-numba
-blosc
-zstd
-tqdm
-dill
+numpy==1.14.2
+pandas==0.22.0
+numba==0.37.0
+blosc==1.5.1
+zstd==1.3.4
+tqdm==4.20.0
+dill==0.2.7.1


JelleAalbers:
Thanks! While I was against this for pax in the past, it makes sense. We've had a few issues when new numpy and numba versions came at one point. So let's pin.

However we should probably stop filling install_requires from requirements.txt after pinning; see e.g.

* https://packaging.python.org/discussions/install-requires-vs-requirements/
* https://caremad.io/posts/2013/07/setup-vs-requirement/
* https://github.com/pyupio/pyup/issues/137

The best practice seems to be to have pinned dependencies in requirements.txt and unpinned ones in install_requires. This way you can have consistent deployments with `pip install -r requirements.txt` while  `pip install` remains flexible, e.g. for people who want to use part of strax in their projects.

Moreover, this allows you to pin dependencies of your dependencies in requirements.txt; essentially requirements.txt can then become just the `pip freeze` output.

tunnell:
@JelleAalbers I'll look at this install requires thing...

JelleAalbers:
Cool, I just did https://github.com/JelleAalbers/strax/commit/718456dd16f4d43199721a65037f2f56ea425d0c now, but the repetition feels a bit wrong.

PR #7, opened 2018-04-04
Avoid repeating requirements

tunnell:
Instead of repeating the requirements, you can just read requirements.txt then strip pinned versions.  See #6

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,5 +17,5 @@ setup(name='strax',
       url='https://github.com/jelleaalbers/strax',
       setup_requires=['pytest-runner'],
       install_requires=req_file('requirements.txt'),
-      tests_require=requires + ['pytest', 'boltons', 'hypothesis'],
+      tests_require=req_file('requirements.txt') + req_file('requirements_tests.txt'),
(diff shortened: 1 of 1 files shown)


JelleAalbers:
I think you omitted the part that actually strips the pinned versions from the strings. Also `requires` was used twice but you replaced only one of them, so it doesn't work yet.

This is certainly a solution, but I recall reading some warnings not to parse requirement.txt yourself since it apparently can do a lot of things. Did you find somebody that recommended this?

tunnell:
I found it in one of the links you included.  The requirements file isn't so complicated, but your call.

tunnell:
This is ready if you want it

JelleAalbers:
Thanks, this seems better than repeating the requirements, especially given that we're still in rapid development. 

I think you took https://github.com/pyupio/pyup/issues/137#issuecomment-323789520; from the discussion in that thread I gather this isn't exactly the standard way. On the other hand I'm not ready to go deeper in the woods with packaging at the moment... let's start with this and see when we need more.

PR #8, opened 2018-04-05
Better TravisCI settings

tunnell:
This is me improving the TravisCI settings in the following ways:

1.  Documentation
2. Adding OSX testing
3. Using miniconda for providing finer control over Python versions
4. Using matrix builds to allow finer testing in the future (Python versions, numba versions, OS, etc)

miniconda is not required to install strax, but makes testing a lot easier.  You will want to squash merge this since it took 10s of tries to get it actually working on TravisCI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,13 +12,6 @@ cache: pip
 # Everything should work in userspace
 sudo: false
 
-# Python versions to test
-python:
-  - "3.6"
-  - "3.6-dev"  # 3.6 development branch
-  - "3.7-dev"  # 3.7 development branch
-  - "3.8-dev"  # 3.8 development branch
-
 matrix:
   - include:
     - python: "3.6"
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,29 @@ sudo: false
 matrix:
   - include:
     - python: "3.6"
+      os: linux
     - python: "3.7-dev"
+      os: linux
     - python: "3.8-dev"
+      os: linux
+    
+    - python: "3.6"
+      os: osx
+    - python: "3.7-dev"
+      os: osx
+    - python: "3.8-dev"
+      os: osx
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Cool, thanks! The documentation and mac builds are certainly useful. Perhaps we should allow the mac builds to fail, or I'll be tempted to add windows builds too ;-)

Can you explain the advantage of miniconda? 'Finer control of python versions' is a bit unclear; it does take quite a few lines to setup which will have to be maintained, and it's useful to test that strax just works on vanilla python + pip. Ideally we'd test both miniconda and vanilla python maybe?

tunnell:
There's a wider range of Python versions in miniconda than TravisCI.  Also, TravisCI doesn't have Python3.6 for Mac.  Therefore, we can have a dedicated test on 3.6.5 (if we use that on Midway), have tests on 3.7 to know when we could move there, etc.  Once multiple versions of Python are supported, same sort of thing.  TravisCI also doesn't let you specify patch releases.

We can just set the Mac build to be allowed to fail once it starts failing.  We could also add a Windows build.  This is what we did on pax.

Using miniconda does add lines of code, which I was hesitant about due to the extra code.  Fortunately, pax and numba also have similar lines of code that we can copy from.  

If you're wanting to do a matrix build where you test Python 3.6.5, 3.7 v.s. Linux/Mac v.s. numba versions v.s. numpy versions, you'll need a finer level of control than you get through the requirements file.  The requirements file is fixed for every TravisCI build.  Therefore, this way, if you want to support multiple versions of packages for backward compatibility or whatever (issue that came up with pax), you can specify this when setting up the conda environment through environmental variables.  The requirements file doesn't allow this.

Therefore, it's just longer term planning even though it's not needed right now.

JelleAalbers:
OK, so miniconda has more and more specific python versions available, and we'd likely have to add more complexity anyway to test different package versions in the future.

PR #11, opened 2018-04-06
Bumpversion and PyPI through Travis

tunnell:
Having strax installable by PyPI means that we can begin building tools upon it and solve the dependency nightmare that we have in pax through Anaconda.

In this PR, there are a few things:

1.  bumpversion for version tracking, which tags releases.
2. We can now do `pip install strax`, where I need to give you the password.  
3. Upload automatically updates PyPI upon tags

The only caveat is that I need you to run a command to determine the encrypted password since TravisCI runs in your userspace at the moment.  

For encryption, see: https://docs.travis-ci.com/user/encryption-keys/

Note: this is a branch from #8 so that has to be merged first.  Also, squash merge since it's TravisCI, which means there are a lot of test commits.


JelleAalbers:
I had no idea you could even do this, this is a lot more convenient than keeping around .pypirc files or whatever it was. Thanks. Once the core works and we have a bit of docs we could make a first pre-release.

PR #14, opened 2018-05-06
Gitter support badge on README

tunnell:
This badge links to a Gitter room (I guess this makes sense over Slack, even though Slack is better?)

@JelleAalbers is this premature?


tunnell:
Wonder why TravisCI is so slow today...

PR #16, opened 2018-05-06
Testing TravisCI PyPI deploy

tunnell:
Testing to see if this encrypted key works.


tunnell:
The plan here is to merge this, which deploys to PyPI each commit. If it deploys, then I'm going to make it once again only deploy on tags.  This is messy, but TravisCI stuff is.

PR #19, opened 2018-05-09
Fix installation settings preventing numba cache

### LINKED ISSUE #12: Bug: cannot import strax
There's some issue with caching in numba?  The versions of all packages are as pinned in the requirements.txt.  When I do 'import strax', I get:

> RuntimeError: cannot cache function 'sort_by_time': no locator available for file '/Users/tunnell/Work/anaconda/envs/strax/lib/python3.6/site-packages/strax-0.0.1-py3.6.egg/strax/utils.py'

See the full output in https://gist.github.com/tunnell/594018c76e83f751bd8612fa22c099f2

Here is the result of `pip freeze` https://gist.github.com/tunnell/e4a6e1590a3edd9c604c52b4126119fe

tunnell:
The package was never tested with "python setup.py install".  Therefore, the submodules of strax weren't actually installed within the egg package.  Furthermore, we cannot zip the installation egg since numba has to be able to find the source code.

This fixes and closes #12.


tunnell:
Note the zip_safe.

PR #24, opened 2018-05-11
ripa and PyPI documentation

tunnell:
This PR has a few small things:

- ripa: in the Travis install, we use ripa.  This has two advantages:
    - This is particularly useful with pyup as new packages are on PyPI but not Anaconda
    - Speeds up the building for TravisCI continuous integration since packages such as llvm-lite can be downloaded in binary form
- Update references in setup.py to the older repository
- Add the HISTORY information to PyPI.


tunnell:
Squash merge.

PR #46, opened 2018-07-10
Split storage classes into frontends and backends, minor fixes

JelleAalbers:
This splits storage classes into frontends, which deal with finding data, and backends, which deal with the actual saving and loading. 

Frontends (such as a runs database) register data locations and which backend (files, databases, etc) was used to store them. They translate requests from the Context ('gimme event_basics with these config options') to requests for specific directories or database collections. Backends just serve these directories/collections without knowing anything about what they represent. Frontends can have one or more backends, and different frontends can make use of the same kind of backend.

This also adds:
  * A critical bug fix in ParallelSourcePlugin. The eventbuilder will not run without this  - specifically the DAQReader, so 1T data conversion is not affected. No idea how this got omitted in the last release.
  * Preparation for fuzzy matching options for getting data. These will enable you to load data or metadata even if not all config options / plugin versions match those set in the current context. This will require a few changes in the Context class too, so it is not functional yet.
  * A small description of the different storage classes in the docs.
  * A test for the file-based storage classes in test_core. This should help our coverage improve.
  * A few minor fixes for the fake DAQ, and a few extra options for the fake DAQ and 'eventbuilder'. 

The DataDirectory and ZipDirectory classes have been refactored to support this change, the MongoDB storage has been deleted for now.  Maybe I'll get to it later.


JelleAalbers:
@tunnell do you still want to review this first? @coderdj is working on the DAQ output, but without this PR he can't really test his code. It would be fine with me if you could look at it some later time.

tunnell:
I don't want to block any work.  It's hard for me to do the code review without a little bit of hand holding.  I'll try to look some this weekend, but as an alternative option: are you fine doing a screen share code review with me and I'd ask @jpivarski on Monday?  Cap it at 40 minutes screen share on Vidyo, collect notes for last two minutes in this PR, day to interate, then merge?

I don't want to block Dan, but this also feels fundamental.

tunnell:
I read through the doc and it seems like the right thing to do.  I guess it lets us to OSG/S3/Ceph object store where object is table, files (HDF5, npz, whatever), or pure NoSQL DB (Mongo/Big Table).   I'm guessing strax itself won't decide where it gets data from (try with priority local data) since that's more snax.  

It also provides a way to test within a backend if the data is present there.

JelleAalbers:
Sounds good, Monday afternoon would be fine, let me know when. I should not have put the essential fixes together with the big feature overhaul in one big PR...

Yes, this would allow you to implement fancy data stores as backends without also implementing a new runsdb-like system for each of them. You can do all the tracking in one or more separate databases (e.g. the xenon runs db or rucio).

Deciding where the data comes from and goes to is the job of a strax frontend. So I'm guessing that some of what you put in snax could be written as a strax frontend, though I haven't really looked at snax. Maybe that should be my homework :-) In any case such frontends would access experiment-specific DBs so they don't go into the strax repo. I only included a generic json-based frontend so strax has batteries included.

coderdj:
Take your time reviewing, you're not blocking DAQ development. Already got master reading DAQ output, which was easy with the nice examples already present.

strax/storage/common.py
```diff
@@ -25,130 +27,259 @@ def __repr__(self):
...
+                                         ignore_config=ignore_config)
+        return self._get_backend(backend).loader(backend_key, executor)
+
+    def saver(self, key, metadata, meta_only):
+        """Return saver for data described by DataKey."""
+        backend, backend_key = self.find(key, write=True)
+        return self._get_backend(backend).saver(backend_key,
+                                                metadata, meta_only)
+
+    def get_metadata(self, key,
```
tunnell: What type of metadata is there?  Is this explained somewhere?

tunnell: Ah, this is datatype metadata and NOT run metadata.  Can we clarify?  Also, datatype means table, which is slightly confusing.

JelleAalbers:
Thanks for your comments Chris! I've added docstrings to the places you indicated, and also an `is_stored` method of Context (+test), which you can use for checking if a piece of data has been saved through any registered front-end.

The name for a piece of data is currently "data type" in many places in strax. We should probably decide on a single name, e.g. array, table, data piece, or something along those lines. It is not going to be minitree, plank, or anything else with plantlike connections :-)

Issue #39, opened 2018-07-04
Processing whole 1T run fails


tunnell:
First, maybe need xe-strax repo for these issues.  Second, trigger monitor confusing things.

```
==> strax_47254785_err.log <==
    st.make(dataset, 'event_info', max_workers=8)
  File "/project/lgrandi/anaconda3/envs/strax/lib/python3.6/site-packages/strax-0.1.2-py3.6.egg/strax/core.py", line 459, in make
    for _ in self.get_iter(run_id, targets, save, max_workers, **kwargs):
  File "/project/lgrandi/anaconda3/envs/strax/lib/python3.6/site-packages/strax-0.1.2-py3.6.egg/strax/core.py", line 446, in get_iter
    components, max_workers=max_workers).iter():
  File "/project/lgrandi/anaconda3/envs/strax/lib/python3.6/site-packages/strax-0.1.2-py3.6.egg/strax/processor.py", line 85, in iter
    yield from final_generator
  File "/project/lgrandi/anaconda3/envs/strax/lib/python3.6/site-packages/strax-0.1.2-py3.6.egg/strax/mailbox.py", line 254, in _read
    raise MailboxKilled(self.killed_because)
strax.mailbox.MailboxKilled: ValueError('Invalid file name: /project2/lgrandi/tunnell/tempzi86l5e5/180206_0549/trigger_monitor_data.zip. Should be tpcname-something-firstevent-lastevent-nevents.zip') in build:raw_records
```

I just delete it manually for now.

Issue #37, opened 2018-07-03
Nondeterministic test failure for test_split_by_containment


JelleAalbers:
See e.g. https://travis-ci.org/AxFoundation/strax/jobs/399538399

tunnell:
Is there a way to set the random seed?

tunnell:
@JelleAalbers can we set random seed?

PR #47, opened 2018-07-10
Pax conversion: load only true zip files

JelleAalbers:
This should fix the pax conversion error #39 due to the presence of trigger monitor data zips. Not tested. Only works for TPC names that start with XENON.


JelleAalbers:
Travis build failure was the known recurrent issue #37

PR #64, opened 2018-08-20
_close should not require implementation

tunnell:
"raise NotImplementedError" is a bit much.  Not all backends require closing if they are atomic.  Therefore, I think we should be able to skip implementing this.  I changed raise `NotImplementedError` to `pass`.

### DIFF
--- a/strax/storage/common.py
+++ b/strax/storage/common.py
@@ -391,4 +391,4 @@ class Saver:
         raise NotImplementedError
 
     def _close(self):
-        raise NotImplementedError
+        pass


tunnell:
Maybe this is needed since Saver is parallel.

JelleAalbers:
Indeed only some backends need a complicated close, e.g. the `FileSaver` has to collect all the individual chunk metadata jsons. But all backends have to do something: at least another write of the metadata (`self.md`) to ensure the `writing_ended` field (and possibly `exception`) is written. 

This should probably be indicated in a comment or the docstring of the abstract method.

PR #71, opened 2018-08-22
Redo S3 schema (use one Bucket due to limited number of buckets

tunnell:
Same thing as before but with a new schema.  Instead of each bucket being a strax key, all objects live in one bucket.  Instead of buckets, I use 'folders' (i.e. key is strax_key/chunk_id).  This seems to be the right way to go.  Once this merges and goes into a release, I'll start using this as the main backend at the moment for XENON data at least.

Also, factor out JSON uploader to S3.

### DIFF
--- a/strax/storage/s3.py
+++ b/strax/storage/s3.py
@@ -150,7 +150,7 @@ class S3Backend(strax.StorageBackend):
                                    compressor=compressor)
 
     def _saver(self, key, metadata, meta_only=False):
-        return S3Saver(dirname,
+        return S3Saver(key,
                        metadata=metadata,
                        meta_only=meta_only,
                        **self.kwargs)
--- a/strax/storage/s3.py
+++ b/strax/storage/s3.py
@@ -1,7 +1,7 @@
 """I/O that speaks the S3 protocol
 
 The S3 protocol is an HTTP-based protocol spoken by Amazon Web Services, but
-also storage systems such as CEPH (used in particle physics).  Therefore,
+also storage systems such as Ceph (used in particle physics).  Therefore,
 this can be widely used if you know the appropriate endpoint.
 
 Be aware that you must specify the following two environmental variables or
@@ -22,6 +22,10 @@ import strax
 from strax import StorageFrontend
 
 export, __all__ = strax.exporter()
+
+# Track versions of S3 interface
+VERSION = 1
+BUCKET_NAME = 'strax_s3_v%d' % VERSION
 
 
 @export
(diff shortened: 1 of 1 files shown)


tunnell:
@JelleAalbers hope we can make release soon (related to processing).  Also, I feel like I am using the backend key wrong since I don't really use it.  For next update, maybe `BUCKET_NAME` should be the backend_key?  What was the reasoning here.

strax/storage/s3.py
```diff
@@ -63,10 +67,21 @@ def __init__(self,
...
-                                     service_name='s3')
+        self.s3 = boto3.client(aws_access_key_id=aws_access_key_id,
+                               aws_secret_access_key=aws_secret_access_key,
+                               endpoint_url=endpoint_url,
+                               service_name='s3')
+
+        # List of all buckets on this S3 store (finally, a bucket list!)
+        # then determine if either no buckets exist or (if they exist)
+        # whether or not the bucket we need exists.  If not, create.
+        # bucket_list = self.s3.list_buckets()
```
JelleAalbers: Is this commented out code still needed?

strax/storage/s3.py
```diff
@@ -78,44 +93,33 @@ def _find(self, key, write, fuzzy_for, fuzzy_for_options):
...
         bk = self.backend_key(key_str)
 
-        # Get list of all buckets / (run_id, lineage)
-        objects_list = self.s3client.list_buckets()
-
-        # If no buckets exist...
-        if 'Buckets' not in objects_list:
-            # No objects yet... so no metadata
+        # See if any objects exist for this key
+        objects_list = self.s3.list_objects(Bucket=BUCKET_NAME,
```
JelleAalbers: Nice! So we don't list the entire universe anymore.

strax/storage/s3.py
```diff
@@ -129,27 +133,34 @@ def __init__(self, **kwargs):
...
+        # Before returning dictionary
         return json.loads(text)
 
     def _read_chunk(self, backend_key, chunk_info, dtype, compressor):
         with tempfile.SpooledTemporaryFile() as f:
-            self.s3client.download_fileobj(Bucket=backend_key,
-                                           Key=chunk_info['filename'],
-                                           Fileobj=f)
+            self.s3.download_fileobj(Bucket=BUCKET_NAME,
+                                     Key=chunk_info['key_name'],
```
JelleAalbers: OK, so you added a new field `key_name` to the metadata of every chunk, to communicate the strax key to your reach_chunk function (because the storage backend base class first reads the metadata). But you already get the strax key in this function with backend_key -- in your implementation they are exactly the same. So no need to save it again for every chunk.

strax/storage/s3.py
```diff
@@ -162,56 +173,65 @@ class S3Saver(strax.Saver):
...
 
-        return dict(filename=filename, filesize=filesize)
+        return dict(key_name=key_name,
+                    filesize=filesize)
 
     def _save_chunk_metadata(self, chunk_info):
         if self.meta_only:
             # TODO HACK!
-            chunk_info["filename"] = '%06d' % chunk_info['chunk_i']
+            chunk_info["key_name"] = f"{self.strax_unique_key}/{chunk_info['chunk_i']:06d}"
```
JelleAalbers: So does this field only get added in the meta_only mode? Meta only saving is a rather hacky thing that may be removed soon. But as I mentioned above, maybe I don't understand what you're doing with key_name at the moment.

strax/storage/s3.py
tunnell: Yup

strax/storage/s3.py
tunnell: The key name also includes the chunk though.  So key_name is key/chunk

strax/storage/s3.py
tunnell: Fixed

strax/storage/s3.py
JelleAalbers: OK, so you could do `key = f"{backend_key}/{chunk_info['chunk_i']:06d}"` instead of saving this field in every chunk metadata.

tunnell:
What is the purpose of key_name?  Can I just set it to be the bucket name (i.e., strax_s3_v1) for now?  This then removes the global variable.  It at least runs :)  This might be a nice thing to look at during F2F.  I fixed some small things.  Can we merge then discuss at F2F the key part?

strax/storage/s3.py
```diff
@@ -162,56 +165,65 @@ class S3Saver(strax.Saver):
...
                                        data=data,
                                        compressor=self.md['compressor'])
             f.seek(0)
-            self.s3client.upload_fileobj(f, self.key, filename)
+            self.s3.upload_fileobj(f,
+                                   BUCKET_NAME,
+                                   key_name)
 
-        return dict(filename=filename, filesize=filesize)
+        return dict(key_name=key_name,
```
JelleAalbers: OK, this is where key_name gets set. Got it. So it's like filename except that it currently has the strax key attached in front of it.

PR #74, opened 2018-08-25
Resolve double boto requirement

tunnell:
setup.py modules have to be installed with pip.  However, our testing uses setup.py.

This is a test of TravisCI to see if I can figure out how to reconcile this.


tunnell:
This can be merged whenever.

JelleAalbers:
OK, you took it out of the xenon requirements instead, we'll leave S3 support in main strax since it might be useful for others later.

Issue #69, opened 2018-08-21
Key naming


tunnell:
Currently, the unique key for data is:

`170324_1748_event_basics_46c6d5c781c13c76a6d3a171d959f2673b4e8e1a`

However, the underscores are used both in the run name and also the key.  Can we switch to `-` to join these so they split() better? i.e.:

`170324_1748-event_basics-46c6d5c781c13c76a6d3a171d959f2673b4e8e1a`

JelleAalbers:
Good idea; we could also use colons or other separators, but underscores are too overloaded. In general strax's requirements on the run_id are a bit weird / to strict at the moment, it's supposed to be a string that can be converted to an int..

tunnell:
Done

PR #76, opened 2018-08-26
Change Delimiter to -

tunnell:
Solves #69 of underscore degeneracy.


tunnell:
Warning: merging this will break all currently processed data.  Do not merge until we start a new processing.  Ideally, we should tag a release without this *then* merge this for next release.

tunnell:
@JelleAalbers can you do a release then we merge this?

JelleAalbers:
Unfortunately this will also make all currently existing file-based data unreadable. Perhaps we should add some logic to temporarily try _-joining for building filenames too, if --joining fails?

tunnell:
Mainly issue for test data.  People stick with current version.  I bumped S3 version so won't see old data.

JelleAalbers:
We still have to update the test data

PR #82, opened 2018-08-29
fix negative drift time

XeBoris:
Simple main s1/s2 pairing: The main s1 (largest s1) have to be before main s2 (largest s2)

### DIFF
--- a/strax/xenon/plugins.py
+++ b/strax/xenon/plugins.py
@@ -460,9 +460,23 @@ class EventBasics(strax.LoopPlugin):
         return dtype
 
     def compute_loop(self, event, peaks):
+        __version__ = '0.1.0'
         result = dict(n_peaks=len(peaks))
         if not len(peaks):
             return result
+
+        #identify largest s2 properties
+        s2_values = None
+        s2_largest = None
+        s2_largest_ind = None
+        s2_largest_time = None
+        if 2 in peaks['type']:
+            s_mask = peaks['type'] == 2
+            ss = peaks[s_mask]
+            s2_values = ss['area']
+            s2_largest = np.max(ss['area'])
+            s2_largest_ind = np.argmax(ss['area'])
+            s2_largest_time = ss['time'][s2_largest_ind]
 
         main_s = dict()
         for s_i in [1, 2]:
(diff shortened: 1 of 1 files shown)


strax/xenon/plugins.py
```diff
@@ -460,10 +460,24 @@ def infer_dtype(self):
         return dtype
 
     def compute_loop(self, event, peaks):
+        __version__ = '0.1.0'
         result = dict(n_peaks=len(peaks))
         if not len(peaks):
             return result
 
+        #identify largest s2 properties
```
tunnell: Can replace with:

```
from collections import defaultdict
s2_info = defaultdict(None)
```

strax/xenon/plugins.py
```diff
@@ -460,10 +460,24 @@ def infer_dtype(self):
...
+        s2_values = None
+        s2_largest = None
+        s2_largest_ind = None
+        s2_largest_time = None
+        if 2 in peaks['type']:
+            s_mask = peaks['type'] == 2
+            ss = peaks[s_mask]
+            s2_values = ss['area']
```
tunnell: Can combine with above liine

JelleAalbers:
Thanks! Here is a suggestion which finds a different main S1 (which might work, I don't know :-) ):
```
        main_s = dict()
        # Note S2 must be first, see below
        for s_i in [2, 1]:
            s_mask = peaks['type'] == s_i

            # For determining the main S1, remove all peaks
            # after the main S2 (if there was one)
            # This is why S2 finding happened first
            if s_i == 1 and result[f's2_index'] != -1:
                s_mask &= peaks['time'] > main_s[2]['time']

            ss = peaks[s_mask]
```

Maybe at some time adding a max drift time option and another &= mask changing to avoid matching to S1s outside the max drift time would be useful

strax/xenon/plugins.py
```diff
@@ -485,6 +494,7 @@ def compute_loop(self, event, peaks):
                 for q in 'xy':
                     result[f'{q}_s2'] = s[q]
 
+        #bind the largest s1 to the largest s2 only for valid s1-s2 pairs
```
JelleAalbers: Ok, you mean: only compute the drift time if we identified a valid s1-s2 pair.

PR #84, opened 2018-08-29
Cuts pr

coderdj:
This is an attempt to add a few basic cuts to strax. Cuts added are the 1.3T FV cylinder, S1MaxPMT, and S1 low energy. They should be direct ports from lax/sciencerun0.

This also adds the max pmt value to peak_basics, previously only the max pmt index had been stored per peak. 

(all the changes are within strax/xenon)


strax/xenon/cut_plugins.py
```diff
@@ -0,0 +1,48 @@
...
+        return ret
+
+
+class S1LowEnergyRange(strax.Plugin):
+    """Pass only events with cs1<200"""
+    depends_on = ('events', 'corrected_areas')
+    dtype = [('cut_s1_low_energy_range', np.bool, "Event under 200pe")]
+
+    def compute(self, events):
+         ret = np.all([events['cs1'] < 200], axis=0)
```
JelleAalbers: Why do you need np.all here? I'm not sure what happens now, sounds like you're computing one boolean per chunk. I would have thought you'd just do `ret = events['cs1'] < 200`

JelleAalbers: Ah, I see, the extra []'s make a new axis, and you're all-ing over that axis which has one element here (so it's not needed here), but above you use it as a way of combining event selections. 

OK, I usually use the & sign for this, e.g. (cs1 > 70) & (r < 20) % ... .

JelleAalbers:
Just ignore Codacy's comments: it doesn't seem to understand the flexible signature of the compute method.

PR #85, opened 2018-08-29
Datastructure docs

JelleAalbers:
This adds a page to the documentation with information on the datastructure: columns provided, dependencies, and applicable configuration options. It also includes the notebook for generating this page. Eventually we should let this be a script that automatically runs on readthedocs.


JelleAalbers:
@tunnell this is the page I screen-shared during the meetings yesterday. I would like to link to this page in answering's Luca's comments. Can we merge this so readthedocs will build the html?

JelleAalbers:
Closing for now, I'll try to make a version that runs automatically

PR #86, opened 2018-08-31
Autogenerated datastructure

JelleAalbers:
This adds a page to the documentation with information on the datastructure: columns provided, dependencies, and applicable configuration options. The page should generate automatically in readthedocs, but I cannot test this until this is merged.

@tunnell can you have a look? It would be very useful to have this for discussions with computing.


docs/source/build_datastructure_doc.py
```diff
@@ -0,0 +1,157 @@
+"""Create datastructure documentation page
```
tunnell: What is this?  Not enough

docs/source/build_datastructure_doc.py
```diff
@@ -0,0 +1,157 @@
...
+    """Add four spaces to every line in x"""
+    y = ''
+    if isinstance(x, str):
+        x = x.split('\n')
+    for q in x:
+        y += '    ' + q
+    return y
+
+
+def build_datastructure_doc():
```
tunnell: docstring

PR #87, opened 2018-08-31
Rethrow exceptions

JelleAalbers:
This changes exception handling in the mailbox and processor systems so the original exception, rather than MailboxKilled, is thrown to the user.

This also adds a test to ensure the correct original exception is stored in the metadata; and changes the warning on loading data with no chunks to an error (new custom exception class DataCorrupted).


strax/mailbox.py
```diff
@@ -164,12 +165,7 @@ def _send_from(self, iterable):
             self.kill(reason=e.args[0])
             # Do NOT raise! One traceback on the screen is enough.
         except Exception as e:
-            try:
-                e_args = e.args[0]
-            except IndexError:
-                e_args = ''
-            self.kill(reason=f"{e.__class__.__name__}('{e_args}') "
-                             f'in {threading.current_thread().name}')
+            self.kill(reason=(e.__class__, e, sys.exc_info()[2]))
```
tunnell: If I add variables to the exception (e.key), will that get propogated out?

JelleAalbers: Yes, it reraises the original object

Issue #49, opened 2018-07-10
Nondeterministic test failure for test_overlap_plugin


JelleAalbers:
Another one like #37 from `test_overlap_plugin` https://travis-ci.org/AxFoundation/strax/jobs/402245124
Setting random seeds would help prevent random failures, on the other hand, then we catch less bugs like these... 

(Oh, I shouldn't restart the travis build after linking, then the error is gone. Well, it will happen again when the dice are against us).

JelleAalbers:
Examples in https://travis-ci.org/AxFoundation/strax/jobs/423461621 and https://travis-ci.org/AxFoundation/strax/jobs/423179604

PR #94, opened 2018-09-02
Fix processing algorithm edge cases

JelleAalbers:
This fixes two edge cases that our randomized testing sometimes flags: #49 and #37. This also includes explicit examples of the underlying problems in the test suite -- so if these problems arise again, we will get build failures every time.


JelleAalbers:
Merging to prevent random failures in all the builds. Note that CodeFactor suggests a change that would cause errors.

Issue #90, opened 2018-09-01
NoSuchKey exception from S3 backend


JelleAalbers:
Example:
```
import strax
st = strax.Context(
    storage=strax.SimpleS3Store(readonly=True),
    register_all=strax.xenon.plugins,
    register=strax.xenon.pax_interface.RecordsFromPax
)

st.get_meta('170401_1024', 'raw_records')
```

Traceback:
```
---------------------------------------------------------------------------
NoSuchKey                                 Traceback (most recent call last)
<ipython-input-2-74b520fd653b> in <module>()
      6 )
      7 
----> 8 st.get_meta('170401_1024', 'raw_records')

~/software/strax/strax/context.py in get_meta(self, run_id, target)
    617         for sf in self.storage:
    618             try:
--> 619                 return sf.get_metadata(key, **self._fuzzy_options)
    620             except strax.DataNotAvailable as e:
    621                 self.log.debug(f"Frontend {sf} does not have {key}")

~/software/strax/strax/storage/common.py in get_metadata(self, key, fuzzy_for, fuzzy_for_options)
    139                                          fuzzy_for=fuzzy_for,
    140                                          fuzzy_for_options=fuzzy_for_options)
--> 141         return self._get_backend(backend).get_metadata(backend_key)
    142 
    143     def find(self, key: DataKey,

~/software/strax/strax/storage/s3.py in get_metadata(self, backend_key)
    133         # Grab metadata object from S3 bucket
    134         result = self.s3.get_object(Bucket=BUCKET_NAME,
--> 135                                     Key=f'{backend_key}/metadata.json')
    136         # Then read/parse it into ASCII
    137         text = result["Body"].read().decode()

~/miniconda3/envs/py36/lib/python3.6/site-packages/botocore/client.py in _api_call(self, *args, **kwargs)
    312                     "%s() only accepts keyword arguments." % py_operation_name)
    313             # The "self" in this scope is referring to the BaseClient.
--> 314             return self._make_api_call(operation_name, kwargs)
    315 
    316         _api_call.__name__ = str(py_operation_name)

~/miniconda3/envs/py36/lib/python3.6/site-packages/botocore/client.py in _make_api_call(self, operation_name, api_params)
    610             error_code = parsed_response.get("Error", {}).get("Code")
    611             error_class = self.exceptions.from_code(error_code)
--> 612             raise error_class(parsed_response, operation_name)
    613         else:
    614             return parsed_response

[...]

Issue #73, opened 2018-08-23
RunDB aware FrontEnd


tunnell:
Currently, the strax frontend does not know about the rundb.  This should be fixed, but is blocked by #48 since this is XENON specific.  The implementation of this will also depend on work from @briedel.

JelleAalbers:
See #100

Issue #48, opened 2018-07-10
Split off xenon plugins


tunnell:
Ditch all the XENON experimental software from strax ;)

tunnell:
Specifically, move xenon folder out into its own repository.

tunnell:
bump

JelleAalbers:
After the current open PRs are merged, and after moving the stuff in `notebooks` into `xenon`, I'll try to do this with filter_branch (https://help.github.com/articles/splitting-a-subfolder-out-into-a-new-repository/).

This leaves two questions:
  * Will the new repo be called `straxen` or `xestrax`? I'm leaning towards `straxen`, since `xestrax` would have xenon it it twice.
  * Host it in the nT or 1T github? Even though the plugins are geared towards 1T reanalysis, I guess nT would make most sense, since it is the next software generation.

JelleAalbers:
This has been done in #107

PR #100, opened 2018-09-12
XENON RunDB Frontend

tunnell:
(This is XENON-specific, still need to do #48)

This addresses Issue #73.  There is a frontend that looks for data in a run database.  Currently, no default is given, so the MongoURL has to be fed in.  For example:

```
import strax

st = strax.Context(
    storage=strax.xenon.RunDB(mongo_url='mongodb://rundb-rw:XXXXX@rundbcluster-shard-00-00-cfaei.gcp.mongodb.net:27017,rundbcluster-shard-00-01-cfaei.gcp.mongodb.net:27017,rundbcluster-shard-00-02-cfaei.gcp.mongodb.net:27017/test?ssl=true&replicaSet=RunDBCluster-shard-0&authSource=admin&retryWrites=true'),
    register_all=strax.xenon.plugins,
    register=strax.xenon.pax_interface.RecordsFromPax
)

d = st.get_array('170321_0620', 'raw_records')
```

The way it works is by searching the RunDB for data of interest (this is the only RunDB call!).  If a run document exists, then it tries to determine if it knows how to read it.  S3 can be read anywhere.  A fuzzy hacky match is used (like in hax) to determine if you're running on `dali`, where other XENON data is.  At the moment, this interface cannot write RunDB aware.  This is reserved for snax.

This relies on a RunDB mirror hosted through MongoDB Atlas on Google Cloud Platform.  Which data is available can also be seen [here](https://snax-209012.appspot.com).

### DIFF
--- a/strax/storage/s3.py
+++ b/strax/storage/s3.py
@@ -73,7 +73,8 @@ class SimpleS3Store(StorageFrontend):
                                service_name='s3')
 
         # Create bucket (does nothing if exists)
-        self.s3.create_bucket(Bucket=BUCKET_NAME)
+        #self.s3.create_bucket(Bucket=BUCKET_NAME)
+        print(BUCKET_NAME)
 
         # Setup backends for reading
         self.backends = [S3Backend(aws_access_key_id=s3_access_key_id,
@@ -108,6 +109,7 @@ class SimpleS3Store(StorageFrontend):
                 raise strax.DataNotAvailable
 
     def backend_key(self, key_str):
+
         return self.backends[0].__class__.__name__, key_str
 
     def remove(self, key):
@@ -207,6 +209,7 @@ class S3Saver(strax.Saver):
         prefix = f'{self.strax_unique_key}/metadata_'
         objects_list = self.s3.list_objects(Bucket=BUCKET_NAME,
                                             Prefix=prefix)
+
         if 'Contents' in objects_list:
             for file in objects_list['Contents']:
                 # Grab chunk metadata as ASCIII
--- a/strax/storage/s3.py
+++ b/strax/storage/s3.py
@@ -17,8 +17,8 @@ import os
 import tempfile
 
 import boto3
-from botocore.exceptions import ClientError
 from botocore.client import Config
+from botocore.exceptions import ClientError
 
 import strax
 from strax import StorageFrontend
@@ -28,6 +28,7 @@ export, __all__ = strax.exporter()
 # Track versions of S3 interface
 VERSION = 2
 BUCKET_NAME = 'snax_s3_v%d' % VERSION
+
 
 @export
 class SimpleS3Store(StorageFrontend):
@@ -70,7 +71,7 @@ class SimpleS3Store(StorageFrontend):
         self.boto3_client_kwargs = {'aws_access_key_id': s3_access_key_id,
                                     'aws_secret_access_key': s3_secret_access_key,
                                     'endpoint_url': endpoint_url,
-                                    'service_name' :'s3',
+                                    'service_name': 's3',
                                     'config': Config(connect_timeout=5,
                                                      retries={'max_attempts': 10})}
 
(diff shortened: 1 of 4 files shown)


tunnell:
I just realized: this PR also includes a cleanup and fix of the S3 interface.

strax/xenon/rundb.py
```diff
@@ -0,0 +1,86 @@
...
+        :param s3_secret_access_key: secret key for S3-readable storage.
+        :param endpoint_url: URL of S3-readable storage.
+
+        For other arguments, see DataRegistry base class.
+        """
+
+        super().__init__(*args, **kwargs)
+
+        self.client = pymongo.MongoClient(mongo_url)
+        self.collection = self.client['xenon1t']['runs']
```
tunnell: This is one use.

strax/xenon/rundb.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _find(self, key: strax.DataKey, write, fuzzy_for, fuzzy_for_options):
+        """Determine if data exists
+
+        Search the S3 store to see if data is there.
+        """
+        if fuzzy_for or fuzzy_for_options:
+            raise NotImplementedError("Can't do fuzzy with S3")
+
+        query = {'name': key.run_id,
```
tunnell: This is another.

briedel:
To understand this correctly, strax only queries for the location of the data of a certain type?

JelleAalbers:
@briedel We will probably add some more rundb calls later for fetching of run-level metadata and possibly some configuration. 

RunDB write support should come soon, but not in this first concept:
  * We need to add a hook when data is done writing that calls back to the storage frontend (mongodb in this case). First of all we'll have to change the storage base classes a bit for this. 
  * Next, we'll hit a problem where we can't put this callback it in the most obvious place (Saver) in the most obvious way (attribute) because Savers can get forked over multiple processes (in low-level multiprocessing, the ParallelSourcePlugin), and you can't put unpickleable attributes (like callbacks, mongo clients, etc.) on things that will be forked. This isn't really a problem, as the forked versions don't need the callback, so we just have to erase it just before forking.

Even if we only use snax for writing, it's still useful to have write support in the strax frontend, to simplify snax's life.

tunnell:
PR good?

strax/storage/s3.py
```diff
@@ -66,19 +68,21 @@ def __init__(self,
...
         #  Initialized connection to S3-protocol storage
-        self.s3 = boto3.client(aws_access_key_id=s3_access_key_id,
-                               aws_secret_access_key=s3_secret_access_key,
-                               endpoint_url=endpoint_url,
-                               service_name='s3')
+        self.s3 = boto3.client(**self.boto3_client_kwargs)
 
         # Create bucket (does nothing if exists)
-        self.s3.create_bucket(Bucket=BUCKET_NAME)
+        # self.s3.create_bucket(Bucket=BUCKET_NAME)
```
JelleAalbers: Was this accidentally commented out or do we no longer need it?

strax/storage/s3.py
```diff
@@ -88,26 +92,26 @@ def _find(self, key, write, fuzzy_for, fuzzy_for_options):
...
-                raise strax.DataExistsError(at=bk)
-            return bk
-        else:
-            # No objects yet...
-            if write:
-                return bk
+        try:
+            self.backends[0].get_metadata(key)
+        except ClientError as ex:
+            if ex.response['Error']['Code'] == 'NoSuchKey':
```
JelleAalbers: Could you catch this in get_metadata and raise DataNotAvailable there? This would fix #90. (Then here, you'd catch DataNotAvailable again if you're writing.)

The S3 frontend is unusual in that `_find` relies on `get_metadata` rather than the other way around. But the context relies on DataNotAvailable being raised if get_metadata fails, so it can try other storage frontends (https://github.com/AxFoundation/strax/blob/master/strax/context.py#L632).

strax/xenon/rundb.py
```diff
@@ -0,0 +1,73 @@
...
+import strax
+
+export, __all__ = strax.exporter()
+
+
+@export
+class RunDB(strax.StorageFrontend):
+    """Frontend that searches RunDB MongoDB for data.
+
+    Loads appropriate backends ranging from Files to S3.
```
JelleAalbers: Maybe add the caveats about writing you wrote in the PR here. If I remember correctly, writing is supported, but it doesn't get registered back in the rundb.

Also perhaps worth noting the chosen write backend is files if you're on dali and S3 otherwise (right?).

strax/storage/s3.py
JelleAalbers: Actually it turns out #90 has already been fixed, so never mind.

PR #120, opened 2018-12-10
Fix merging of dtypes and adding number of saturated channels to peaks

petergaemers:
Fixed merging dtypes, the wrong area calculation and added saturated channels


JelleAalbers:
I should have clicked squash-merge, since there are many fix-commits... Anyway, I'm not going to revert since that would cause an even bigger mess.

PR #134, opened 2019-02-01
Several bugfixes in peak processing

ErikHogenbirk:
Errors appeared in a rare situation where the zero-padded part of a record overlaps with a peak, causing a nonphysical selection range. 'Lucky'  enough this situation was not so rare in XAMS-type data :) I believe the right behavior is a ''continue'' meaning to skip this record in the peak consideration.

### DIFF
--- a/strax/processing/peak_building.py
+++ b/strax/processing/peak_building.py
@@ -164,7 +164,7 @@ def sum_waveform(peaks, records, adc_to_pe, n_channels=248):
             assert r_end > r_start
             
             max_in_record = r['data'][r_start:r_end].max()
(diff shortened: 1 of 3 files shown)


JelleAalbers:
The width fix looks good too, at least it looks more like https://github.com/XENON1T/pax/blob/master/pax/plugins/peak_processing/BasicProperties.py#L118 now. Probably we should add a test for this...

ErikHogenbirk:
Happy to be able to help strax :) Yes, you are right, the assert was triggered. Good coding on that, would have been tricky to spot without it. Hope the width fix will help some cut acceptance problems that I heard @petergaemers mention.

PR #137, opened 2019-02-06
Adds more width calculations

darrylmasson:
Strax only calculates the central widths, but for peak classification other quantities like risetime are very useful. This PR adds `area_decile_from_midpoint` into the `peak_dtype` and its calculation into `compute_widths`. I tested this on some data from our small TPC and it seems to work for `n_widths = 11`. Other width numbers are untested.


jhowl01:
I played with this at the strax workshop in Chicago and it worked great for me, so approving.

JelleAalbers:
Thanks for implementing this @darrylmasson !

PR #159, opened 2019-05-09
MongoDB Output

coderdj:
Started with a mongodb store that Jelle wrote then removed from strax ~1 year ago and updated it to the frontend/backend/saver paradigm. It's for use in the online monitoring where some basic plugins will dump info to MongoDB for display online.

### DIFF
--- a/strax/storage/mongo.py
+++ b/strax/storage/mongo.py
@@ -1,0 +1,146 @@
+"""I/O format for MongoDB
+
+This plugin is designed for small amounts of basic information and 
+is specifically implemented for an online data monitoring system.
+Each chunk is processed separately and extracted information is stored
+in a single MongoDB document of the following form:
+
+doc = {
+  number: run_id,
+  plugin: plugin_name,
+  chunk: chunk_id,
+  data: {  PLUGIN_DATA },
+  }
+}
+
+Note that you must define Mongo connectivity parameters are environment
+variables. Note also that there is a 16 MB size limit on MongoDB documents,
+so plugins must be relatively light on data.
+"""
+
+try:
+    from pymongo import MongoClient
+except ImportError:
+    MongoClient = None
+
+import strax
+import numpy as np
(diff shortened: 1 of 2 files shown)


strax/storage/mongo.py
```diff
@@ -0,0 +1,137 @@
...
+
+This plugin is designed with data monitoring in mind, to put smaller
+amounds of extracted data into a database for quick access. However
+it should work with any plugin.
+
+Note that there is no check to make sure the 16MB document size
+limit is respected!
+"""
+
+try:
```
JelleAalbers: pymongo is now a dependency (due to using json_util also in the files backend), so this check is not needed.

strax/storage/mongo.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if write:
+            return self.backends[0].__class__.__name__, str(key)
+        if self.db[str(key)].count():
+            self.log.debug(f"{key} is in cache.")
+            return self.backends[0].__class__.__name__, str(key)
+        self.log.debug(f"{key} is NOT in cache.")
+        raise strax.DataNotAvailable
+
+    def _read_meta(self, coll_name):
```
JelleAalbers: This is not used right?

coderdj: Thanks for the suggestions. Updated.

One thing I just noticed is that I never call 'drop'. Is it standard in strax to leave partial data written on disk or clear it in case of a crash? I saw the allow_partial option so I'm assuming I could use that if I want to handle it somehow.

JelleAalbers: Thanks! Indeed it's standard to leave it; this dates back to when we thought online deletion was necessary. The only time strax clears data is while overwriting, I guess you could drop the collection if it already exists when you init the saver, just like in https://github.com/AxFoundation/strax/blob/master/strax/storage/files.py#L277. The frontend has an allow_overwrite option that can prevent savers getting created if the data already exists.

PR #192, opened 2019-06-20
fetch data sets based on run_id. Modif. in select_runs

mlotfiben:
allow for selection of data  set by giving one specific run_id and also it can  be used with matching a given numbers of run_ids (example: run_id="110*").

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -957,6 +957,18 @@ class Context:
         if pattern_type not in ('re', 'fnmatch'):
             raise ValueError("Pattern type must be 're' or 'fnmatch'")
 
+        if run_id is not None:
+            run_ids = dsets['run_id'].values
+            mask = np.zeros(len(run_ids), dtype=np.bool_)
+            if pattern_type == 'fnmatch':
+                for i, x in enumerate(modes):
+                    mask[i] = fnmatch.fnmatch(x, run_id)
+            elif pattern_type == 're':
+                for i, x in enumerate(run_ids):
+                    mask[i] = bool(re.match(run_id, x))
+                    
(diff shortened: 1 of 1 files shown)


JelleAalbers:
Thanks Lotfi. It currently crashes because there is not yet a run_id argument in the select_runs function. You can add one with a default of None (and update the docstring).

It seems the logic for run_id is very similar to that of run_mode. Maybe it would work to replace both of these with one for loop, like so:
```
for field_name, requested_value in (
        ('run_id', run_id), ('mode', run_mode)):

    values = dsets[field_name].values
    mask = np.zeros(len(values), dtype=np.bool_)
    if pattern_type == 'fnmatch':
        for i, x in enumerate(values):
            mask[i] = fnmatch.fnmatch(x, requested_value)
    elif pattern_type == 're':
        for i, x in enumerate(values):
            mask[i] = bool(re.match(requested_value, x))

    dsets = dsets[mask]
```
but I have not tested this.

mlotfiben:
Hi Jelle, thanks. when I used the fork version, I forgot to add the run_id
as argument to select_run.
Actually we need to leave it as it is, because then if the user give one of
them, either run_mode or run_id, the logic, as it is now will do the job.
But if we use the loop then both should be given at the same time, which I
think not the goal of the selection_runs.
Please correct me if I am wrong.
Best, Lotfi

JelleAalbers:
Ah sorry, in the loop we would have to add a check for None, otherwise indeed you'd have to give both.
```
for field_name, pattern in (
        ('run_id', run_id), ('mode', run_mode)):
    if pattern is None:
        continue
```

Issue #162, opened 2019-05-10
Handle empty data gracefully


JelleAalbers:

JelleAalbers:
After fixing several issues in strax (see above, and f54ce50a1d691c) and straxen (https://github.com/XENONnT/straxen/commit/f9a696040df5cdbd929d842214e90555d3595410), we can now process runs such as 180216_2226, 180219_0243 and 180217_2135, which (at least according to strax) have one or more chunks without any events. One or more of them even had a chunk without records, if I'm not mistaken.

I did not investigate what is going on with the data in these runs, but at least strax seems significantly more robust to empty chunks. Let's close this and open a new one if a specific problem comes up.

PR #212, opened 2019-10-10
Protect OverlapWindowPlugin against empty chunks

JelleAalbers:
As discovered by @jorana, strax's OverlapWindowPlugin crashes with an IndexError if an empty chunk is found in the data stream. In general strax is not very robust against this case (see #162), but we are improving.

You can read how OverlapWindowPlugin works [here](https://strax.readthedocs.io/en/latest/developer/overlaps.html#overlap-window-plugins). For an empty chunk, this PR changes strax to simply not send or discard any data; the input/output caches remain the same. The case of the last chunk (after which you must send everything even it if is empty) is already handled by the `iter` method. 

This escaped the property-based testing due to a spurious condition in the test (I guess I just didn't consider empty chunks would be possible) that is now removed.

(This also removes some old but totally unrelated commented code)

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -145,43 +145,18 @@ class Plugin:
             lineage=self.lineage)
 
     def dependencies_by_kind(self, require_time=None):
+        """Return dependencies grouped by data kind
+        i.e. {kind1: [dep0, dep1], kind2: [dep, dep]}
+        :param require_time: If True, one dependency of each kind
+        must provide time information. It will be put first in the list.
+
+        If require_time is omitted, we will require time only if there is
+        more than one data kind in the dependencies.
+        """
         return strax.group_by_kind(
             self.depends_on,
             plugins=self.deps,
             require_time=require_time)
-        # """Return dependencies grouped by data kind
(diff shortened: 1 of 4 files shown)


jorana:
Thanks Jelle! It does solve the problem but causes problems elsewhere as it will keep waiting after which it crashes due to an ```strax.plugin.InputTimeoutExceeded``` error. 

Could this be that the problem lies in that from line https://github.com/AxFoundation/strax/pull/212/commits/52335ec112607147c27d9a9f0d6bd8bb4b9ce76d#diff-db9a6da7a746e72e441939269037c9e6R126 there will never be an ‘end’ so it will not know that that chuck is finished (but just empty)?

JelleAalbers:
InputTimeoutExceeded should come from the DAQReader, this is an exception raised as part of the  [online input support](https://github.com/AxFoundation/strax/blob/45de045812f3c0ab129f09e01add386b60c75ead/strax/plugin.py#L226). I would guess this is an unrelated problem, e.g. maybe something is wrong with how that particular run ended?

PR #219, opened 2019-11-12
Cleanup tests, make helpers accessible

JelleAalbers:
This cleans up strax' test helpers file, which contains mainly [hypothesis](https://hypothesis.works/) strategies for making random records and peaks for testing data processing routines.

There are now no side effects from importing it -- in particular, we no longer mock numba for performance during tests. This makes the tests take slightly longer, but only if the compilation cache is not present (e.g. on Travis or after changes). Since we use numba in production we should clearly also test with it. Also, since numba is considering using strax as an integration test (https://github.com/numba/numba-integration-testing/issues/29) I'm sure they would appreciate if we don't replace `sys.modules['numba']` with a MagicMock...

The test helpers file is made accessible as strax.testutils, since some of the functions defined in it could be useful in writing tests for straxen.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,10 @@ deploy:
 matrix:
     include:
         - env: PYTHON=3.7 DEPLOY_ME=true
+        - env: PYTHON=3.7 NUMBA_DISABLE_JIT=true
         - env: PYTHON=3.6
         - os: osx
           env: PYTHON=3.7
-        - env: PYTHON=3.7
-          dist: trusty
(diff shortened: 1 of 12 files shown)


JelleAalbers:
The coverage drop is due to the numba-jitted functions [being missed in the coverage computation](https://stackoverflow.com/questions/26875191). I disabled jitting in one of the matrix builds and am now submitting coverage only for that, but this does not seem to work either (at least not yet).

PR #223, opened 2019-12-18
Specify defaults via run metadata / run doc

JelleAalbers:
This allows runs to specify default values for strax processing options via their metadata (run documents / database) using the `strax_defaults` field. This is very similar to the `processor...` field in XENON1T and will hopefully make correction handling a lot easier. Strax configruration priority is now as follows (high to low):
  * Context options (set by user)
  * Defaults specified in run metadata (i.e. run document/database)
  * Defaults specified with plugin definition, using `default`, `default_factory` or `default_per_run` (in that order of priority).

The `default_per_run` option will soon be deprecated/removed , now that this new way of specifying run-dependent defaults is available. Currently `default_per_run` forces strax runs to have names that are castable to integers, which breaks proper functioning of superruns (as @petergaemers noted) and is a pain more generally.

The original motivation behind `default_per_run` was that I liked to enable processing independent of a runs DB. Since run metadata can for some time already be provided by json files (as we currently do on dali), that motivation no longer applies.


JelleAalbers:
Hmm, this might be a mistake: we should not like XENON1T set changing options in the run doc, that makes it very difficult to go back and reproduce old results. Have to think about this...

PR #227, opened 2020-01-08
Change the default compressor to LZ4

jorana:
For very large datafiles, ``blosc`` fails since the input buffer becomes larger than ~2 GB. Changing to LZ4 as default compressor allows to safely safe these larges chunks.

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -48,7 +48,7 @@ class Plugin:
     depends_on: tuple
     provides: tuple
 
-    compressor = 'blosc'
+    compressor = 'lz4'
 
     rechunk_on_save = True    # Saver is allowed to rechunk
 


jorana:
It turns out that during multiprocessing of ~2-3 GB chunks this gives a rather incomprehensible error. 

If the chunks get any larger than 3 GB the same error is still raised.

The error looks like:
```
Exception in thread Thread-1:
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/threading.py", line 864, in run
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/concurrent/futures/process.py", line 295, in _queue_management_worker
    shutdown_worker()
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/concurrent/futures/process.py", line 253, in shutdown_worker
    call_queue.put_nowait(None)
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/multiprocessing/queues.py", line 129, in put_nowait
    return self.put(obj, False)
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/multiprocessing/queues.py", line 83, in put
    raise Full
queue.Full

162689 MainThread root Still processing run 006318
Exception in thread save_0:raw_records:
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/threading.py", line 864, in run
  File "/home/xedaq/software/strax/strax/storage/common.py", line 456, in save_from
    source.throw(e)
  File "/home/xedaq/software/strax/strax/storage/common.py", line 439, in save_from
    for chunk_i, s in enumerate(source):
  File "/home/xedaq/software/strax/strax/mailbox.py", line 316, in _read
    res = msg.result(timeout=self.timeout)
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
  File "/home/xedaq/software/strax/strax/processor.py", line 196, in iter
    yield from final_generator
  File "/home/xedaq/software/strax/strax/mailbox.py", line 316, in _read
    res = msg.result(timeout=self.timeout)
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/concurrent/futures/_base.py", line 432, in result
    return self.__get_result()
  File "/home/xedaq/miniconda/envs/py36/lib/python3.6/concurrent/futures/_base.py", line 384, in __get_result
    raise self._exception
concurrent.futures.process.BrokenProcessPool: A process in the process pool was terminated abruptly while the future was running or pending.
```

JelleAalbers:
Thanks Joran! The 2GB limitation is a serious issue. On the other hand, from a comparison on X1t background data (on my laptop), lz4 is usually considerably worse than blosc. It can be up to 3x slower for loading or saving while the filesize is usually the same: 

compress_benchmark_load

compress_benchmark_save

So perhaps we should use lz4 for the moment only for the low-level datatypes that really need it, rather than making it the default? We could also look into blosc2 (https://blosc.org/posts/blosc2-first-beta/, https://github.com/Blosc/c-blosc2, https://github.com/Blosc/python-blosc2), which is apparently out in beta and has 64-bit support. Or maybe https://github.com/Blosc/bloscpack.

jorana:
Thanks for the feedback Jelle, I'm going to move this PR to straxen and the lower level plugins there.

Issue #230, opened 2020-02-10
Problems with time range selection


JelleAalbers:
In the latest version of straxen @tunnell and @aladinggit caught a problem with the time range selection, as used for the waveform plotting (https://github.com/XENONnT/straxen/issues/48#issuecomment-584316043). This is most likely due to the recent change https://github.com/AxFoundation/strax/commit/5f795afecc6d2d05109cc50f590cc7ff1fe5f303.

JelleAalbers:
@tunnell @aladinggit Could you see if your issue is resolved in v0.8.8? I deployed it to the backup environment.

tunnell:
The backup environment is the conda environment on dali?

JelleAalbers:
The time range selection code has been completely rewritten (and much simplified) in https://github.com/AxFoundation/strax/pull/235. Let's make a new issue if one arises.

Issue #79, opened 2018-08-28
Store time info with all strax data types?


JelleAalbers:
Currently not all strax data types contain time information, which makes merging logic in a few places of strax hard. E.g. for random access we first have to convert a time range to row numbers of the corresponding data kind.

The most storage-expensive do already have time information, so storing time info with everything would only slightly increase the storage requirement of the high-level data. It will require some changes in strax to add this information automatically though (so not every plugin developer has to copy the time fields)

JelleAalbers:
If we don't do this, it would at least be useful to support seconds_range / time_range for dtypes without registered time info. Perhaps users will have to pass an extra time_info_from argument or something

PR #231, opened 2020-02-11
Fix time range finding issues

JelleAalbers:
This is a fix for #230, i.e. the treatment of the time_range / seconds_range / time_within options.

Background info:
  * Strax supports (a) data without time information, such as `peaklet_classification` or `corrected_areas` and (b) data that is never stored, such as peaks, or the temporary datatypes that are made to merge e.g. `peaks` and `peak_basics` if you load them together.
  * Strax even allows time selection with a no-time datatype as long as you load it alongside data of the same kind that does have time info (like `peaks`). It will then use the latter to map the time range to row numbers, which can be selected in the no-time datatypes.

There were two bugs:
  * Strax was too coarse when converting time ranges to row numbers. We picked whichever time-bearing datatype was listed first, found out in which chunk the start and stop of the range were contained, then loaded the row numbers corresponding to those entire chunks. If this data type happened to be very slim (few columns), a chunk would correspond to way more time than you selected (since we try to save in fixed-size files), possibly the entire run. If the simultaneously loaded data types were not slim, your RAM might explode.
  * The time range argument and associated conversion to row ranges could be spuriously skipped. When loading a never-stored datatype such as `peaks` together with a time-bearing datatype of the same kind, such as `peak_basics`, `peaks` still has to be reconstructed from `peaklets` and `merged_s2s`. However, the time range was not passed on to the latter, so strax attempted to load the full data in the run, starting with the first chunk, which would usually cause a merge failure. (... but not always; due to the very coarse chunk mapping you were fine if the run only had one chunk or perhaps if you were just loading the first chunk.)

Now:
  * Strax actually loads data to find out the exact row numbers that correspond to the time range (only for chunks containing the start and stop time, so possibly just one chunk). This does mean that some data is loaded twice when doing time_range selections.
  * If a `time_range` is specified, it will apply to all data types that are loaded. There are likely still cases where the current logic fails, e.g. when you make some deep chain of no-time-bearing and never-stored plugins. Then you will get a neater error: it will say which data type it tried in vain to find time information for (rather than a weird merge failure or blowing up your memory).
  * `_time_range_to_n_range` has been renamed to `time_range_to_row_range` for clarify.
  * An additional option `_chunk_number` is added to `get_array`, `get_df`, etc, to load just a single chunk. This is used internally in `time_range_to_row_range`.

Taking a step back, this is all additional complexity that comes from supporting data types without time information (see also https://github.com/AxFoundation/strax/issues/79). If I knew how much trouble this would have brought I would never have supported it -- it saves a bit of space, but not so much. However, by now, it is probably too big of a downstream change to stop supporting this.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -569,10 +569,23 @@ class Context:
             # Can we load this data?
             loading_this_data = False
             key = strax.DataKey(run_id, d, p.lineage)
+
+            if time_range:
+                kind = p.data_kind_for(d)
+                if kind in row_range_for:
+                    row_range = row_range_for[kind]
+                else:
+                    # Fallback, can happen if target and one of its dependencies
+                    # is not stored (e.g. merging peaks with something else)
+                    row_range = self._time_range_to_row_range(
+                        run_id, time_range, d)
+            else:
+                row_range = None
+
             ldr = self._get_partial_loader_for(
                 key,
                 chunk_number=chunk_number,
-                row_range=row_range_for.get(p.data_kind_for(d)))
+                row_range=row_range)
 
             if not ldr and run_id.startswith('_'):
                 if time_range is not None:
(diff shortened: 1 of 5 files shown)


JelleAalbers:
So now codefactor thinks that using an f-string in an error message is a possible SQL injection vulnerability... right.

Issue #81, opened 2018-08-29
Factor out mailbox


tunnell:
Mailbox should be factored out into an independent repository rather than using something else.  Closes #62.

JelleAalbers:
By now I've come to think the mailbox approach is fundamentally flawed. The original idea was to build a simple class where one function could push and another could iterate. But it has turned into a complicated piece of custom threading code, with lots of special cases, and always the specter of a hang or timeout if things go really wrong.

I'm now working on replacing the mailbox system with a single scheduler / work distributor.  This will be a medium-term project of a few weeks or months most likely, since I can only work on it off and on. There is a possibility it won't work out. The mailbox system has worked well for us so far fortunately, but it does make debugging complicated problems (e.g. Joran's comment in #212), well, complicated.

tunnell:
This might be a good project for a CS student, you think?  At very least, can speak with some CS systems people who might have ideas on how to replace.  Pivarski might also be a resource.

JelleAalbers:
Indeed this would be an interesting project, especially to generalize it to a multi-machine system, perhaps using zeromq or something similar. 

I've put what I have now here https://github.com/JelleAalbers/plarx

jmosbacher:
Is this issue being worked on? if not I can take a look at it.

tunnell:
Doubt it... It's a bit of a black hole that is ripe for overengineering, so be careful :)  Be aware that the DAQ use case drives the performance requirements

jmosbacher:
# The Mailboxectomy
I made a quick review of the options we have for using 3rd party tooling for processing instead og the ThreadedMailboxProcessor. Here are my findings.

### Frameworks looked at
- Pyleus (Apache Storm)
- streamparse (Apache Storm)
- Wallaroo
- Apache Pulsar
- Ray
- Heron
- Faust

## Details
#### Apache Storm and its successor Heron
The topology of spouts and bolts are conceptually very similar to strax plugins, and have extremely battle tested implementations in python, they are mostly suitable for distibuted processing on large clusters and would not be suitable for a "one guy with a laptop" processing situation. 
#### Apache Pulsar
If strax moved to a pub-sub configuration this would be suitable for the online processing mode of running, on a cluster of high performance machines. It boasts great throughput and reliability. Probably overkill for what xenonnt needs.
#### Wallaroo
Very interesting project, but implemented in a pre-version 1 language called pony, has a limited python API. May be worth revisiting for DARWIN.
#### Faust
[...]

PR #235, opened 2020-02-21
Explicit time tracking, remove raw_records fields

### LINKED ISSUE #79: Store time info with all strax data types?

JelleAalbers:
This makes time handling in strax more consistent and explicit. From now on:
  1. **Every strax datatype must include `time` and either `endtime` or `length` and `dt`**. This closes #79 and makes selecting time ranges in a run much easier.
  2. **Chunks state what time range and run id they correspond to**. This means even an empty chunk has an associated time. Only from-scratch plugins  (DAQReader, WFsim, and RecordsFromPax) need to change something: for other plugins, the info is fetched from their dependencies and added to their output's chunk metadata automatically.

These are the first backwards-incompatible changes in strax since April 2019 (https://github.com/AxFoundation/strax/pull/152). Existing data will become unreadable by strax versions released after this PR is merged (since chunk-level metadata must now contain two extra fields, `start` and `end`).

From a user perspective, [1] seems like the big change, since it requires a small change in several plugins like peaklet_classification or corrected_areas. I considered auto-adding the fields behind the scenes, but then a voice in my head said 'explicit is better than implicit'.

For strax's internals, [2] is a much bigger change. Without this, empty chunk handling is really hard and not robust (see e.g. https://github.com/AxFoundation/strax/issues/162, https://github.com/AxFoundation/strax/pull/212, https://github.com/AxFoundation/strax/pull/201, https://github.com/XENONnT/straxen/issues/49). Actually [2] is what forced my hand on [1]: chunks of data without time info cannot be split, since we wouldn't know what time range the fragments correspond to.

Since it's best to do backwards-incompatible changes in one go, I added one more:

  3. **raw_records no longer contains the `baseline`, `area`, and `reduction_level` fields**. Again, this impacts only from-scratch pluigins.

The tests pass, but it isn't quite ready. I'd still like to:
  * Add a converter script to change old to new raw_records
  * Try this on straxen, fix bugs this reveals, then make a straxen compatibility PR

I hope this helps prepare the ground for a refactor or replacement of chunk_arrays and mailboxes, see #81.


WenzDaniel:
About [1] I think dt should be always mandatory since we will have different sample rates for nVETO and TPC/µVeto.

JelleAalbers:
Allright, let's do this.

@WenzDaniel: dt is necessary for data types that have embedded waveform data, such as peaks or raw_records. For higher-level data such as events, providing just time and endtime should be sufficient -- both can be done to ns precision if needed.

PR #242, opened 2020-03-12
Add info to ValueErrors for np.frombuffer

jorana:
This could be a very small change to add some extra info in to a ValueError arising from np.frombuffer.

This has proven to be useful after the latest strax(en) format change and sorting out if redax-output and strax(en) input are in sync.


jorana:
Closing PR and reopening in a moment. GitKraken has scrambled up the git history.

PR #253, opened 2020-04-08
Wrapper for get_iter to get accumulated results. 

WenzDaniel:
This function allows to get an array with accumulated results for the specified field names. All other redundant fields are returned only a single time. A typical use case of this feature is e.g. the hitfinder acceptance, for which one is rather interested in how many hits did I find per threshold per pmt per calibration run rather than per chunk. 

Note: The doc string has to be finished. I saw in the docstring of get_iter and get_array the {get_docs}-statement, but I do not have any clue how this works.


strax/context.py
```diff
@@ -904,6 +904,41 @@ def get_array(self, run_id: ty.Union[str, tuple, list],
             results = [x.data for x in source]
         return np.concatenate(results)
 
+    def get_accum_array(self,
```
JelleAalbers, line 907: Maybe `get_accumulated`? A one-element array with datatypes is perhaps a bit inconvienent/odd: how about returning a dictionary instead?

strax/context.py
```diff
@@ -904,6 +904,41 @@ def get_array(self, run_id: ty.Union[str, tuple, list],
...
+                        fields,
+                        **kwargs):
+        """
+        Computes targets for run_id and accumulated result for specified
+        fields.
+
+        :param fields: Column names of the data which should be
+            accumulated.
+        """
+        ind = 0
```
JelleAalbers, line 919: Unless I misread something, it doesn't look like this ever gets incremented, so the result will always be reset. You could wrap get_iter in enumerate if you want to know the chunk index.

strax/context.py
```diff
@@ -904,6 +904,41 @@ def get_array(self, run_id: ty.Union[str, tuple, list],
...
+            data = chunk.data
+            chunk_start = chunk.start
+            chunk_end = chunk.end
+
+            if not ind:
+                # Initiate result and put common values:
+                res = np.zeros(1, dtype=data.dtype)
+                for name in data.dtype.names:
+                    if not name in fields:
+                        res[name] = data[0][name]  # # Take static values from very first chunk.
```
JelleAalbers, line 930: Double # here.

strax/context.py
```diff
@@ -904,6 +904,41 @@ def get_array(self, run_id: ty.Union[str, tuple, list],
...
+                        # previous chunks
```
JelleAalbers, line 931: ?

strax/context.py
WenzDaniel: Hej Jelle, sorry for my late reply. Sounds great, 

> A one-element array with datatypes is perhaps a bit inconvienent/odd: how about returning a dictionary instead? 

Do you mean something like: `{'channel': np.array([1, 2, 3, 4, 5 ...]), 'dt': 10, ....}` or only in the case of a real one-element array like idk: `np.array([1,2,3,4,5,6], ('Single elements', 'bla'), dtype=np.int32)` 

>  How about passing the sum function as an argument, so the user could also do e.g. std or other simple reductions quickly?

I like that idea, will change it.

strax/context.py
WenzDaniel: Oh yes, good catch thanks.

strax/context.py
```diff
@@ -904,6 +904,64 @@ def get_array(self, run_id: ty.Union[str, tuple, list],
...
+                        res[name] = data[0][name]  # Take static values from very first chunk.
+                res['time'] = chunk_start  # Easier to simply overwrite value.
+
+            # Now accumulate other values:
+            chunk_res = function(data, fields)
+            for key, value in chunk_res.items():
+                if not nchunks:
+                    res[key] = value
+                else:
+                    res[key] += value
```
WenzDaniel, line 957: I think I implemented now your suggestions. Though, I did not see how I could implement the user specific function argument that one could compute e.g. the std or mean of a parameter over the entire run. I think here you still have to use get_iter.

JelleAalbers:
Thanks Daniel! I also made a few changes, see below. I'm realizing this is quite a versatile analysis tool... surprised we never added something like this to strax earlier.

  * Support more `function` types. They can now either return:
    * A dict or record array (fields will be accumulated as before);
    * A flat array or scalar (result will be accumulated under 'result');
    * None (nothing is accumulated, the function will just be run over the data).
  * `function` and `fields` are now optional: if you do not pass them, we will use the identity function and accumulate all fields
  * Added the `n_rows` field to the output
  * By default the function does *not* take fields anymore; though you can pass a flag to re-enable that behaviour. 
  * Storing the first entry of non-accumulated fields is now also optional and off by default.
  * Minor changes for robustness, e.g. if fields is a string we will interpret it as a 1-tuple string, if the first chunk is empty we won't try to get the first row from it, etc.

Here are a few examples, assuming this setup:
```python
import multihist
import straxen
import matplotlib.pyplot as plt

st = straxen.contexts.xenon1t_dali()
run_id = '180215_1029'
```

**Count pulses in a channel**
```python
# Count pulses in a channel
acc = st.get_accumulated(
    run_id, 
    'records',
    selection_str='(channel == 126) & (record_i == 0)')
acc['n_rows']
```
gives 7355. Note the record_i == 0 selection to make sure multi-record pulses are counted only once.

**Very rough pulse shape**
```python
# obtain acc as before
plt.plot(acc['data'] / acc['n_rows'])
```
gives

index

The rise at late times should due to records that are more than just one single PE, e.g. in S1s or S2s, or photons with afterpulses. A (slightly) better pulse shape analysis is [here](https://straxen.readthedocs.io/en/latest/tutorials/pulse_analysis.html), a much better once is [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenon1t:gaemers:spe_waveform_shape).

**Make a (channel, height) 2D histogram for lone hits**
```python
mh = multihist.Histdd(dimensions=(
    ('channel', np.arange(249) -0.5),
    ('height', np.linspace(0, 200, 100))))

st.get_accumulated(run_id, 
                   'lone_hits', 
                   function=mh.add)
mh.plot()
```
index_1

This shows the use of supporting functions that don't return anything.

**Plot the lone hit rate per channel**
```python
h = multihist.Hist1d(bins=np.arange(249))
acc = st.get_accumulated(run_id, 
                         'lone_hits', 
                         function=lambda x : h.add(x['channel']))
h /= (acc['end'] - acc['start']) / 1e9
straxen.plot_pmts(h.histogram, log_scale=True, vmin=20,
                  xenon1t=True, show_tpc=False, r=52, 
                  label='Lone hit rate [Hz]')
```

index_2


**Standard deviation of peak areas**

Approximately (average of stdevs per chunk):
```python
acc = st.get_accumulated(run_id, 'peaks', function=lambda x: x['area'].std())
acc['result'] / acc['n_chunks']
```
gives 32338.845703125.

Or exactly in two passes:
```
acc = st.get_accumulated(run_id, 'peaks')
mean_area = acc['area'] / acc['n_rows']

sum_sq = st.get_accumulated(run_id, 'peaks', function=lambda x: (x['area'] - mean_area)**2)['result']
(sum_sq/acc['n_rows'])**0.5
```
gives 32793.147536563476. 

The last two examples show the use of supporting functions that return a flat array.

JelleAalbers:
I keep changing my mind about the name... ;-) How about just `accumulate` instead of the somewhat awkward `get_accumulated`? I know some other methods start with `get_`, but `accumulate` is an actual verb while `array` and `df` are not.

WenzDaniel:
Wow I am impressed. I like the changes, good job. :-) For me it is fine and we can merge, as I  see you already extensively tested the function ;-) 

About the name, I totally see your points and thought the same. I think it is fine if we just go with accumulate but we should then advertise it a bit. Otherwise I could imagine people will simply overlook it.

JelleAalbers:
Thanks Daniel! Yes, we would need to make sure others know about this, maybe by making a new straxen tutorial or extending one of the old ones.

PR #266, opened 2020-05-20
fix flip boolean bug in pulse_processing

sophiaandaloro:
Replaced (-1*flip) with np.where(flip, -1, 1) to fix flip=False bug in records flipping.
using np.where with flip as boolean, will return -1 if flip=True, else returns 1. 
Previously boolean meant if flip=False d['data'] will go to 0.

Also I believe that calculating https://github.com/AxFoundation/strax/blob/8e94cd95d823c1537cd49a568d8a3ee4e0204ea5/strax/processing/pulse_processing.py#L40
in baseline computation is unnecessary.

### DIFF
--- a/strax/processing/pulse_processing.py
+++ b/strax/processing/pulse_processing.py
@@ -37,7 +37,6 @@ def baseline(records, baseline_samples=40, flip=True,
     """
     if not len(records):
         return records
-    samples_per_record = len(records[0]['data'])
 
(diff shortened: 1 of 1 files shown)


sophiaandaloro:
Yes, I like `(-1 if flip else 1)` the most, much simpler for people. I will edit it for that and thanks for the more elegant suggestion.

PR #271, opened 2020-06-09
Added default check to register

WenzDaniel:
Checking if defaults are set consistently when registering a new plugin.


jorana:
Hi Daniel, you beat me. I added:
```
    if hasattr(plugin_class, 'takes_config'):
        for p in self._plugin_class_registry.values():
            for k ,v in p.takes_config.items():                
                if (k in plugin_class.takes_config):
                    if (v.get_default('0') !=
                        plugin_class.takes_config[k].get_default('0')):
                        raise ValueError(f'option {k} for {p} differs from {plugin_class}')    
```
Maybe you can also use the ```plugin_class.takes_config```, other than that it looks good!

strax/context.py
```diff
@@ -239,6 +239,39 @@ def register(self, plugin_class):
         for p in plugin_class.provides:
             self._plugin_class_registry[p] = plugin_class
 
+        already_seen = []
```
jorana, line 242: - Why would you expect to see plugins multiple times? Even if you would, you make a new list each time you call ```register``` so I'm not sure if it'd help.

strax/context.py
```diff
@@ -239,6 +239,39 @@ def register(self, plugin_class):
...
+                                           f' for the same option. The option',
+                                           f' "{new_option}" in "{plugin.__name__}" takes',
+                                           f' as a default "{default}"  while in',
+                                           f' "{plugin_class.__name__}" the default value'
+                                           f' is set to "{new_default}". Please change'
+                                           ' one of the defaults.'
+                                           ))
+                            raise ValueError(mes)
+
+                except strax.InvalidConfiguration:
```
jorana, line 271: -  I haven't ran into ```trax.InvalidConfiguration```, could it be because I used the ```k in plugin_class.takes_config```?

strax/context.py
```diff
@@ -239,6 +239,39 @@ def register(self, plugin_class):
...
+
+            for option, items in plugin.takes_config.items():
+                try:
+                    # Looping over the options of the new plugin and check if
+                    # they can be found in the already registered plugins:
+                    for new_option, new_items in plugin_class.takes_config.items():
+                        if not new_option == option:
+                            continue
+                        default = items.get_default('1')  # Have to pass will be changed.
+                        new_default = new_items.get_default('1')
```
jorana, line 257: - I think '0' is a safe default?

strax/context.py
```diff
@@ -239,6 +239,39 @@ def register(self, plugin_class):
         for p in plugin_class.provides:
             self._plugin_class_registry[p] = plugin_class
 
+        already_seen = []
+        for provides, plugin in self._plugin_class_registry.items():
```
jorana, line 243: - ```for **provides, plugin** in self._plugin_class_registry**.items()**: ```->```for **plugin** in self._plugin_class_registry**.values(**)```:

jorana:
By the way I've checked your implementation* and it works like a charm!

*See (at bottom of): ```home/angevaare/software/straxferno/4.Contexts_run_selection_and_configuration.ipynb```

```ValueError: Two plugins have a different default value for the same option. The option "trigger_min_area" in "Events" takes as a default "100"  while in "DummyClass" the default value is set to "110". Please change one of the defaults.```

File: strax/context.py [hunk 5]
```diff
@@ -239,6 +239,39 @@ def register(self, plugin_class):
...
+                    # they can be found in the already registered plugins:
+                    for new_option, new_items in plugin_class.takes_config.items():
+                        if not new_option == option:
+                            continue
+                        default = items.get_default('1')  # Have to pass will be changed.
+                        new_default = new_items.get_default('1')
+                        if default == new_default:
+                            continue
+                        else:
+                            mes = ''.join((f'Two plugins have a different default value',
```
jorana, line 261: Somewhat cleaner is:
```                      
                            mes = (f'Two plugins have a different default value'
                                   f' for the same option. The option'
                                   f' "{new_option}" in "{plugin.__name__}" takes'
                                   f' as a default "{default}"  while in'
                                   f' "{plugin_class.__name__}" the default value'
                                   f' is set to "{new_default}". Please change'
                                   ' one of the defaults.'
                                   )
```

strax/context.py
jorana: Jelle just answered this question. Your way of keeping track of what has been seen is somewhat faster indeed.

WenzDaniel:
Thanks for your comments. As discussed, I added all needed changes.

WenzDaniel:
> 
> 
> Great! I think you only missed this one:
> 
> for **plugin** in self._plugin_class_registry.values():

Super strange, I changed .times() into values(), but forgot to remove the provides field. Sometimes I doubt my mental health... :D

PR #274, opened 2020-06-23
CutPlugin class for cuts in straxen

jorana:
The cuts currently in straxen (https://github.com/XENONnT/straxen/blob/master/straxen/plugins/x1t_cuts.py ) have a quite lot of overlap with one another. Most of the time a simple Boolean array is all that is returned. However, we each time have to add the time fields et cetera which may somewhat obscure the functionality.
To simplify the cuts somewhat I've written a simple Plugin Class. In straxen it could be used as for example:
```
class CutS1s(CutPlugin):
    depends_on = ('events', 'corrected_areas')
    provides = 'cs1_cut'
    cut_description = "Remove cS1 above 100 PE"
    cut_name ='cut_high_cs1'
    __version__ = '0.0.0'
    
    def cut_by(self, events):
        return events['cs1'] > 100
```
```
class CutS2s(CutPlugin):
    depends_on = ('events', 'corrected_areas')
    cut_description = "Remove cS2s below 100 PE"
    
    def cut_by(self, events):
        return events['cs2'] < 100
```

```
class CutRecord0(CutPlugin):
    depends_on = ('records',)
    provides = 'is_not_record0_cut'
    cut_description = "is record 0"
    cut_name ='is_not_record0_cut'
    __version__ = '0.0.0'
    
    def cut_by(self, records):
        return records['record_i'] != 0
```

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -692,6 +692,7 @@ class CutPlugin(Plugin):
         # This should be provided by the user making a CutPlugin
         raise NotImplementedError()
 
+
 ##
 # "Plugins" for internal use
 # These do not actually do computations, but do other tasks
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -847,3 +847,50 @@ class ParallelSourcePlugin(Plugin):
             for s in savers:
                 s.close(wait_for=wait_for)
         super().cleanup(iters, wait_for)
+
+
+@export
+class CutPlugin(Plugin):
+    """Generate a plugin that provides a boolean for a given cut specified by 'cut_by'"""
+    save_when = SaveWhen.NEVER
+
+    def __init__(self):
+        super().__init__()
+
+        _name = strax.camel_to_snake(self.__class__.__name__)
+        if not hasattr(self, 'provides'):
+            self.provides = _name
+        if not hasattr(self, 'cut_name'):
+            self.cut_name = _name
+        if not hasattr(self, 'cut_description'):
+            _description = _name
+            if 'cut_' not in _description:
+                _description = 'Cut by ' + _description
+            else:
+                _description = " ".join(_description.split("_"))
+            self.cut_description = _description
+
+    def infer_dtype(self):
+        dtype = [(self.cut_name, np.bool, self.cut_description)]
+        dtype = dtype + strax.time_fields
+        return dtype
+
+    def compute(self, **kwargs):
+        if hasattr(self, 'cut_by'):
+            cut_by = self.cut_by
+        else:
+            raise NotImplementedError(f"{self.cut_name} does not have attribute 'cut_by'")
+
+        # Take shape of the first data_type like in strax.plugin
+        buff = list(kwargs.values())[0]
+
+        # Generate result buffer
+        r = np.zeros(len(buff), self.dtype)
+        r['time'] = buff['time']
+        r['endtime'] = buff['endtime']
+        r[self.cut_name] = cut_by(**kwargs)
+        return r
+
+    def cut_by(self, **kwargs):
+        # This should be provided by the user making a CutPlugin
+        raise NotImplementedError()
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -871,7 +871,7 @@ class CutPlugin(Plugin):
             self.cut_description = _description
 
     def infer_dtype(self):
-        dtype = [(self.cut_name, np.bool, self.cut_description)]
+        dtype = [(self.cut_name, np.bool_, self.cut_description)]
         dtype = dtype + strax.time_fields
         return dtype
 
@@ -887,7 +887,7 @@ class CutPlugin(Plugin):
         # Generate result buffer
         r = np.zeros(len(buff), self.dtype)
         r['time'] = buff['time']
-        r['endtime'] = buff['endtime']
+        r['endtime'] = strax.endtime(buff)
         r[self.cut_name] = cut_by(**kwargs)
         return r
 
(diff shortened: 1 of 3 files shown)


WenzDaniel:
Hej Joran, 

nice Idea. I have not checked your code yet, but one quick question. In your current version cuts would be the same data_kind as events would not they be? Maybe it is better to have a new "cut"-data_kind instead?

jorana:
> 
> 
> Hej Joran,
> 
> nice Idea. I have not checked your code yet, but one quick question. In your current version cuts would be the same data_kind as events would not they be? Maybe it is better to have a new "cut"-data_kind instead?

Hi Daniel,
Thanks for the suggestion! The datakind is derived from the dependency specified in the ```cut_by(self <datakind>)```, so in principle you could cut on peak level. I think it would make analysts lives harder if we create a new datatype that is not events as that is where most analysts will actually apply the cuts. That can get tricky if you want to load multiple datakinds as ```get_df``` (the function most analysts will use) will complain that this is not possible.

strax/plugin.py
```diff
@@ -847,3 +847,50 @@ def cleanup(self, iters, wait_for):
...
+        else:
+            raise NotImplementedError(f"{self.cut_name} does not have attribute 'cut_by'")
+
+        # Take shape of the first data_type like in strax.plugin
+        buff = list(kwargs.values())[0]
+
+        # Generate result buffer
+        r = np.zeros(len(buff), self.dtype)
+        r['time'] = buff['time']
+        r['endtime'] = buff['endtime']
```
WenzDaniel, line 890: This will only work if the data_kind has an endtime field. Which is not the case for the lower levels.

strax/plugin.py
```diff
@@ -847,3 +847,50 @@ def cleanup(self, iters, wait_for):
...
+            else:
+                _description = " ".join(_description.split("_"))
+            self.cut_description = _description
+
+    def infer_dtype(self):
+        dtype = [(self.cut_name, np.bool, self.cut_description)]
+        dtype = dtype + strax.time_fields
+        return dtype
+
+    def compute(self, **kwargs):
```
WenzDaniel, line 878: To get the start and endtime of your data you can do 

`compute(<data_kinds>, start, end)` 

See: https://github.com/AxFoundation/strax/blob/7c0641000301653adbf4144967b8a9cc18bf8103/strax/plugin.py#L448 and my comment below.

strax/plugin.py
```diff
@@ -847,3 +847,50 @@ def cleanup(self, iters, wait_for):
...
+        if not hasattr(self, 'cut_description'):
+            _description = _name
+            if 'cut_' not in _description:
+                _description = 'Cut by ' + _description
+            else:
+                _description = " ".join(_description.split("_"))
+            self.cut_description = _description
+
+    def infer_dtype(self):
+        dtype = [(self.cut_name, np.bool, self.cut_description)]
```
WenzDaniel, line 874: Does not it have to be np.bool_ ? https://numpy.org/doc/stable/user/basics.types.html

strax/plugin.py
jorana: thanks very good point!

strax/plugin.py
jorana: I will use strax.endtime as it is somewhat easier

strax/plugin.py
jorana: thanks, didn't know that, will change it!

jorana:
As @WenzDaniel just correctly pointed out the endtime needs to be computed when we are working with lower level plugins. Although it's not really the envisaged application it's easy enough to implement.

See now for example:
```
class CutRecord0(CutPlugin):
    depends_on = ('records',)
    provides = 'is_not_record0_cut'
    cut_description = "is record 0"
    cut_name ='is_not_record0_cut'
    __version__ = '0.0.0'
    
    def cut_by(self, records):
        return records['record_i'] != 0
```

jorana:
In 752f521 I've added a test mainly since I think I should learn how how to write tests. Also it's completely fine to add more tests.

strax/plugin.py
```diff
@@ -847,3 +847,51 @@ def cleanup(self, iters, wait_for):
...
+            cut_by = self.cut_by
+        else:
+            raise NotImplementedError(f"{self.cut_name} does not have attribute 'cut_by'")
+
+        # Take shape of the first data_type like in strax.plugin
+        buff = list(kwargs.values())[0]
+
+        # Generate result buffer
+        r = np.zeros(len(buff), self.dtype)
+        r['time'] = buff['time']
```
WenzDaniel, line 890: You have to add a test whether buff will be an empty. In that case buff['time'] will crash.

jorana: Actually, you can make copies of empty data. For example the line below does not fail:
```
strax.endtime(np.zeros(0, dtype=strax.time_fields))
```

jorana:
@ ACs this is ready for merge

tests/test_cut_plugin.py
```diff
@@ -0,0 +1,100 @@
...
+
+    st = strax.Context(storage=[])
+    st.register(ToBeCut)
+    st.register(CutSomething)
+
+    result = st.get_array(run_id='some_run',
+                          targets=strax.camel_to_snake(CutSomething.__name__))
+    correct_answer = np.sum(input_peaks[_dtype_name] > cut_threshold)
+    assert len(result) == len(input_peaks), "WTF??"
+    assert correct_answer == np.sum(result['cut_something']), "Cut plugin does not give boolean arrays correctly"
```
WenzDaniel, line 100: Maybe just add one additional test checking the endtime since we had these issues in peaks and you are computing some new endtimes here. Does not hurt much to have this little check. All the rest looks fine.

jorana: Thanks, this is done in 3673c28 !

PR #275, opened 2020-06-24
New hitlets data_kind

WenzDaniel:
Dear all, 
it has been quite some time since my last PR https://github.com/XENONnT/straxen/pull/86 for the nVETO plugins in straxen and many things have changed. Back then, Jelle and Joran asked me to change some of my code so that it would build up  more on already existing starx functions. The two major point were:

- Changing the function `get_hitlets_data` such that it is based on `record_links` and `overlapping_indices` rather than new custom made code.
- Adopting split_peaks such that it can be used for hitlets as well. 

In addition, based on my work in the analysis Team B and Fast Response Team I noticed, that many of the functions which will be used by the nVETO are also useful to very low level analyses for the TPC. Therefore, I decided to add a more generic data_kind called hitlets with a coresponding submodul to strax rather than having the functions next to the straxen plugins. Hitlets are an extension of the already existing hits which allow to fully characterize PMT signals  for SPE, DPE or lone_hitlet analyses or the operation of the neutron-veto. Beside these use cases I think also smaller TPC could benefit from this work. 

Not all functions have been implemented yet (e.g. the conditional_entropy) is still missing, but since this PR is rather large I decided to open it already now. Please let me know your thoughts and comments. Once all issues are solved I can proceed and update the corresponding plugins.


WenzDaniel:
I should add, I have just started to look into hypothesis test. But I would like to add some.

strax/processing/hitlets.py
```diff
@@ -0,0 +1,372 @@
...
+                print(ind, last_ri)
+                raise RuntimeError('Tried too often to find correct record_i.')
+
+
+# ----------------------
+# Hitlet properties:
+# ----------------------
+@export
+@numba.njit(cache=True, nogil=True)
+def get_fwxm(data, index_maximum, percentage=0.5):
```
jorana, line 308: Could you not drop the ```index_maximum``` and do a ```index_maximum=np.argmax(data)``` in the function itself

strax/dtypes.py
```diff
@@ -103,6 +103,48 @@ def record_dtype(samples_per_record=DEFAULT_RECORD_LENGTH):
...
 
+def hitlet_dtype(n_widths=11, n_sample=0):
+    """
+    Hitlet dtype same as peaklet or peak dtype but for hit-kind of
+    objects.
+
+    Note:
+        If n_samples is set to zero returns an array without "data"-field.
+    """
+    hitlet_dtype = interval_dtype + [
```
jorana, line 114: Don't you overwrite the function ```hitlet_dtype``` here

strax/dtypes.py
```diff
@@ -103,6 +103,48 @@ def record_dtype(samples_per_record=DEFAULT_RECORD_LENGTH):
...
+        (('Original length of the hit interval in samples',
+          'hit_length'), np.int32),
+        (('Total hit area',
+          'area'), np.float32),
```
jorana, line 118: perhaps nice to give the units

strax/processing/hitlets.py
```diff
@@ -0,0 +1,372 @@
...
+        offset += 1
+        if offset == len(buffer):
+            yield offset
+            offset = 0
+    yield offset
+
+
+@export
+@numba.njit(nogil=True, cache=True)
+def refresh_hit_to_hitlets(hits, hitlets):
```
jorana, line 113: Where is this used (I cannot find it) I'm not sure where this would be needed.

strax/dtypes.py
WenzDaniel: No, since `hitlet_dtype` is just defined locally within `hitlet_dtype`, but pycharm is complaining about the style. So I will change it, thanks

```

File: strax/processing/hitlets.py (hunk 1 above)
Comment (WenzDaniel) [reply strax/processing/hitlets.py:346]: Sorry, the export should not have been there. This function is only called within hitlet_properties

File: strax/processing/hitlets.py (hunk 4 above)
Comment (WenzDaniel) [reply strax/processing/hitlets.py:111]: See comment below.

**Comment by WenzDaniel** [comment]
Dear all I updated the pull request for the nveto-plugins https://github.com/XENONnT/straxen/pull/86. In there you will find the use cases for all functions in this PR. Additionally I wrote an introduction notebook which explains the data_flow for the nveto plugins and all new functions which are called inside (including the once in this PR).  

https://github.com/XENONnT/analysiscode/blob/master/nveto/HdM_test_setup/HdM_analysis_introduction.ipynb

In addition the notebook explain how some processed nveto test data can be accessed. I hope this will give the reviewers a better understanding of functions reasoning and help nveto people to start working with the data. Since some of the test done in https://github.com/XENONnT/straxen/pull/86 depend on the functions in this PR they fail.

**Comment by jorana** [comment]
@WenzDaniel what is the state of this PR? Do you want a re-review?

File: strax/processing/hitlets.py (hunk 1 above)
Comment (jorana) [reply strax/processing/hitlets.py:345]: sure but it's easy if people (or you) want to use it via strax.processing.hitlets.fet_fwxm. If the arguments can be made simpler it allows for easier maintenance as you can also use the function elsewhere

File: strax/dtypes.py [hunk 5]
```diff
@@ -103,6 +103,47 @@ def record_dtype(samples_per_record=DEFAULT_RECORD_LENGTH):
...
+        (('Left edge of the FWHM in ns (minus "time")',
+          'left'), np.float32),
+        (('FWTM of the PMT pulse in ns', 'fwtm'),
+         np.float32),
+        (('Left edge of the FWTM in ns (minus "time")',
+          'low_left'), np.float32),
+        (('Fragment index in which hit was found (hitlets can be in more than 1 fragment)',
+          'record_i'), np.int32)]
+
+    if n_sample:
```
jorana, line 140: Do you foresee any problems if we change this as it won't be reflected in the lineage hash while it might render your data incompatible with what you expect to load as the dtype doesn't match.

strax/processing/hitlets.py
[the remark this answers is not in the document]
jorana: Otherwise please make explicit that you define percentage [0,1] in the docstring

strax/dtypes.py
WenzDaniel: Good catch have not thought about this. I will change it.

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+    rlink = strax.record_links(records)
+    for h in hitlets:
+        data, start_time = get_single_hitlet_data(h, records, *rlink)
+        h['length'] = len(data)
+        h['data'][:len(data)] = data * to_pe[h['channel']]
+        h['time'] = start_time
+        h['area'] = np.sum(data * to_pe[h['channel']])
+
+
+TRIAL_COUNTER_NEIGHBORING_RECORDS = 100
```
jorana, line 159: perhaps set this at the top of the file

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+@export
+@numba.njit(nogil=True, cache=True)
+def get_single_hitlet_data(hitlet, records, prev_r, next_r):
+    """
+    Function which gets the data of a single hit or hitlet according
+    to the specified time and length of the object.
```
jorana, line 165: I'm not sure if I follow the docstring, especially given the arguments

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+
+        h['fwhm'] = width
+        h['left'] = left_edge
+        h['low_left'] = left_edge_low
+        h['fwtm'] = width_low
+
+
+NO_FWXM = -42
+@numba.njit(cache=True, nogil=True)
+def get_fwxm(hitlet, fraction=0.5):
```
jorana, line 364: probably can @ export this as it might be usefull elsewhere too

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+        The function searches for the last sample below and above the
+        specified height level on the left and right hand side of the
+        maximum. When the samples are found the width is estimated based
+        upon a linear interpolation between the respective samples. In
+        case, that the samples cannot be found for either one of the
+        sides the corresponding outer most bin edges are used: left 0;
+        right last sample + 1.
+    """
+    data = hitlet['data'][:hitlet['length']]
+    max_val = hitlet['amplitude'] * fraction
```
jorana, line 382: For peaks this would be area right? perhaps we can make it an kwarg such that it's also sable for peaks

strax/processing/peak_splitting.py
```diff
@@ -112,16 +153,20 @@ def _split_peaks(split_finder, peaks, orig_dt, is_split, min_area,
...
             for split_i, bonus_output in split_finder(
                     w, p['dt'], p_i, *args_options):
+                
+                # This is a bit odd here. Due tp the specific_outputs we have to get r 
+                # although we may not need it at all, but I do not see any nice way around
+                # this.
+                r = new_peaks[offset]
+                specific_output(r, p, split_i, bonus_output)
                 if split_i == NO_MORE_SPLITS:
-                    p['max_goodness_of_split'] = bonus_output
```
jorana, line 119: why do you remove this line?

strax/processing/peak_splitting.py
```diff
@@ -130,8 +175,6 @@ def _split_peaks(split_finder, peaks, orig_dt, is_split, min_area,
                 r['dt'] = orig_dt
                 r['length'] = (split_i - prev_split_i) * p['dt'] / orig_dt
 
-                r['max_gap'] = -1    # Too lazy to compute this
```
jorana, line 133: why do you remove this line too?

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+
+    # Buffer for concat_overlapping_hits, if specified in 
+    # _concat_overlapping_hits numba crashes.
+    last_hit_in_channel = np.zeros(nchannels,
+                                   dtype=(hit_dtype
+                                          + [(('End time of the interval (ns since unix epoch)',
+                                               'endtime'), np.int64)]))
+
+    if len(hits):
+        hits = _concat_overlapping_hits(hits, extensions, first_channel, last_hit_in_channel)
```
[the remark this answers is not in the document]
WenzDaniel: It is a growing result function.

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+                             _result_buffer=None):
+    buffer = _result_buffer
+    offset = 0
+
+    le, re = extensions
+    dt = hits['dt'][0]
+    assert np.all(hits['dt'] == dt), 'All hits must have the same dt!'
+
+    for h in hits:
+        st = h['time'] - int(le * h['dt'])
```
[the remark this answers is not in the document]
WenzDaniel: I think theoretically yes, you can get this edge case if you only apply this function to the data. But only if the chunk start is equivalent with the start of a record and you find a hit within the very first sample of this record. And this should basically never happen. 
Further get_hitlets_data, will correct the start time in case a hit is extended into a not recorded region (same for et).

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+        et = strax.endtime(h) + int(re * h['dt'])
+        hc = h['channel']
+        r_i = h['record_i']
+
+        lhc = last_hit_in_channel[hc - first_channel]
+        # Have not found any hit in this channel yet:
+        if lhc['time'] == 0:
```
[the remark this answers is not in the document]
WenzDaniel: lhc is sort of buffer which initialized as an empty array.

strax/processing/hitlets.py
```diff
@@ -0,0 +1,526 @@
...
+        offset += 1
+        if offset == len(buffer):
+            yield offset
+            offset = 0
+    yield offset
+
+
+@export
+@numba.njit(nogil=True, cache=True)
+def refresh_hit_to_hitlets(hits, hitlets):
```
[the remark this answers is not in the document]
WenzDaniel: I am not sure if you can ask within numba if a specific numpy field exists. At least .dtype.names etc. is not working, but I can check.

strax/processing/hitlets.py
WenzDaniel: Okay

strax/processing/hitlets.py
WenzDaniel: Okay I will change it. My intention was to say something like: For hits you will only get the central part without LE/RE

strax/processing/hitlets.py
WenzDaniel: Sure never tried it on S1/S2.

strax/processing/hitlets.py
WenzDaniel: Okay no just checked peaks do not have a "height" field.

strax/processing/peak_splitting.py
WenzDaniel: I had to put this part into a separate function since hitlets do not have those fields. See line 89 - 105

strax/processing/hitlets.py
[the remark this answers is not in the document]
jorana: > And this should basically never happen.

why not? Also, get_hitlets_data doesn't do endtimes right

strax/processing/hitlets.py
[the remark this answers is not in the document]
jorana: can you also not pass on the dtype names as a typing.List?

strax/processing/hitlets.py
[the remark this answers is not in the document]
WenzDaniel: Because the digitizers and hitfinder should share the same thresholds. Otherwise Acceptance computation becomes odd. get_hitlet_data will change the `length` field according to the length of the found data.

I do not see any feasible way to change this.

strax/processing/hitlets.py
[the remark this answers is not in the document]
WenzDaniel: I never worked with numba lists, but I can give it a try.

strax/processing/hitlets.py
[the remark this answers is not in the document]
WenzDaniel: Never mind I just had an idea. I could require to pass also the chunk start and end time as additional arguments to the function.

strax/processing/hitlets.py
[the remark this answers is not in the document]
jorana: a numpy array of strings should also work I expect it was more about the wapping in a non-numba function to extract the dtypes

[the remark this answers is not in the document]
WenzDaniel: I just remembered yesterday, that I think tied this already.  Since I was not sure I just gave it another shot. No it does not work. My guess would be that under the hood of numba it does not work for the same reason as array.dtype.names does not.

PR #277, opened 2020-07-06
Allows for plugin-specific chunks sizes

darrylmasson:
This changes the default uncompressed chunk size from 100 MB (specified in an obscure, undocumented location) to a plugin-definable value, with new default 200 MB. This value gets stored with the metadata, but if it's missing the default is assumed. This should affect writing data much more than reading it, but the hope is that 200 MB uncompressed is a more acceptable trade between fewer, larger files (easier on the bookkeeping) and more, smaller files (easier on the end user).

### DIFF
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -5,8 +5,9 @@ import numba
 
 import strax
 export, __all__ = strax.exporter()
+__all__ += ['default_chunk_size']
 
-@export
+
 default_chunk_size = int(2e8)
 
 @export
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -5,10 +5,10 @@ import numba
 
 import strax
 export, __all__ = strax.exporter()
-__all__ += ['default_chunk_size']
+__all__ += ['default_chunk_size_mb']
 
 
-default_chunk_size = int(2e8)
+default_chunk_size_mb = 200
 
 @export
 class Chunk:
@@ -25,7 +25,7 @@ class Chunk:
     end: int
 
     data: np.ndarray
-    target_size: int
+    target_size_mb: int
 
     def __init__(self,
                  *,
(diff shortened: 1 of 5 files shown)


darrylmasson:
It's worth pointing out that having this as a user-specifiable value means there is considerable misuse potential, and so maintainers should make sure that instanced Plugins don't set this to absurd values.

jorana:
@darrylmasson as Travis is complaining, you want to export via __all__. See https://github.com/xenonnt/straxen/blob/master/straxen/common.py#L20 for an example

darrylmasson:
Let's see if this fixes it

strax/chunk.py
```diff
@@ -31,7 +34,8 @@ def __init__(self,
                  run_id,
                  start,
                  end,
-                 data):
+                 data,
+                 target_size=default_chunk_size):
```
jorana, line 39: Can we be smart about the naming such that no-one accidentally assumes that the target_size is in MB (which would result in saying target_size=200 which would then result in infinite files in the ./strax_data directory where one tests a plugin prior to pushing it to github).

For example, how about target_size_mb (and multiply later by 1000).

darrylmasson: We could, or we explicitly document it as bytes. We should not be held accountable if someone decides that they don't need to familiarize themselves with the documentation before changing attributes.

jorana: Agree, people should read the documentation. However, all numbers that can be used here will be in x MB so I think having it in some int between 10-3000 would make it quite readable and easier to spot if someone is doing something that he/she shouldn't. Another example is if you cat the metadata. Having a 1000 vs 1000000000 is somewhat easier to  read.

darrylmasson: Valid point

jorana:
@ ACs this is ready for merge

jorana:
Moved the # Todo to https://github.com/AxFoundation/strax/issues/291.

I'm not sure if I agree with CodeFactor though that this makes the code better...

PR #281, opened 2020-07-09
Correctly check endtimes in chunk.py

jorana:
**Problem**
We observed that although the data is always sorted by (start)time, this is not per se true for the endtime. Therefore the last record may not have the latest endtime. We need to check this and not proceed if we observe such a thing. 

**Example**
As Daniel illustrated nicely. Image we have to following data where `-` means data, `_` means no data, and `|` means start and stop.
```
ch0: |---|__
ch1: _|-|___
```
Here the signal in ch0 starts before the signal in ch1 while the end of ch0 is __later__ than ch1. Hence we need to check both channels.

**Solution**
We used to look only at the last sample only but we need to look into the endtimes better to assert that we are not having any data sticking out of the chunk. This is the safest way to get around this (that is look at all the data in the chunk).

**Alternative**
Depending on the datarate one might also argue that for example we look at the last 100 samples :
```
data_ends_at = strax.endtime(self.data[-100:]).max()
```

**Straxen**
A similar PR in straxen will actually fix the problem:
https://github.com/XENONnT/straxen/pull/146

### DIFF
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -65,7 +65,7 @@ class Chunk:
 
         if len(self.data):
             data_starts_at = self.data[0]['time']
-            data_ends_at = strax.endtime(self.data[-1])
+            data_ends_at = strax.endtime(self.data).max()
 
             if data_starts_at < self.start:
                 raise ValueError(f"Attempt to create chunk {self} "
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -65,7 +65,7 @@ class Chunk:
 
         if len(self.data):
             data_starts_at = self.data[0]['time']
-            data_ends_at = strax.endtime(self.data).max()
+            data_ends_at = strax.endtime(self.data[-500:]).max()
 
             if data_starts_at < self.start:
                 raise ValueError(f"Attempt to create chunk {self} "
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -65,6 +65,7 @@ class Chunk:
 
         if len(self.data):
             data_starts_at = self.data[0]['time']
+            # Check the last 500 samples (arbitrary number) as sanity check
             data_ends_at = strax.endtime(self.data[-500:]).max()
 
             if data_starts_at < self.start:


strax/chunk.py
```diff
@@ -65,7 +65,7 @@ def __init__(self,
 
         if len(self.data):
             data_starts_at = self.data[0]['time']
-            data_ends_at = strax.endtime(self.data[-1])
+            data_ends_at = strax.endtime(self.data).max()
```
WenzDaniel, line 68: Can we apply here the same logic as for raw_records, but I am afraid not since peaks have no fixed length?

darrylmasson: For records we can make arguments about the number of channels because they overlap in time, but peaks don't. The second-last peak has to end before the last peak begins.

jorana: For peaks this should really be peanuts to compute as there are so few (relative to raw_records)

WenzDaniel: I am wondering if our improved approach in https://github.com/XENONnT/straxen/pull/146 becomes obsolete with this change. Was this change included in your timing benchmark?

jorana: No I didn't include it in the benchmark. 

Further, the computation here is different from the one in https://github.com/XENONnT/straxen/pull/146. Here this number is only used to make the ValueError (such that we don't make mistakes like https://github.com/XENONnT/straxen/pull/146) but hit the ValueError if we do try.

WenzDaniel: I agree, my concern was also more regarding the computation time. Either way I guess we do not have much other options.

jorana: Sure, I guess it will never be worse than for raw_records but it'd be good to try.

WenzDaniel: Maybe I am wrong but this check would be applied for all data_kinds hence you add again your benchmarked  84.7 ms/chunk for raw_record and record etc at least. Is not it?

jorana: Okay, so this is the deal:
I did a run with 1.6+0.4 s chunking (ergo many checks).

**Before update**
3211346 MainThread root bootstrax (eb0.xenon.local) finished run 8600 in 3204 seconds

**After update**
3477742 MainThread root bootstrax (eb0.xenon.local) finished run 8600 in 3471 seconds

I think this is an unacceptable loss of performance for a check that shouldn't be hit for properly constructed chunks. I'd argue that for a little less safety we might want to limit looking back the last say _500_ samples? This would be an arbitrary number but we have been running so long with just 1 sample I would be surprised if we ever ever hit this problem at sample [:-501]. 

I'm running this test again to see how reproducible these results are and will edit these numbers into this comment.

PR #284, opened 2020-07-14
Fix endtimes in peaks merging

jorana:
For peaks with an endtime we need to recompute those numbers as otherwise we end up with a 0 as an `endtime `and `strax.endtime(peak)` will simply compute the endtime.

For example I tried merging peaks like these:
afbeelding

Which gave me this obviously wrong peak:
afbeelding

Maybe we should also change `strax.endtime `to not return `0`'s unless we consider going back in time to the 70s.

**Why haven't we seen this before**
Endtime is a field for peaks and higher while this code is only used for `peaklets`. Therefore we haven't run into this before but I foresee that solving issues that might arise might appear very strange if we want to remerge peaks at some point with strax.


jorana:
@WenzDaniel can you approve and give OK to ACs?

PR #286, opened 2020-07-15
Changed dtype of saturated_channel

WenzDaniel:
I changed the dtype of saturated_channel from np.int16 to np.bool_ since it is a boolean https://github.com/AxFoundation/strax/blob/6d0efb9ee4bb144c5528804a2adc0f7d34a29948/strax/processing/peak_building.py#L212. With this change the memory consumption of peaklets is reduced by 13% which is huge since it is for free. 

MWE:

```
run_id = blablub
data = st.get_array(run_id, 'peaklets', seconds_range=(0, 60))

import numpy
dtype = [(('Start time since unix epoch [ns]', 'time'), numpy.int64),
 (('Length of the interval in samples', 'length'), numpy.int32),
 (('Width of one sample [ns]', 'dt'), numpy.int16),
 (('Channel/PMT number', 'channel'), numpy.int16),
 (('Classification of the peak(let)', 'type'), numpy.int8),
 (('Integral across channels [PE]', 'area'), numpy.float32),
 (('Integral per channel [PE]', 'area_per_channel'), numpy.float32, 494),
 (('Number of hits contributing at least one sample to the peak ', 'n_hits'),
  numpy.int32),
 (('Waveform data in PE/sample (not PE/ns!)', 'data'), numpy.float32, 200),
 (('Peak widths in range of central area fraction [ns]', 'width'),
  numpy.float32,
  11),
 (('Peak widths: time between nth and 5th area decile [ns]',
   'area_decile_from_midpoint'),
  numpy.float32,
  11),
 (('Does the channel reach ADC saturation?', 'saturated_channel'),
  numpy.bool_,
  494),
 (('Total number of saturated channels', 'n_saturated_channels'), numpy.int16),
 (('Hits within tight range of mean', 'tight_coincidence'), numpy.int16),
 (('Largest gap between hits inside peak [ns]', 'max_gap'), numpy.int32),
 (('Maximum interior goodness of split', 'max_goodness_of_split'),
  numpy.float32)]

data2 = np.zeros(len(data), dtype)
for name in data.dtype.names:
    data2[name] = data[name] 

2data.nbytes/data.nbytes
```


WenzDaniel:
There seems to be numba bug with np.bool_ in multidimensional structured arrays. np.int8 seems to work and gives the same result

jorana:
> 
> 
> Yes, nice catch. Let's make sure to bump the peaklets version in straxen when merging this.

Good point, that might require reprocessing of all the >peaklets 1T data too (something to keep in mind).

jorana:
@ ACs this is ready for merge

PR #289, opened 2020-07-20
Fix error in forbid_creation_of

jorana:
**Problem**
We forbid the creation of records if only raw_records is in the forbid_creation of. If we compare stings, 'records' is in 'raw_records'. 

afbeelding
yields the following error: 
afbeelding
This is not entirely correct, we just specified we couldn't create raw_records. 

**Solution/test**
afbeelding
This starts making 'records'

**Additional**
I don't think we want to allow users to create 'records' in straxen. As such I think we should specify that we forbid creation of ('raw_records', 'records') on line https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L68.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -564,6 +564,7 @@ class Context:
                 del plugins[d]
             else:
                 # Data not found anywhere. We will be computing it.
+                self._check_forbidden()
                 if (time_range is not None
                         and plugins[d].save_when != strax.SaveWhen.NEVER):
                     # While the data type providing the time information is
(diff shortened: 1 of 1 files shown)


petergaemers:
Good catch, saw it happen before but never looked at it.
Since I don't have write acces I guess my approval doesn't mean much

JelleAalbers:
Maybe not according to Github but your opinion definitely matters Peter :-)

strax/context.py
```diff
@@ -1152,6 +1153,16 @@ def is_stored(self, run_id, target, **kwargs):
...
 
+    def _check_forbidden(self):
+        """Check that the forbid_creaction_of config is of tuple type.
+        Otherwise, try to make it a tuple"""
+        forbid_creation_of = self.context_config['forbid_creation_of']
+        if not isinstance(forbid_creation_of, tuple):
+            if not isinstance(forbid_creation_of, str):
+                raise ValueError(f'forbid_creation_of should be of type tuple not'
+                                 f'{type(forbid_creation_of)}')
+            self.context_config['forbid_creation_of'] = ((forbid_creation_of), )
```
WenzDaniel, line 1164: Sorry to be the bad guy here, but I think we should use https://github.com/AxFoundation/strax/blob/16650cc50f8fa3dd39ef28621e95051656fc644c/strax/utils.py#L226 instead. 

This would make the ValueError obsolete as well.

jorana: Superb, that is much cleaner! Thanks Daniel! I knew that there was a function somewhere but I thought it was some ty.Union-function. 

I'll leave it in the _check_forbidden to have a traceback that is very clear where one would be wrong.

jorana:
@ ACs ready for merge

PR #294, opened 2020-07-23
Auto compression selection

jmosbacher:
DO NOT MERGE! not fully tested yet. PR created to discuss implementation.
This PR adds a field to all compressors called "conditions", which is a list of functions that accept an array and return true or false. All conditions must be met (return true) for a compression method to be used. 
The FileSaver._save_chunk method has been adapted to iterate over the compressor values in the metadata and use the first one that passes all conditions. The compressor field of the metadata is passed to the to_str_tuple utility function for backwards compatibility.
This was implemented specifically to allow using blosc in all cases except the chunks above 2GB in size, but can be used also to select compression based on arbitrary criteria.
An alternative implementation I thought about was looping over the compression options with a try, except block. The advantage of a try except block would be that you get resilience against unknown compression issues not just the ones you anticipated. The downside would be speed (catching an exception is slower than boolean check) and having an ugly try except block in your code (personal preference).


strax/storage/files.py
```diff
@@ -281,14 +281,20 @@ def _chunk_filename(self, chunk_info):
 
     def _save_chunk(self, data, chunk_info, executor=None):
         filename = self._chunk_filename(chunk_info)
-
         fn = os.path.join(self.tempdirname, filename)
-        kwargs = dict(data=data, compressor=self.md['compressor'])
+        
+        for compressor in strax.to_str_tuple(self.md['compressor']):
```
darrylmasson, line 286: Where does this quantity get set? It might be prudent to use `get('compressor', some_default_value)` instead. Also, what type are we expecting this to take? A specific compressor? A list of desired compressors? I thought the initial idea was to default to blosc unless it's too large, in which case switch to lz4.

strax/io.py
```diff
@@ -14,20 +14,33 @@
...
+        conditions=BASE_CONDITIONS),
     blosc=dict(
         compress=partial(blosc.compress, shuffle=False),
-        decompress=blosc.decompress),
-    lz4=dict(compress=lz4.compress, decompress=lz4.decompress)
+        decompress=blosc.decompress,
+        conditions=BLOSC_CONDITIONS),
+    lz4=dict(compress=lz4.compress,
+     decompress=lz4.decompress, 
+     conditions=BASE_CONDITIONS)
```
darrylmasson, line 38: Missing some whitespace here, looks like. Different tab sizes? Use spaces instead anyway, python complains.

tunnell:
@sophiaandaloro can you share the info that @shuaichengli0428 produced on compression?  Might be useful since he found that LZMA is slow but might be good for long term storage.

strax/storage/files.py
jmosbacher: Sorry, somehow i missed the activity on this thread. As you can see the current implementation assumes there is a compressor set for each plugin. The default value is already set to "blosc" since the compressor property of the Plugin base class is "blosc".  As I mentioned in my initial comment, the type is a list of compression names or a single name. It will use the first one in the list that can be used. if none of the compressions in the list can be used then it will throw an error. If you change the compression property of the Plugin class to ["blosc", "lz4"] you would get the desired behavior you want.

strax/io.py
jmosbacher: thats super weird, spaces shouldnt matter inside a dictionary definition. what error did you get? or are you talking about a linter error?

jorana:
@jmosbacher , does it make sense to close these stale PRs or stash them to your own fork? We can always reopen?

jmosbacher:
yup

Issue #295, opened 2020-07-23
Progressbar raises error


jorana:
There is a small error in progressbar when you:
```
peaks = st.get_array(run_id, targets = ('peak_basics'), seconds_range=(0,1),
                    progress_bar = False)
```
it raises:
```python
    858         try:
    859             with contextlib.ExitStack() as stack:
--> 860                 if start_time != 0 and end_time != float('inf'):
    861                     # Get initial time
    862                     pbar = stack.enter_context(tqdm(total=1, postfix=post_fix, bar_format=bar_format))

UnboundLocalError: local variable 'start_time' referenced before assignment

```

WenzDaniel:
Solved

PR #296, opened 2020-07-23
Small progress-bar bug fix

jorana:
Fix issue #295 where the progress-bar raises an error if disabled as start_time is not defined.

Relates to https://github.com/AxFoundation/strax/pull/276 where the progress bar was introduced.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -857,7 +857,7 @@ class Context:
 
         try:
             with contextlib.ExitStack() as stack:
-                if start_time != 0 and end_time != float('inf'):
+                if progress_bar and (start_time != 0 and end_time != float('inf')):
                     # Get initial time
                     pbar = stack.enter_context(tqdm(total=1, postfix=post_fix, bar_format=bar_format))
                     last_time = pbar.last_print_t
--- a/strax/context.py
+++ b/strax/context.py
@@ -843,7 +843,9 @@ class Context:
                         chunks = self.get_meta(run_id, t)['chunks']  
                         start_time = max(start_time, chunks[0]['start'])
                         end_time = min(end_time, chunks[-1]['end'])
-                    except (strax.DataNotAvailable, KeyError):
+                    except (strax.storage.common.DataCorrupted,
+                            strax.DataNotAvailable, 
+                            KeyError):
                         # Maybe at least one target had some metadata.
                         start_time = max(start_time, 0)
                         end_time = min(end_time, float('inf'))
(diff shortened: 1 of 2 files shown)


WenzDaniel:
Could catch, sorry for that.

jorana:
I pushed e5b5ea9 since e1d7878 was failing for some strange reason on one of the builds.

jorana:
@WenzDaniel I ran into the following error on the ebs. I guess progress-bar doesn't really help if we make data from scratch. Maybe it's overkill but I guess st.make will only be called in scripts where we don't have to have a progress bar. What do you think?

```python
11165 MainThread root Repeated failure on run 8746:5f19a2a44bd9657224dbf7ea: Strax exited with exit code 1. Exception info: Traceback (most recent call last):
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1041, in run_strax
    st_make()
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1038, in st_make
    max_workers=process_mode['cores'])
  File "/home/xedaq/software/strax/strax/context.py", line 992, in make
    save=save, max_workers=max_workers, **kwargs):
  File "/home/xedaq/software/strax/strax/context.py", line 843, in get_iter
    chunks = self.get_meta(run_id, t)['chunks']
  File "/home/xedaq/software/strax/strax/context.py", line 1155, in get_meta
    return sf.get_metadata(key, **self._find_options)
  File "/home/xedaq/software/strax/strax/storage/common.py", line 180, in get_metadata
    return self._get_backend(backend).get_metadata(backend_key)
  File "/home/xedaq/software/strax/strax/storage/files.py", line 215, in get_metadata
    raise strax.DataCorrupted(f"Data in {dirname} has no metadata")
strax.storage.common.DataCorrupted: Data in /data/xenonnt_processed/008746-event_info-6t4ptr7h3p has no metadata

 11167 MainThread root bootstrax (eb5.xenon.local) finished run 8746 in 10 seconds
```

strax/context.py
```diff
@@ -857,7 +857,7 @@ def get_iter(self, run_id: str,
 
         try:
             with contextlib.ExitStack() as stack:
-                if start_time != 0 and end_time != float('inf'):
+                if progress_bar and (start_time != 0 and end_time != float('inf')):
```
WenzDaniel, line 862: Sorry, just saw it, could we change this. Instead of adding progress_bar here, we move the definition of start_time and end_time from line 838 and 839 to line 830.

WenzDaniel:
Wait what I do not understand why you have to hardcode it? Could not you simply add it to the kwargs of make https://github.com/AxFoundation/strax/blob/d59ee015dfd4964036cafeaf12d8ec25ba7a157f/strax/context.py#L973 and set `progress_bar=False`?

PR #297, opened 2020-07-28
Add "child" options for inherited plugins

WenzDaniel:
In this PR draft I try to modify the options handling such that we can specify some child plugins which are inherited from already existing once.


WenzDaniel:
I made a notebook which shows all important feature including some tests. The notebook can be found here: https://github.com/XENONnT/analysiscode/blob/master/StraxTests/ChangeStraxOptions.ipynb

WenzDaniel:
> 
> 
> Just briefly skimming though the code it all looks fine.
> 
> The notebook looks great! Can you extract from this:
> 
>     * Some concise documentation and add it to the plugins section?

Good point I will add it.

>     * Extract a few tests. 

I am already on it.

>I must admit I don't really understand the assertions statements you make e.g. for the channel but I must have looked at it to shortly.

Ah okay sorry. Maybe I should have explained it a bit more explicitly. Just ignore the field names. I thought the easiest way to check whether settings are propagated/changed/kept correctly is  just by putting their values an array and check if the array has in the end the correct values. I just had chosen arbitrarily in which fields I put the values.

strax/context.py
```diff
@@ -1078,7 +1121,7 @@ def function(arr):
             if store_first_for_others and not seen_data and len(data):
                 # Store the first value we see for the non-accumulated fields
                 for name in data.dtype.names:
-                    if name in fields:
+                    if name not in fields:
```
WenzDaniel, line 1124: Dear reviewer, I also caught a little but in accumulate which I fixed. Since it is a trivial change I did not want to open a separate PR.

strax/context.py
```diff
@@ -394,9 +394,34 @@ def _set_plugin_config(self, p, run_id, tolerant=True):
...
+                        v = config[k]
+                        kparent = k[:-len(p.child_ends_with)]
+
+                        if is_parent_reg:
+                            mes = f'Option {kparent} is not taken by parent plugin.'
+                            assert kparent in parent_class.takes_config.keys(), mes
+
+                        p.config[kparent] = v
+                    else:
+                        warnings.warn(f'You specified plugin {p.__class__.__name__} as a child plugin.'
```
jorana, line 420: Would a ValueError also be appropriate? I imagine that if people do this kind of development they should *not* miss this message to prevent people from writing messy plugins

tests/test_child_plugins.py
```diff
@@ -0,0 +1,236 @@
...
+        # Unchanged values and common values::
+        assert np.all(parent['time'] == child['time']), 'Time field is different for parent and child'
+        assert np.all(parent['length'] == child['length']), 'Length field is different for parent and child'
+        assert np.all(parent['dt'] == child['dt']), 'dt field is different for parent and child'
+
+        m = np.all(parent['max_gap'] == child['max_gap'])
+        assert m, 'parent_unique_option is not equal in parent and child'
+
+        # Values which should have changed:
+        assert np.all(parent['area'] == 2), 'by_child_overwrite_option has changed the value of the parent'
```
jorana, line 56: You really went all the way for these tests! I haven't looked at them in great detail but it looks like you put a lot of effort in it!

Perhaps an idea for the future is to set these numbers above in the script (as a dict?) as if one ever changes a number because we ran into a new testcase we don't have to remember (or ctrl+f the number '2' in the tests to look for) all the places were we have to change the number. Usually you test many numbers (e.g. what if its 0 instead of 2) but I don't feel this is an essential perquisite here.

It seems quite unlikely though that one will change these numbers so I think this will do!

strax/context.py
WenzDaniel: I had the same idea in the beginning and learned afterwards in the hard way that this is not a good choice. Imagine you would like to introduce a new child-only-option. Most likely you will call it with the same ending as the other options. Hence since it is a child-only-option you wont have any corresponding parent which will raise then this error. This is why I changed it in a warning. 
I thought a bit about it, but I was not able to find a more elegant solution for this problem.

tests/test_child_plugins.py
WenzDaniel: Agreed, it has become a bit messy. I will try to check whether I can make it a bit cleaner.

strax/context.py
jorana: Hmm yes I see, but, on the other hand, you are changing the option to be a child_only option right so you can also expect someone to change the naming. The thing I worry about is that if someone is making a new plugin and you are not checking it out locally to verify everything works you won't hit this warning whereas if you make it an error you'd instantly fail travis. 

Later you might be stuck with something that throws errors at you all the time but you don't feel like changing it because it would require going back to some low level data.

tests/test_child_plugins.py
WenzDaniel: Added your suggestion of a default config dict

jorana:
I think this is ready for merging. Go ahead if you agree

petergaemers:
Finally tested everything. At first I worried if you'd first make _he and then regular data it would keep the he config and make a mess of everything. This doesn't happen so everything is nice :)

jorana:
@WenzDaniel can you add the MWE to the top description of this PR.

PR #300, opened 2020-07-31
Add a backend for rucio

SidAhmedMa:
Hello everyone, given the particular structure of rucio, which for each file matches a sub-sub-folder, it was mandatory to create a new backend to manage this new structure,

In the new structure, we use strax.rucio as the backend and the database frontend, you will find a notebook below which shows the tests I performed.

Notebook: /dali/lgrandi/sahmedma/rucio-backend-test.ipynb


strax/storage/rucio.py
```diff
@@ -0,0 +1,41 @@
...
+        #print('yes')
+        fn = rucio_path(chunk_info['filename'], dirname)
+        return strax.load_file(fn, dtype=dtype, compressor=compressor)
+
+    def _saver(self, dirname, metadata):
+        raise NotImplementedError(
+            "Cannot save directly into rucio, upload with admix instead")
+
+
+def rucio_path(filename, dirname):
```
jorana, line 34: This bit is a bit hard to follow without comments (sorry I'm not an expert in the rucio naming convention). What do the paths look like and are we sure its always these hard-coded conversions?

strax/storage/rucio.py
```diff
@@ -0,0 +1,41 @@
...
+    root_path ='/dali/lgrandi/rucio'
```
jorana, line 35: We shouldn't hard-code it like this as this might change to e.g. '/dali/lgrandi/xenonnt/rucio' I guess you want to get the info from the dirname.

jorana:
Thanks Sid, can you address the comments and/or provide some more info to this PR (e.g. the rucio naming convention would be useful)

rucio.py
```diff
@@ -0,0 +1,41 @@
+import json
```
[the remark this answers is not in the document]
SidAhmedMa: Done, thanks for the remark

strax/storage/rucio.py
```diff
@@ -0,0 +1,41 @@
...
+    def get_metadata(self, dirname, **kwargs):
+        dirname = str(dirname)
+        prefix = dirname_to_prefix(dirname)
+        metadata_json = f'{prefix}-metadata.json'
+        fn = rucio_path(metadata_json, dirname)
+        with open(fn, mode='r') as f:
+            return json.loads(f.read())
+
+    def _read_chunk(self, dirname, chunk_info, dtype, compressor):
+        #print('yes')
```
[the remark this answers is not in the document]
SidAhmedMa: Sorry 😅

strax/storage/rucio.py
SidAhmedMa: Te path look like ('/dali/lgrandi/rucio/xnt_008710/1f/fb/peaklets-b7dgmtzaef-000047'), It is a hard-coded convention on the Rucio code, you can see it here https://github.com/rucio/rucio/blob/671fe6253981eb632aae3c9ddfe54eb83e63fd1a/lib/rucio/rse/protocols/protocol.py#L114

strax/storage/rucio.py
SidAhmedMa: Unfortunately, dirname does not include the root path 😕

strax/storage/rucio.py
```diff
@@ -0,0 +1,41 @@
...
+@export
+class rucio(strax.StorageBackend):
+    """Get data from a rucio directory
+    """
+
+    def get_metadata(self, dirname, **kwargs):
+        dirname = str(dirname)
+        prefix = dirname_to_prefix(dirname)
+        metadata_json = f'{prefix}-metadata.json'
+        fn = rucio_path(metadata_json, dirname)
```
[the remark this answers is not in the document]
SidAhmedMa: Yes, furthermore, I had much more outputs with this error than without 😅

SidAhmedMa:
Thank you very much @jorana for reviewing my pr, I hope that I answered to all your remarks 🙂

strax/storage/rucio.py
jorana: Hmm okay. So how about this:

You pass it on at initialization:
https://github.com/XENONnT/straxen/blob/5b762f32eec7d3fb8a1a96e31ae504fe8c8cf3cd/straxen/rundb.py#L110
and set the default to /dali/lgrandi/rucio at the init of the rundb here:
https://github.com/XENONnT/straxen/blob/5b762f32eec7d3fb8a1a96e31ae504fe8c8cf3cd/straxen/rundb.py#L42

You would also have to change these lines:
https://github.com/AxFoundation/strax/blob/6404ae5f5733d1f53301bf0976fcc9f637acbfe3/strax/storage/rucio.py#L15 (with adding a def __init___ (..) ) and
https://github.com/AxFoundation/strax/blob/6404ae5f5733d1f53301bf0976fcc9f637acbfe3/strax/storage/rucio.py#L25

If you want some help I can also commit this idea to prevent hardcodes?

strax/storage/rucio.py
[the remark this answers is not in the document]
jorana: Alright thanks for adding it

jorana:
If you'd use https://github.com/SidAhmedMa/strax/pull/1 you can prevent specifying the rootdir

jorana:
Hi Sid, I tested your implementation on dali and found some things about data not being available that are fixed by 52d6a99.

Since this is working and a really nice addition I'm going to merge this PR.

SidAhmedMa:
> Hi Sid, I tested your implementation on dali and found some things about data not being available that are fixed by [52d6a99](https://github.com/AxFoundation/strax/commit/52d6a99f52a19b69f0642dec2611d1d532a896dd).
> 
> Since this is working and a really nice addition I'm going to merge this PR.

I see I thought that the test to see if the directory exists or not was done upstream before going to the backend, but it is probably also useful to do a new test.

PR #301, opened 2020-08-07
Corrections Management Tool (CMT)

ahiguera-mx:
Corrections are an important part of the event building process, applying these corrections help us to remove detectors effects so we can have a “correct” measurement e.g corrected S2 area [PE]. Having a common tool that maintains and does bookkeeping can be beneficial. Thus, we are proposing a corrections management tool (CMT). Corrections are placed in DB, the format is a panda DataFrame with DatetimeIndex. Currently examples of electron lifetime and pmt gains corrections are implemented via cmt_services.py. 
This pull request should be transparent


WenzDaniel:
Hej Aaron, 

thank you very much for your work. I tend to agree with Joran, but before we can start the review procedure there is one odd thing with the git history of this PR. Currently it is impossible to review since it shows, that about 90 files are new although they are actually not. You have to fix this first. I am not an expert of git, but I think in this case the easiest way might be to fork a new branch from the strax master and copy your changes in this new branch.

Cheers,
Daniel

tunnell:
There's something weird going on with Git... I'd probably do a fresh checkout, then just modify what you need (like a squash merge).  It looks like everything was deleted then readded?

On strax v.s. straxen, that was actually my suggestion.  The correction tools themselves (interface to RunDB, interpolation routines, correction management, etc) are experiment agnostic, and any experiment using strax will need (nEXO doesn't have data yet :P).  The implementation of the specific corrections in XENONnT should then be in straxen, like lifetime etc.  strax won't know about lifetimes, or whatever, but will know how to get parameter X evaluated at datetime Y.

The core changes to strax should be minor while avoiding straxen bloat (if that is even avoidable).

ahiguera-mx:
Due to issues with git history this pull request is going to be close

PR #302, opened 2020-08-10
Create templates for strax

jorana:
When writing PRs in strax it might be good to add a template for both filing issues and adding PR. This can help keeping the pull requests in a similar format. It might also help authors to include things (like tests, documentation, examples) that would probably be requested by the reviewers.

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,0 +1,12 @@
+**What is the problem / what does the code in this PR do**
+
+**Can you briefly describe how it works?**
+
+**Can you give an minimum working example (or illustrate with a figure)?**
+
+Please include the following if applicable:
+  - Update the docstring(s)
+  - Update the documentation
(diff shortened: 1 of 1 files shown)


docs/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
+**What is the problem / what does the code in this PR do**
+
+**Can you briefly describe how it works?**
+
+**Can you give an minimum working example (or illustrate with a figure)?**
```
WenzDaniel, line 5: Sorry, but `an ...` -> `a minimum` and I think it is actually minimal working example.

docs/pull_request_template.md
```diff
@@ -0,0 +1,12 @@
...
+
```
WenzDaniel, line 6: Maybe add something like: 

```
**Make sure that all automated tests have passed before asking for a review.**
```

PR #304, opened 2020-08-12
Apply function to data

jorana:
Apply a function in `get_array` just before returning the data.

This way, you can make a last-minute correction (or apply a blinding cut in the context).

When one loads data you do either `st.get_array` or `st.get_df` (which then calls `st.get_array` for you). In the lines changed in this PR, you see that the results that is usually returned directly can now be changed by a function. Such a function should be a function that takes a single argument (the data) and does something to it. For example in https://github.com/XENONnT/straxen/pull/166 it remaps some channels but this could also be used to apply a blinding cut.

As written in the Strax.Option a function should have the  form of:

```python
def function(data, targets):
    # Do something
    return data
```

See for example the MWE below.

```python
def set_time_from_start(data, targets, 
                        works_on_target='*records*', 
                        kwargs_are_fine=True):
     print('Time field is changed to start of run')
     
     # You can use kwargs as long as you obey the positional arguments 'data' and 'targets'
     if kwargs_are_fine:
          pass
    
     # Let's check in our function that we actually want to work on this kind of target(s)
     if np.any([fnmatch.fnmatch(t, works_on_target) for t in strax.to_str_tuple(targets)]):
          # At least one of the targets is of the kind we want to apply this function to
          data['time']=data['time'][0]
     else: 
          # Ok this some non (raw-)record like data. Let's not do anything to the data
          pass
     return data

#Initiate some context 
st = strax.Context(...) # such as st = straxen.contexts.xenonnt_online()

# Now wrap whatever function we have around the data
st.set_context_config({'apply_data_function': (set_time_from_start,)})

# When doing this we will get the print message and data where 'time' starts to count in ns
# since run
st.get_array(run_id='008657', targets = ('raw_records'), seconds_range = (0,1))
```

**What does it affect?**
`get_df`, `get_array` and `accumulate`. It does not affect `get_iter` (directly) otherwise also `st.make` would be affected which would be problematic as one might be applying corrections twice (which is fine if your cutting data but not fine if you switch channels).

After discussing the PR, `accumulate` is also affected to keep loading of data consistent across all methods.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1018,17 +1018,10 @@ class Context:
                 max_workers=max_workers,
                 **kwargs)
             results = [x.data for x in source]
-        
-        result = np.concatenate(results)
-        
-        for function in self.context_config['apply_data_function']:
-            if not hasattr(function, '__call__'):
-                raise TypeError(f'apply_data_function in the context_config got '
-                                f'{function} but expected callable function with')
-            # Make sure that the function takes single argument data (or set other 
-            # arguments as kwargs)
-            result = function(result, targets)
-        return result
+
+        results = np.concatenate(results)
+        results = self._apply_function(results, targets)
+        return results
 
     def accumulate(self,
                    run_id: ty.Union[str, tuple, list],
@@ -1048,6 +1041,10 @@ class Context:
                * A record array or dict -> fields accumulated individually
                * None -> nothing accumulated
             If not provided, the identify function is used.
+
+            NB: Additionally and independently, if there are any functions registered
+            under context_config['apply_data_function'] these are applied first directly
+            after loading the data.
 
         :param fields: Fields of the function output to accumulate.
             If not provided, all output fields will be accumulated.
@@ -1081,6 +1078,7 @@ class Context:
 
         for chunk in self.get_iter(run_id, targets, **kwargs):
             data = chunk.data
+            data = self._apply_function(data, targets)
 
             if n_chunks == 0:
                 result['start'] = chunk.start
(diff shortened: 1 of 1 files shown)


WenzDaniel:
>Also, accumulate is not affected
While I think it is a nice to have thing for `get_array`  I do not see its need in `accumulate` since you can apply functions here anyhow.

strax/context.py
```diff
@@ -1014,7 +1017,17 @@ def get_array(self, run_id: ty.Union[str, tuple, list],
                 max_workers=max_workers,
                 **kwargs)
             results = [x.data for x in source]
-        return np.concatenate(results)
+        
+        result = np.concatenate(results)
+        
+        for function in self.context_config['apply_data_function']:
```
WenzDaniel, line 1023: This function will be applied to all data_kinds/types which might not be desirable since it can be easily get forgotten while working in the same context. Maybe some dictionary would be here more desirable as strax.Option e.g. {'peak_basics': lambda res: res['area'] + 10}.

strax/context.py
```diff
@@ -1014,7 +1017,17 @@ def get_array(self, run_id: ty.Union[str, tuple, list],
                 max_workers=max_workers,
```
WenzDaniel, line 1017: Since this function will be applied outside of get_iter it wont be included in the computation time of the progressbar. Now, I also understand why progressbars always tend to take ages for the last 0.2 % ;-) Maybe save thing would be something like if self.config['apply_data_function'] switch off progressbar.

jorana: We could do that, on the other hand, it's fine if the progress bar reflects the loading. What happens after that can fal outside of the scope of the progress bar. For example, the select_runs also contains several steps that take different amounts of time.

strax/context.py
jorana: Good suggestion. In the implementation I envisage for straxen this is desirable (want to check all datatypes). See https://github.com/XENONnT/straxen/pull/166 and more importantly https://github.com/XENONnT/analysiscode/blob/master/DAQ/remap_sectors/Software_remap.ipynb. 

I'd propose to also make the datatype the second arg of each function and do a regex in the function. 

>  ... since it can be easily get forgotten while working in the same context.

Perhaps it's good to stress that I think this is quite a big hammer that really should be expert only as the data that you load is not the data that is stored on disk (what you want for blinded data for example).

jorana:
>> Also, accumulate is not affected

> While I think it is a nice to have thing for get_array I do not see its need in accumulate since you can apply functions here anyhow

True, perhaps my question makes more sense if you look at XENONnT/straxen#166. There I implement a remapping of channels that would be applied to any loading of data. Technically accumulate can also return data. If that is not remapped we might run into inconsistency issues between accumulate and get_array. On the other hand, I'm not sure if many people actually use this nice feature (and how they use it).

strax/context.py
jorana: @WenzDaniel I've updated this PR and updated the description above. This allows you to check it in the function you are feeding strax.

WenzDaniel: Okay now I see, and now I also agree that we should apply this kind of correction in accumulate.

jorana:
As mentioned by Daniel the function is now also applied to accumulate. This PR is ready for merging

strax/context.py
```diff
@@ -1283,6 +1295,29 @@ def _check_forbidden(self):
...
+        if not isinstance(apply_functions, (tuple, list)):
+            raise ValueError(f"apply_data_function in context config should be tuple of "
+                             f"functions. Instead got {apply_functions}")
+        for function in apply_functions:
+            if not hasattr(function, '__call__'):
+                raise TypeError(f'apply_data_function in the context_config got '
+                                f'{function} but expected callable function with two '
+                                f'positional arguments: f(data, targets).')
+            # Make sure that the function takes two arguments (data and targets)
+            data = function(data, targets)
```
WenzDaniel, line 1319: Maybe, just check how traceable a "targets" does not exist error would be. In case the error message is not traceable/understandable at all (e.g. due to numba), maybe add an explicit check with ValueError

jorana: This is about as understandable as it gets:
afbeelding

Issue #285, opened 2020-07-15
Fuzzy_for not working as documented?


WenzDaniel:
Hi I was a bit playing around and I noticed that fuzzy_for does not seem to work always as intended. I used my own strax(en) environment including my DAQreader which has some nveto related modification with respect to the current default DAQreader. 

E.g. my hash of raw_records is lgjhqtmqfj, while with the default DAQreader we get rfzvpzj4mf. I thought this would be an excellent exercise to look a bit into the fuzzy_for option and I tried the following:

```
st = straxen.contexts.xenonnt_online(fuzzy_for=('raw_records'))  # Using my DAQreader

st.storage.pop(0) # RunDb is not working with fuzzy_for

print(st.is_stored(run_id_led, 'raw_records'))  # which returns false
for i in st.get_iter(run_id_led, 'raw_records'):
    pass  # not found error
```

**Update:**

I think this example did not work since `fuzzy_for` expects its input to be a tuple. But if specified as `fuzzy_for=('raw_records')` things will changed behind the scene into a string. The correct way would be `fuzzy_for=('raw_records', )`. I think this looks like a bug.

WenzDaniel:
I continued working a bit with fuzzy_for and it can be quite annoying since things die silently if not used correctly.

WenzDaniel:
Ubs did not want to press the big red button.... 

Maybe we should implement a user warning that if fuzzy_for is specified and no data is "stored" that the user may want to check if the lower level plugins changed as well. Or add an option "load_hash" or something like this.

WenzDaniel:
Solved in https://github.com/AxFoundation/strax/pull/306

PR #306, opened 2020-08-17
Workaround for fuzzy_for issue.

WenzDaniel:
The issue is elaborated in https://github.com/AxFoundation/strax/issues/285
This PR is a small patch which will make the fuzzy_for option more flexible. The user can now specify any provides field in `fuzzy_for`. The user can now also just specify a string instead of a tuple.    
`st.set_context_config({'fuzzy_for': 'peaklets'})`


jorana:
Thanks Daniel

PR #309, opened 2020-08-28
Fix memory leak peaksplitting

jorana:
The way hitlets were now using the peak splitting induced a serious memory leak. Also it utterly degraded strax' performance.

I removed the associated changes in https://github.com/AxFoundation/strax/pull/275 that induced this bug.

**Can you describe the symptoms?**
When running straxer (e.g. like this):
```
python -m memory_profiler straxer.py 170206_1355 --target peaklets --context xenon1t_dali --build_lowlevel
```

Would yield something like:
```
Got 1525 items. Now 629.4 sec / 17.5% into the run. Using 7034.7 MB RAM. ETA 26021.48 sec.
....
Got 1936 items. Now 756.6 sec / 21.0% into the run. Using 8234.6 MB RAM. ETA 25669.28 sec.
```
Until the memory usage went up to 45 GB! Also the processing speed by a factor of ~5.


strax/processing/peak_splitting.py
```diff
@@ -189,6 +170,54 @@ def _split_peaks(split_finder, peaks, orig_dt, is_split, min_area,
...
+                is_split[p_i] = True
+                r = new_hits[offset]
+                r['time'] = p['time'] + prev_split_i * p['dt']
+                r['channel'] = p['channel']
+                # Set the dt to the original (lowest) dt first;
+                # this may change when the sum waveform of the new peak
+                # is computed
+                r['dt'] = orig_dt
+                r['length'] = (split_i - prev_split_i) * p['dt'] / orig_dt
+                r['max_gap'] = -1  # Too lazy to compute this
```
WenzDaniel, line 206: hitlets does not support 'max_gap' so simply remove this line.

WenzDaniel (commented):
I have one more idea which prevents us from copy-pasting the code. I would like try this first before we use your suggestion. Its a bit inspired by your solution. Can you forward me the memory and performance tests you did to so I can compare?

WenzDaniel:
Here is a different kind of solution https://github.com/AxFoundation/strax/pull/310 which avoids the copy and pasting of the code.

strax/processing/peak_splitting.py
```diff
@@ -189,6 +170,54 @@ def _split_peaks(split_finder, peaks, orig_dt, is_split, min_area,
...
 
+    @staticmethod
+    @strax.growing_result(dtype=strax.hitlet_dtype(), chunk_size=int(1e4))
+    @numba.jit(nopython=True, nogil=True)
+    def _split_hitlets(split_finder, hits, orig_dt, is_split, min_area,
+                       args_options,
+                       _result_buffer=None, result_dtype=None):
+        """Loop over peaks, pass waveforms to algorithm, construct
+        new peaks if and where a split occurs.
+        """
```
WenzDaniel, line 181: Please add here some warning that changes in this function might also be applied in _split_peaks

jorana: Thank Daniel, did so in 17b4c4e

strax/processing/peak_splitting.py
jorana: Fixed in 17b4c4e

jorana:
Thanks Daniel, I do think we do indeed have a different opinion on this one. (see #310)

I just addressed your feedback, would you mind verifying if this line is okay for some nVeto data (I don't have any at hand):
https://github.com/AxFoundation/strax/blob/17b4c4e7a1225b3d3990177c8e3f4d5a87bd0ba4/strax/processing/peak_splitting.py#L176

strax/processing/peak_splitting.py
```diff
@@ -86,42 +86,28 @@ def __call__(self, peaks, records, to_pe, data_type,
...
-                if split_i == NO_MORE_SPLITS:
-                    return
-                r['record_i'] = p['record_i']
-        else:
-            raise TypeError(f'Unknown data_type. "{data_type}" is not supported.')
-        new_peaks = self._split_peaks(
+        new_peaks = split_function[data_type](
             # Numba doesn't like self as argument, but it's ok with functions...
             split_finder=self.find_split_points,
             peaks=peaks,
```
WenzDaniel, line 97: You have to change these lines into arguments, since `peaks` is called `hits` in `_split_hitlets`

WenzDaniel (approved):
Looks fine go ahead

PR #312, opened 2020-09-02
Checks for multioutput plugins and loop dependencies

### LINKED ISSUE #311: Loop plugin fails when it depends on a multi-output plugin
As seen [here](https://github.com/AxFoundation/strax/blob/master/strax/plugin.py#L610), if a loop plugin needs to figure out what it loops over, it uses the first data kind as listed in its `depends_on` member. However, if this is the product of a multi-output plugin, `loop_over` will evaluate to a dictionary, and an error will be thrown at L614 when it tries accessing `kwargs`.

darrylmasson:

If the first dependency of a loop plugin is produced by a multi-output plugin, the automatic determination of `loop_over` fails. During the automatic determination of loop order, `loop_over = self.deps[self.depends_on[0]].data_kind` ([plugin.py L610](https://github.com/AxFoundation/strax/blob/master/strax/plugin.py#L610)) evaluates to a dictionary rather than a string, and this fails 4 lines down when it dereferences `kwargs`.

This PR checks to see if the `data_kind` of the plugin providing the first dependency is a multi-output plugin and asks the user to be more specific if it is. Fixes #311 .

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -608,9 +608,9 @@ class LoopPlugin(Plugin):
             loop_over = self.loop_over
         else:
             loop_over = self.deps[self.depends_on[0]].data_kind
-            if not isinstance(loop_over, str):
-                raise TypeError("Please add \"loop_over = <base>\""
-                                " to your plugin definition")
+        if not isinstance(loop_over, str):
+            raise TypeError("Please add \"loop_over = <base>\""
+                            " to your plugin definition")
 
         # Group into lists of things (e.g. peaks)
         # contained in the base things (e.g. events)
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -608,6 +608,9 @@ class LoopPlugin(Plugin):
             loop_over = self.loop_over
         else:
             loop_over = self.deps[self.depends_on[0]].data_kind
+            if isinstance(loop_over, dict):
+                loop_over = loop_over[self.depends_on[0]]
+                # We can't have nested dictionaries here, can we?
 
         # Group into lists of things (e.g. peaks)
         # contained in the base things (e.g. events)
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -608,9 +608,9 @@ class LoopPlugin(Plugin):
             loop_over = self.loop_over
         else:
             loop_over = self.deps[self.depends_on[0]].data_kind
-            if isinstance(loop_over, dict):
-                loop_over = loop_over[self.depends_on[0]]
-                # We can't have nested dictionaries here, can we?
+            if not isinstance(loop_over, str):
+                raise TypeError("Please add \"loop_over = <base>\""
+                                " to your plugin definition")
 
         # Group into lists of things (e.g. peaks)
         # contained in the base things (e.g. events)


WenzDaniel:
Looks good, maybe here an additional MWE since it took me quite some time to get deps initialized:

```
plugins = st._get_plugins(targets=('events',), run_id='0')  # Generates plugins up to events
loop_over = plugins['records'].deps[plugins['records'].depends_on[0]].data_kind
```
returns a dict

```
plugins = st._get_plugins(targets=('events',), run_id='0')  # Generates plugins up to events
loop_over = plugins['records'].deps[plugins['records'].depends_on[0]].data_kind
loop_over[plugins['records'].depends_on[0]]
```
returns 'raw_records'

strax/plugin.py
```diff
@@ -608,6 +608,9 @@ def compute(self, **kwargs):
             loop_over = self.loop_over
         else:
             loop_over = self.deps[self.depends_on[0]].data_kind
+            if isinstance(loop_over, dict):
```
WenzDaniel, line 611: You have to changes this line in `isinstance(loop_over, (dict, immutabledict))` since the DAQReader has an immutable dict as data_kind. See

```
plugins = st._get_plugins(targets=('events',), run_id='0')
plugins['records'].deps[plugins['records'].depends_on[0]].data_kind
```

strax/plugin.py
```diff
@@ -608,6 +608,9 @@ def compute(self, **kwargs):
...
+                loop_over = loop_over[self.depends_on[0]]
+                # We can't have nested dictionaries here, can we?
```
WenzDaniel, line 613: I think you showed, that we cannot. So maybe remove this comment.

WenzDaniel (changes requested):

strax/plugin.py
jorana, line 613: Perhaps you can be on the safe side by making a check just below:
```python
if not istinstance(loop_over, str): 
    raise TypeError(f'Trying to loop over {loop_over} which is not a string? Please add loop_over = <base> to your plugin')
```

jorana (commented):
Thanks Darryl for spotting this, the error is indeed very non-instructive and unpacking it from a multioutput plugin like this will help recognizing when something is off / badly constructed.

Maybe it's good to add that the high-level problem is that this plugin wasn't designed for these kinds of low-level datatypes. You will (with anything below the non-overlapping peaklets) run into this assertion error: https://github.com/AxFoundation/strax/blob/a19cc214da6b24f4e7779df9e2366cfea8ee2fc3/strax/plugin.py#L619. Your base needs to be split and sorted by time. Records have overlapping time-information and therefore aren't suitable for this kind of plugin.

darrylmasson:
> Maybe it's good to add that the high-level problem is that this plugin wasn't designed for these kinds of low-level datatypes.

So what's the high-level solution we want to aim for? Prevent loop plugins from depending on `records` or lower (or other overlapping data types)?

jorana:
Hi Darryl,
Thanks, I think your solution is fine, for example for peaklets this should work (although presumably slow). We just shouldn't expect this plugin to work for things like records.

The lines I suggested were in _addition_ to what you added before. 

Combining this would lead to :
```python
...
if isinstance(loop_over, (dict, immutabledict)):
    loop_over = loop_over[self.depends_on[0]]
if not isinstance(loop_over, str):
    raise TypeError("Please add \"loop_over = <base>\""
                            " to your plugin definition")
...
```

WenzDaniel:
Would not your proposal the lines above obsolete? 

```
if not isinstance(loop_over, str):
    raise TypeError("Please add \"loop_over = <base>\""
                            " to your plugin definition")
```
Would always be raised when 
```
if isinstance(loop_over, (dict, immutabledict)):
    loop_over = loop_over[self.depends_on[0]]
```
is true. Hence we could remove it.

jorana:
> 
> 
> Would not your proposal the lines above obsolete?
> 
> ```
> if not isinstance(loop_over, str):
>     raise TypeError("Please add \"loop_over = <base>\""
>                             " to your plugin definition")
> ```
> 
> Would always be raised when
> 
> ```
> if isinstance(loop_over, (dict, immutabledict)):
>     loop_over = loop_over[self.depends_on[0]]
> ```
> 
> is true. Hence we could remove it.

No, you redefine loop_over the line just above it?

darrylmasson:
If we do go for something that looks like this:
```
if hasattr(self, 'loop_over'):
    loop_over = self.loop_over
else
    loop_over = self.deps[self.depends_on[0]].data_kind
if isinstance(loop_over, (dict, immutabledict)):
    loop_over = loop_over[self.depends_on[0]]
if not isinstance(loop_over, str):
    raise TypeError("message")
```
The second `isinstance` will only do anything in the situation where a plugin is particularly convoluted. I would argue to skip the first check. The second is a catch-all case that should handle any ambiguous situation.

jorana:
The first ```isinstance``` changes the loop_over and is not a check as such. I don't follow your reasoning as you will end up with the ```TypeError``` if you skip the first ```isinstance```.

darrylmasson:
That's my point - the second `isinstance` is a catch-all that should handle any ambiguous situation, and multioutput plugins are at least somewhat ambiguous in this regard. Rather than individually handle these cases, only having `if not isinstance(loop_over, str)` both makes the code easier to maintain and also deals with any possible future features that could affect this in some way.

jorana:
There is nothing wrong with multi-output plugins per se but perhaps you are right that rather than doing the thinking for the analysts that just accepting a simple format is clearer (and throwing an error otherwise).

jorana (approved):

WenzDaniel (approved):
Neat

PR #315, opened 2020-09-11
Rewrite mongo.py for OnlineMonitor

jorana:
For online monitoring we'd like to have a Mongo Backend where we can save the output of plugins into a database. Using this PR we add such a backend.

The nice thing is that we we should even be able to open selected plugins at the very moment we are writing them on the eventbuilders.

- This writes just like any other frontend but this time to a mongo database. 
- Change the logic in mongo.py to write to a single collection (only if not provided we use the datakey to make a new collection)

The output will look like:
https://gist.github.com/jorana/d3176c4e561ccca4e8ca2fc2420acf7c

For code see:
https://github.com/XENONnT/analysiscode/pull/27

**Edit 28-09-20 Minimal examples**
Below I add the minimal examples of what this actually does. For more info please check the above linked analysis repo.

_input_ (see the PulseCountMongoSaver2 plugin in the notebook above)
```
array([(1598164719000000000, 1598164724499999000, 7137111),
       (1598164724499999000, 1598164729999999000, 7137111),
       (1598164729999999000, 1598164735100000000, 7137111)],
      dtype=[(('Start time of the chunk', 'time'), '<i8'), (('End time of the chunk', 'endtime'), '<i8'), (('Tot Number of pulses', 'pulse_count'), '<i8')])
```
The output is the same

_Database Document_
```python
{"_id": ObjectId("5f71f91f5de55f1788fa7e6e"),
  "number": 9104,
  "data_type": "mongo_pulses2",
  "lineage_hash": "5qeiyy4tfm",
  "metadata": {"run_id": "009104",
   "data_type": "mongo_pulses2",
   "data_kind": "pulse_counts",
   "dtype": "[(("Start time of the chunk", "time"), "<i8"), (("End time of the chunk", "endtime"), "<i8"), (("Tot Number of pulses", "pulse_count"), "<i8")]",
   "lineage_hash": "5qeiyy4tfm",
   "compressor": "blosc",
   "lineage": {"mongo_pulses2": ["PulseCountMongoSaver2",
     "0.0.8",
     {"n_tpc_pmts": 494}],
    "pulse_counts": ["PulseProcessing",
     "0.2.2",
     {"n_tpc_pmts": 494,
      "hev_gain_model": ["disabled", None],
      "baseline_samples": 40,
      "tail_veto_threshold": 0,
      "tail_veto_duration": 3000000,
      "tail_veto_resolution": 1000,
      "tail_veto_pass_fraction": 0.05,
      "tail_veto_pass_extend": 3,
      "pmt_pulse_filter": None,
      "save_outside_hits": [3, 20],
      "hit_min_amplitude": "pmt_commissioning_initial"}],
    "raw_records_mv": ["DAQReader", "0.0.0", {}]},
   "chunk_target_size_mb": 200,
   "strax_version": "0.12.1",
   "writing_started": 1601304863.9505675,
   "chunks": [{"chunk_i": 0,
     "n": 3,
     "start": 1598164719000000000,
     "end": 1598164735100000000,
     "run_id": "009104",
     "nbytes": 72,
     "first_time": 1598164719000000000,
     "first_endtime": 1598164724499999000,
     "last_time": 1598164729999999000,
     "last_endtime": 1598164735100000000}],
   "writing_ended": 1601304865.2791328},
  "chunk_0": [{"data": {"time": 1598164719000000000,
     "endtime": 1598164724499999000,
     "pulse_count": 7137111}},
   {"data": {"time": 1598164724499999000,
     "endtime": 1598164729999999000,
     "pulse_count": 7137111}},
   {"data": {"time": 1598164729999999000,
     "endtime": 1598164735100000000,
     "pulse_count": 7137111}}]}
     
```


strax/storage/mongo.py
```diff
@@ -11,100 +11,159 @@
...
-    def __init__(self, uri, database):
+    """Mongo storage backend"""
+    def __init__(self, uri, database, col_n=None):
+        """
+        Backend for reading/writing data from Mongo
+        :param uri: Mongo url (with pw and username)
+        :param database: name of database (str)
+        :param col_n: collection name (str) to look for data
+        """
         if MongoClient is None:
```
darrylmasson, line 29: If PyMongo doesn't import, you'll get an `ImportError` when you try to load this file, and this line will never call. Did you mean to try-except the import statement?

strax/storage/mongo.py
```diff
@@ -11,100 +11,159 @@
...
 
     def _read_chunk(self, backend_key, chunk_info, dtype, compressor):
+        """See strax.Backend"""
+        query = backend_key_to_query(backend_key)
+        chunk_name = f'chunk_{chunk_info["chunk_i"]}'
+        # Query for the chunk and project the chunk info
+        doc = self.db[self.col_name].find_one(
+            {**query, chunk_name: {"$exists": True}},
+            {f"{chunk_name}.$": 1})
+        doc = doc.get(chunk_name, None)
```
darrylmasson, line 44: Might want to check first for `doc is None` in case the query doesn't find anything. (You know, the line beneath this one)

strax/storage/mongo.py
```diff
@@ -11,100 +11,159 @@
...
+
         super().__init__(*args, **kwargs)
         if MongoClient is None:
             raise ImportError("Pymongo did not import: "
                               "cannot use MongoDBStore")
         self.client = MongoClient(uri)
         self.db = self.client[database]
-        self.backends = [MongoBackend(uri, database)]
+        self.backends = [MongoBackend(uri, database, col_n=col_n)]
+        self.col_name = col_n
```
darrylmasson, line 92: What's the argument for using `col_n` in the argument and `col_name` as the attribute? Wouldn't things be a bit simpler if they were the same?

strax/storage/mongo.py
```diff
@@ -11,100 +11,159 @@
...
 
     def _find(self, key, write, allow_incomplete, fuzzy_for,
               fuzzy_for_options):
-
+        """See strax.Frontend"""
         if write:
             return self.backends[0].__class__.__name__, str(key)
-        if self.db[str(key)].count():
+        query = backend_key_to_query(str(key))
+        if self.db[self.col_name].find_one(query):
```
darrylmasson, line 100: `self.db[self.col_name].count_documents(query)` might be the more efficient query here.

darrylmasson (changes requested):
I think a few examples could help tie this together and also motivate the cunning indexing to support the queries this makes.

strax/storage/mongo.py
jorana: I'm also not sure why Dan coded it this way. But yeah, this is a good opportunity to clean this up. It doesn't make sense to me either

strax/storage/mongo.py
jorana: Good catch, let me add it

strax/storage/mongo.py
jorana: If you think its easier I see little reason not to change it so will do that. There was a reason why I did it like this but I cannot remember.

strax/storage/mongo.py
jorana: I'm not sure why? Looking at https://docs.mongodb.com/manual/reference/method/db.collection.countDocuments/ it just aggregates the query. I don't mind changing it though.

jorana:
Hi @darrylmasson, thanks for the review. Will implement the requested changes and tidy up the exemplary notebook I provided in the description of this PR. Please specify what examples you'd want to see that are omitted in the notebook.

strax/storage/mongo.py
darrylmasson: The main reason is that `find_one` returns an entire document, while `count_documents` returns an integer. If the document in question happens to be "large", then the API downloads and returns the entire thing just to see if it exists.

strax/storage/mongo.py
darrylmasson: I don't have a strong opinion on this, I'm just thinking about readability.

darrylmasson:
I think the main gotchas it would be nice to catch ahead of time are be related to key/query generation and data formatting. The useful examples would be the encoding and decoding of a chunk between the database and strax, so we can make sure that everything lines up as we expect it to.

strax/storage/mongo.py
jorana: Thanks Darryl!

jorana:
@darrylmasson thanks, the comments above should be addressed now. I have uploaded the notebook again and will add some minimal examples to description of this PR

strax/storage/mongo.py
darrylmasson: I think I've made some occasional similar comments in bootstrax PRs when it calls `find_one_and_update` rather than `update_one` (the difference being whether or not a document is returned), but at least in those cases it's usually much clearer exactly what size the document is that it's downloading and discarding.

strax/storage/mongo.py
```diff
@@ -40,15 +38,16 @@ def _read_chunk(self, backend_key, chunk_info, dtype, compressor):
...
-        doc = doc.get(chunk_name, None)
+            {f"{chunk_name}": 1})
 
-        if doc is None:
+        # Unpack info about this chunk from the query. Return empty if not available.
+        if doc is None or chunk_name not in doc:
             return np.array([], dtype=dtype)
-
+        else:
+            chunk_doc = doc[chunk_name]
```
darrylmasson, line 47: The c++ part of my brain (which is most of it right now) is complaining that `chunk_doc` goes out of scope after this line and so shouldn't exist when it's used later on, but it works. Though, the most pedantic observers could argue that because the `if` statement includes a return, the `if-else` construct doesn't actually do anything beyond `if` alone.

darrylmasson (approved):

strax/storage/mongo.py
```diff
@@ -40,15 +38,16 @@ def _read_chunk(self, backend_key, chunk_info, dtype, compressor):
         # Query for the chunk and project the chunk info
         doc = self.db[self.col_name].find_one(
             {**query, chunk_name: {"$exists": True}},
-            {f"{chunk_name}.$": 1})
-        doc = doc.get(chunk_name, None)
+            {f"{chunk_name}": 1})
```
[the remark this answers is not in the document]
jorana: Sorry yeah it was a good request to add the examples. I kind of shuffled it under the carpet in the previous commit. Please find the edited description above. Does this help?

strax/storage/mongo.py
jorana: The 'else' makes it more explicit this is arguably preferable as otherwise someone might overlook that the if statement checks that the ```doc[chunk_name]``` will not fail. On the internet you can find many pros/cons for these kind of else statements. See e.g. https://dmerej.info/blog/post/else-after-return-yea-or-nay/. 

I don't think it matters so if you think it's better I'm fine with it.

strax/storage/mongo.py
jorana: Thanks, will remember it!

PR #318, opened 2020-09-11
Scada interface

WenzDaniel:
In this PR we make some changes to the time range querying via strax. 

I slightly modified estimate_run_start such that it returns the run endtime as well.
Further to_absolute_time_range was modified such that it returns the run endtime in case no time range is selected.

At the end of this [notebook](https://github.com/XENONnT/analysiscode/blob/master/StraxTests/ScadaInterface.ipynb) briefly show the changes.


strax/context.py
```diff
@@ -781,36 +797,44 @@ def estimate_run_start(self, run_id, targets=None):
...
-        if ((time_range is None)
-                + (seconds_range is None)
-                + (time_within is None)
-                < 2):
+        
+        selection = ((time_range is None) +
+                     (seconds_range is None) +
+                     (time_within is None) +
+                     (full_range is None))
+        if selection < 2:
```
jorana, line 833: @WenzDaniel , could you make it a bool and update the Runtime error?
As in : 
full_range=False # as kwarg
full_range==False # in selection comparison

and 
 " time_range, seconds_range, time_within or full_range"

jorana (approved):
Seems clean and straigtforward!

sorry for 49c0674 it was just a super minor change and some reformating

strax/context.py
jorana: I think a bool is cleaner for an option but it's not too relevant

WenzDaniel: I corrected the raise message. About the other thing, first I thought so too. But I think Jelle did this intentionally to allow arrays as input. Otherwise things like 

`if np.array([start, stop]) :`

would give an ambiguity error and one would have to change it in something like:

`if np.any(selection) :`

PR #320, opened 2020-09-29
Add python3.8 to travis

jorana:
As illustrated in figure 4 of https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:dsg:daq:eb_speed_tests, python 3.8 showed an increased performance on the XENONnT eventbuilder machines. Therefore, we might want to test strax also for python 3.8.


darrylmasson (approved):
Hopefully we'll find out if it causes any crashes before we deploy fully.

PR #321, opened 2020-10-07
quick fix for zero division error - and Travis issues

jorana:
Quick patch to an error we encountered in one of the latest runs. Although strange we shouldn't make this kind of errors cause the entire stack of the processing crash.

don't do x/0

**Edit after all the commits you can see in this PR**
Should have renamed this PR to 'fix travis issues'. This was the colateral damage:
- Apparently travis has started running with multiple threads. This means that we cannot rely on the old counting of threads.
- Secondly we ran into one of the conditional entropy tests. I slightly relaxed the math.isclose requirement.

### DIFF
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -111,12 +111,12 @@ class Chunk:
         return self.end - self.start
 
     def _mbs(self):
-        if self.duration:
-            return (self.nbytes / 1e6) / (self.duration / 1e9)
-        else:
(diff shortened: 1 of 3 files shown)


jorana:
Okay something is broken in the mailbox test. Look at 629f79e. This is not at all an issue related to PR..

jorana:
Ok as this is a trivial PR I'm merging it.

PR #322, opened 2020-10-07
Option to give range to sum_waveform

zhut19:

The desaturation requires re-calculating the sum waveform of only a small fraction of the peaklets (that are saturated), once the (temporary copy of) records are corrected. It is best to be able to do partial re-calculation.

`func: sum_waveform` now takes a keyword argument `peak_list` of the type `numpy.array`. When not specified, default value would be converted into `np.arange(len(peaks))`

### DIFF
--- a/strax/processing/peak_building.py
+++ b/strax/processing/peak_building.py
@@ -137,12 +137,11 @@ def store_downsampled_waveform(p, wv_buffer):
 
 @export
 @numba.jit(nopython=True, nogil=True, cache=True)
-def sum_waveform(peaks, records, adc_to_pe,
-                 peak_list=np.array([-3, -14])):
+def sum_waveform(peaks, records, adc_to_pe, *arg):
     """Compute sum waveforms for all peaks in peaks
     Will downsample sum waveforms if they do not fit in per-peak buffer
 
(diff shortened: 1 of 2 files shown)


jorana:
Hi @zhut19 , the style seems very convoluted. Can you explain why you need -3 and -42? I think you can more easily and clearly do this with a simple None kwarg?

By the way, travis is failing untill we merge https://github.com/AxFoundation/strax/pull/321. Don't worry about that too much.

zhut19:
Thanks, @jorana. I wasn't thinking. The keyword argument was one way to keep compatibility with the original code, a silly one tbh.  With the switching to `*arg` in https://github.com/AxFoundation/strax/pull/322/commits/6aeb711adf63db8e63522fa5bcc53155fdb1fa83, is this good for you?

jorana (approved):
Sure, would work! But if at all possible I'd be a little more explicit, would you like 6275049? Otherwise I can revert it?

zhut19:
This is even better! I remove the raise error, as an empty list is okay and does exist.

jorana:
Sure thanks, I was about to remove that raise statement for exactly the reason you mention. Just for completeness I've added this to the tests in f92086f

Issue #324, opened 2020-10-08
New mongo versions projection rules cause Path collision errors


jmosbacher:
Following a slack discussion:
From version 4.4 onward it is not allowed to project a field and its subfield in a single query. Since strax adds some fields to the projections here:
   ```
   store_fields = tuple(set(
        list(strax.to_str_tuple(store_fields))
        + ['name', 'number', 'tags', 'mode',
           strax.RUN_DEFAULTS_KEY]
        + list(self.context_config['store_run_fields'])))
    check_available = tuple(set(
        list(strax.to_str_tuple(check_available))
        + list(self.context_config['check_available'])))
```
which makes it impossible to replace a default field with a subfield of that field from the context parameter.

To replicate the old mongo behavior, strax should only add default fields for which none of the sub fields are requested in the context. Something like:
```
default_fields = ['name', 'number', 'tags', 'mode',
           strax.RUN_DEFAULTS_KEY] + list(strax.to_str_tuple(store_fields))
explicit_fields = [field.split(".")[0] for field in self.context_config['store_run_fields']]
extra_fields = [x for x in default_fields if x not in explicit_fields]
   store_fields = tuple(set(    
        + extra_fields
        + list(self.context_config['store_run_fields'])))
    check_available = tuple(set(
        list(strax.to_str_tuple(check_available))
        + list(self.context_config['check_available'])))
```
Or something more elegant

jorana:
As you pointed out in #325 it would be good to also add some automatized tests for this.

jorana:
This seems fixed and mostly on the `straxen` side. I don't think it's worth keeping the issue open as a reminder that we need to add tests to the mongo interfaces. Feel free to reopen.

PR #325, opened 2020-10-09
Fix mongo path conflicts

jmosbacher:
This should fix issue #324

After collecting all the projections, the most specific projection for each field is chosen. e.g if both "tags" and "tags.name" are requested, only "tags.name" will actually go into the mongo query. This will reproduce the old mongodb behavior.

A new test for this should be added. Hope to have time to do this soon.


jorana:
@jmosbacher , thanks for the quick fix! I'm merging it already. I just tested it and it seems to work. Perhaps we can keep issue #324 open and mention the tests there.

jorana:
Looking at this a little more closely, it looks like a straxen-issue rather than a strax issue. It seems that https://github.com/XENONnT/straxen/pull/241 would be a viable solution right? At least, if I quickly run that on my laptop it seems to solve the issue.

jmosbacher:
Sorry I just realized my quick fix doesnt work as intended for the edge case when two fields have names that start the same way. So if you request eg both date and datetime for example (not real fields but could be) then you would only get datetime. I added another commit, this is shorter and should work in all cases (i think).

jmosbacher:
> Looking at this a little more closely, it looks like a straxen-issue rather than a strax issue. It seems that [XENONnT/straxen#241](https://github.com/XENONnT/straxen/pull/241) would be a viable solution right? At least, if I quickly run that on my laptop it seems to solve the issue.

@jorana Yes, I think you are right. Good catch! I didnt think it through, just chased down the source of the error. But I guess a non mongo frontend could support fields and sub-fields in parallel so no reason to limit ourselves here.

jorana:
@ jmosbacher , would we still want to add this to strax or otherwise add a comment with a reference to this PR/issue such that it's easier to debug if others run into it?

jmosbacher:
@jorana Good question. I think this pull request is now redundent so definitely no need to merge it. I like the idea of a comment. In any case I am closing this pull request.

Issue #326, opened 2020-10-09
Floats in mongo.py not removed


jorana:
Issue floats in mongo.py as the types are not tested broadly enoug.

See for example:
```python
d = np.float32(1)
isinstance(d, np.float), isinstance(d, np.floating)
# returns
(False, True)
```

PR #327, opened 2020-10-09
Do proper typetesting for np.floats

jorana:
Solves issue: https://github.com/AxFoundation/strax/issues/326

See:
https://stackoverflow.com/questions/28292542/how-to-check-if-a-number-is-a-np-float64-or-np-float32-or-np-float16

```python
d = np.float32(1)
isinstance(d, np.float), isinstance(d, np.floating)
# returns
(False, True)
```


WenzDaniel (approved):
Nice catch

Issue #328, opened 2020-10-13
Array view instead of copy


WenzDaniel:
Users should be aware that lines like the following:
```
for d in data_type:
    data= d['data'][:d['length']]
    modifies_data(data)
```
will create only a "view" onto data in memory. Hence when `data` is modified by any function, e.g. `modifies_data`, `data_type['data']`will be modified as well. This behavior is in many cases not desired and leads to bugs which are really hard to trace. The correct way looks like:
```
for d in data_type:
    data = np.copy(d['data'][:d['length']])
    modifies_data(data)
```
See also https://stackoverflow.com/questions/11524664/how-can-i-tell-if-numpy-creates-a-view-or-a-copy and  https://stackoverflow.com/questions/4370745/view-onto-a-numpy-array for some additional information.

PR #329, opened 2020-10-13
Fixing _conditional_entropy does not modifiy hitlet data.

WenzDaniel:

Solves issue https://github.com/AxFoundation/strax/issues/328 for the conditional_entropy.


jorana (approved):
I personally like `array.copy()` better but that is all I can say about this fix. Thanks Daniel!

PR #335, opened 2020-10-21
Create seperate doc for new chunk

jorana:
We experienced that some documents got bigger than the 16 MB limit imposed by mongo due to many chunks being written to the same document. This is solved in this PR by splitting out the chunks-data and the meta-data into separate documents.

https://github.com/XENONnT/analysiscode/blob/master/StraxTests/add_mongo_saver_finetuning.ipynb

**Doc format**
See the gist below for an example. Here we uploaded three chunks with a length == 5 datafield (to keep it simple). This means we have four documents in total:
https://gist.github.com/jorana/c4fccefa1a152b403bace944a8993699

### DIFF
--- a/strax/storage/mongo.py
+++ b/strax/storage/mongo.py
@@ -35,11 +35,11 @@ class MongoBackend(StorageBackend):
     def _read_chunk(self, backend_key, chunk_info, dtype, compressor):
         """See strax.Backend"""
         query = backend_key_to_query(backend_key)
-        chunk_name = f'chunk_{chunk_info["chunk_i"]}'
+        chunk_i = chunk_info["chunk_i"]
 
         # Query for the chunk and project the chunk info
         doc = self.db[self.col_name].find_one(
-            {**query, "chunk_i": chunk_name},
+            {**query, "chunk_i": chunk_i},
             {f"data": 1})
 
         # Unpack info about this chunk from the query. Return empty if not available.
@@ -49,7 +49,7 @@ class MongoBackend(StorageBackend):
         else:
             chunk_doc = doc.get('data', None)
             if chunk_doc is None:
-                raise ValueError(f'Doc for {chunk_name} in wrong format:\n{doc}')
+                raise ValueError(f'Doc for chunk_{chunk_i} in wrong format:\n{doc}')
 
         # Convert JSON to numpy
         chunk_len = len(chunk_doc)
@@ -153,8 +153,7 @@ class MongoSaver(Saver):
 
     def _save_chunk(self, data, chunk_info, executor=None):
         """see strax.Saver"""
-        chunk_number = chunk_info['chunk_i']
-        chunk_name = f'chunk_{chunk_number}'
+        chunk_i = chunk_info['chunk_i']
 
         aggregate_data = []
         # Remove the numpy structures and parse the data. The dtype information
@@ -168,7 +167,7 @@ class MongoSaver(Saver):
 
         # Get the document to update, if none available start a new one for
         # this chunk
-        chunk_id = self.ids_chunk.get(chunk_name, None)
+        chunk_id = self.ids_chunk.get(chunk_i, None)
         if chunk_id is not None:
             self.col.update_one({'_id': chunk_id},
                                 {'$addToSet': {f'data': aggregate_data}})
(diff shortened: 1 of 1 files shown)


strax/storage/mongo.py
```diff
@@ -133,46 +136,81 @@ def __init__(self, key, metadata, col):
...
+        # this chunk
+        chunk_id = self.ids_chunk.get(chunk_name, None)
+        if chunk_id is not None:
+            self.col.update_one({'_id': chunk_id},
+                                {'$addToSet': {f'chunk_name.data': aggregate_data}})
+        else:
+            # Start a new document, update it with the proper information
+            doc = self.basic_md.copy()
+            doc['write_time'] = datetime.now(py_utc)
+            doc[chunk_name] = {"data": aggregate_data}
```
darrylmasson, line 178: This line here is going to make life difficult for the analyst. You'll end up with chunks looking like this:
```
{chunk_0: {data: ...}},
{chunk_1: {data: ...}},
...
```
You can't easily loop over this (the field you want is different for each document), projections become more difficult (strictly inclusive, not exclusive), indexing chunk number is impossible, and probably some other things. It would make life much simpler if we just make a `data` field, rather than `chunk_%i.data`, and have a separate `chunk_i` field.

strax/storage/mongo.py
```diff
@@ -133,46 +136,81 @@ def __init__(self, key, metadata, col):
...
                                  {'metadata.chunks': chunk_info}})
 
     def _close(self):
         """see strax.Saver"""
+        # First update the run-starts of all of the chunk-documents as this is
+        # a TTL index -candidate
+        if self.run_start is not None:
+            update = {'run_start_time': self.run_start}
+            for chunk_id in self.ids_chunk.keys():
+                self.col.update_one({'_id': chunk_id}, {'$set': update})
```
darrylmasson, line 206: Probably faster to do one call to `update_many({'_id': {'$in': list(self.ids_chunk.keys())}}, ...)` than many calls to `update_one`. Also probably possible to just specify `{'number': run_number}`.

strax/storage/mongo.py
```diff
@@ -133,46 +136,81 @@ def __init__(self, key, metadata, col):
...
-                                                   {"data": ins}}
-                                 })
+            aggregate_data.append(ins)
+
+        # Get the document to update, if none available start a new one for
+        # this chunk
+        chunk_id = self.ids_chunk.get(chunk_name, None)
+        if chunk_id is not None:
+            self.col.update_one({'_id': chunk_id},
+                                {'$addToSet': {f'chunk_name.data': aggregate_data}})
```
darrylmasson, line 173: `'f{chunk_name}.data'`? Anyway, see above comment about `chunk_name` as a field.

darrylmasson (changes requested):
A minor change with significant quality-of-life implications.

strax/storage/mongo.py
jorana: You are right, I considered it did not make much of a difference as people will use st.get_array but that is a lousy excuse for making confusing documents

strax/storage/mongo.py
jorana: Thank, good catch

strax/storage/mongo.py
```diff
@@ -39,14 +39,15 @@ def _read_chunk(self, backend_key, chunk_info, dtype, compressor):
 
         # Query for the chunk and project the chunk info
         doc = self.db[self.col_name].find_one(
-            {**query, chunk_name: {"$exists": True}},
-            {f"{chunk_name}": 1})
+            {**query, "chunk_i": chunk_name},
```
darrylmasson, line 42: Is there a compelling reason why we carry `chunk_name` rather than its number? Integers index much more cleanly than `chunk_%i`, and I'm not aware of cases when chunks get non-integer identifiers.

strax/storage/mongo.py
```diff
@@ -202,8 +204,9 @@ def _close(self):
         # a TTL index -candidate
         if self.run_start is not None:
             update = {'run_start_time': self.run_start}
-            for chunk_id in self.ids_chunk.keys():
-                self.col.update_one({'_id': chunk_id}, {'$set': update})
+            query = {k: v for k, v in self.basic_md.items()
+                     if k in ('run_id', 'data_type', 'lineage_hash')}
```
darrylmasson, line 205: Ok, these looks like the fields to index. If `{**query}` above looks like this we can make a single compound index.

darrylmasson (approved):

strax/storage/mongo.py
darrylmasson: Also we don't specify `run_id`, it looks like we use `number` instead.

strax/storage/mongo.py
jorana: Changed it, it's also more consistent

strax/storage/mongo.py
jorana: The query has already been indexed but not in a compound manner. I remember that compound indexing and TTLs don't go well together but these indexes are not part of the TTL indices.

darrylmasson: Mileage varies, obviously, but my usual strategy is to index for the specific queries I know are being made. If we always query against these three fields at the same time, a compound index probably provides a bit of a performance bonus. If we sometimes make queries against these fields individually, then a separate index might be better, or it might not.

PR #336, opened 2020-10-22
improve mongo-backend-loading

jorana:
Slightly tweak w.r.t. #335 (and #315). Only query the database for chunks once.

Use `find` to find all chunks, save it in ram and load it from there instead of doing up to 600 queries.

**Performance**
Using 
```python
%%timeit
om = st.get_array('009524', 'online_peak_monitor', allow_incomplete=True) # 161 chunks
```
_Before_
23 s ± 175 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

_After_
1.91 s ± 70.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

### DIFF
--- a/strax/storage/mongo.py
+++ b/strax/storage/mongo.py
@@ -41,10 +41,16 @@ class MongoBackend(StorageBackend):
         if self.chunks_registry is None:
             self._build_chunk_registry(backend_key)
 
-        doc = self.chunks_registry.get(chunk_i, None)
-        # Unpack info about this chunk from the query. Return empty if not available.
+        # Unpack info about this chunk from the query. Return empty if
+        # not available. Use a *string* in the registry to lookup the
+        # chunk-data (like we do in _build_chunk_registry).
+        doc = self.chunks_registry.get(str(chunk_i), None)
+
         if doc is None:
-            # Did not find the data
+            # Did not find the data. NB: can be that the query is off in
+            # the _build_chunk_registry. In case you end up here but did
+            # not expect that, double check that self.chunks_registry is
+            # not an empty dict!
             return np.array([], dtype=dtype)
         else:
             chunk_doc = doc.get('data', None)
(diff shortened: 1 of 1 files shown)


strax/storage/mongo.py
```diff
@@ -73,6 +73,39 @@ def get_metadata(self, key):
...
+        for doc in chunks_registry:
+            chunk_key = doc.get('chunk_i', None)
+            if chunk_key is None:
+                # Should not happen because of the projection in find
+                # but let's double check
+                raise ValueError(
+                    f'Projection failed, got doc with no "chunk_i":\n{doc}')
+            # Update our registry with this chunks info. Use chunk_i as
+            # chunk_key
+            self.chunks_registry[chunk_key] = doc.copy()
```
darrylmasson, line 106: Integers as dictionary keys can cause issues later if you try to json-encode it. Probably not an issue here, but something to remember.

darrylmasson (changes requested):
The `DataNotAvailable` exception needs to be moved if we want it to be throwable.

strax/storage/mongo.py
```diff
@@ -73,6 +73,39 @@ def get_metadata(self, key):
...
+        n-chunk queries to the database. Just one will do. As the
+        documents-size is limited to 16 MB, it's unlikely that we will
+        run into memory issues (that we otherwise would not run into).
+        """
+
+        query = backend_key_to_query(backend_key)
+        chunks_registry = self.db[self.col_name].find(
+            {**query, **{"chunk_i": {'$exists':True}}},
+            {"chunk_i": 1, "data": 1})
+        if chunks_registry is not None:
```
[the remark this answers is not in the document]
jorana: Thanks, changed it

strax/storage/mongo.py
jorana: I changed it just to be on the safe side

darrylmasson (approved):

PR #339, opened 2020-10-27
return sorted dataFrame by index when using read()

ahiguera-mx:
Need to return sorted dataFrame by index when using function read()

here is an example of using read() and why is important to return sorted data to avoid confusions 
https://github.com/ahiguera-mx/xenon/blob/master/plot_gains.ipynb

here is an example of using read() and why is important to return sorted data to avoid confusions 
https://github.com/ahiguera-mx/xenon/blob/master/plot_gains.ipynb

  - Update the docstring(s)
  - Update the documentation


jorana:
Hi Aaron, can you please update your master branch before adding this PR? The two are out of sync.

ahiguera-mx:
duplicated PR

Issue #330, opened 2020-10-13
Check size mongo documents in mongo.py


jorana:
Mongo has a maximum document size of 16 MB. We should check that we don't exceed this.

We can additionally add a try: except: loop arround the place where we try to push the data.

jorana:
Sufficiently addressed in https://github.com/AxFoundation/strax/pull/344

Issue #313, opened 2020-09-03
dtype_for is super misleading for the user.


WenzDaniel:
The function dtype_for() of the Plugin class is super misleading for users. They see its usage like in here  https://github.com/XENONnT/straxen/blob/56abca546b2e6f31ff3a49b04513cd6ba3fe00bb/straxen/plugins/peaklet_processing.py#L116

In case of an multi-output plugin it will return the dtype of the specified output. In case of a regular plugin it will just return the dtype of the plugins provides. However the function looks like you could call any dtype simply by specifying its "provides" name.

jorana:
I think the first thing to do would be adding a docstring. Perhaps we could add a renamed function that redirects to this function to prevent incompatibility. We could throw a deprecation warning and delete the function after a given time.

jorana:
I think #344 sufficiently addresses this. It makes actually quite a lot of sense that the plugin object cannot just provide a dtype for something it does not know about.

Issue #307, opened 2020-08-17
Remove S3 and Mongo backends


tunnell:
At least the S3 one, I wrote and isn't used.  I doubt it works.  Backends that aren't tested automatically or understood, should be removed.  Objections?  People can always fetch out of version control.

jorana:
I'm okay with removing S3, mongo we should start using somewhere soon to log things like pulse_counts.

jorana:
Removal of S3 is initiated in #344

jorana:
With #358 in the pipeline I'm closing this issue

PR #344, opened 2020-10-30
Update documentations and adress issues

jorana:
Add docstrings where needed.

Fixes https://github.com/AxFoundation/strax/issues/313 (partially)
Fixes https://github.com/AxFoundation/strax/issues/330 (sufficiently)
Initializes #307 for S3 frontend

### DIFF
--- a/strax/storage/mongo.py
+++ b/strax/storage/mongo.py
@@ -213,14 +213,16 @@ class MongoSaver(Saver):
         # Get the document to update, if none available start a new one
         # for this chunk
         chunk_id = self.ids_chunk.get(chunk_i, None)
+
+        # We can fail here if the document is too large to be written
+        # out to mongo. One could do a try: except
+        # pymongo.errors.WriteError: pass, but that potentially leads to
+        # abuse of a Mongo instance going unnoticed.
         if chunk_id is not None:
-            # We can fail here if the document is too large to be
-            # written out to mongo. One could do a try:
-            # except pymongo.errors.WriteError: pass
-            # but that potentially leads to abuse of a Mongo instance
-            # going unnoticed.
+            # In principle this should not end up here as each chunk
+            # should be it's own document unless you re-chunk
             self.col.update_one({'_id': chunk_id},
-                                {'$addToSet': {f'data': aggregate_data}})
+                                {'$push': {f'data': aggregate_data}})
         else:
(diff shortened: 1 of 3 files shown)


strax/storage/mongo.py
```diff
@@ -208,6 +214,11 @@ def _save_chunk(self, data, chunk_info, executor=None):
         # for this chunk
         chunk_id = self.ids_chunk.get(chunk_i, None)
         if chunk_id is not None:
+            # We can fail here if the document is too large to be
+            # written out to mongo. One could do a try:
+            # except pymongo.errors.WriteError: pass
+            # but that potentially leads to abuse of a Mongo instance
+            # going unnoticed.
             self.col.update_one({'_id': chunk_id},
                                 {'$addToSet': {f'data': aggregate_data}})
```
darrylmasson, line 223: This probably isn't the best place for this discussion, but I've been thinking about our use here of `$addToSet`. This is a relatively "heavy" operator, because it has to check every other element in the list for equality. For cases where we're saving a histogram, there are a lot of elements that need to be checked. I assume it optimizes, but if the two histograms are equal, it's still a _lot_ of operations. I don't remember the details of when Savers get invoked (middle of a chunk? End of a chunk?), but the performance would probably increase if we can find a different operator to use.

darrylmasson (approved):

strax/storage/mongo.py
jorana: Perhaps https://docs.mongodb.com/manual/reference/operator/update/push/ is a better operator looking at the documentation now. There is no need to indeed double check that the data is not stored yet.

Also, in principle I'm actually expecting many documents to be updated here but it depends a little on the re-chunking of documents (disabled by default). In principle each document is just uploaded as soon as the input-chunk has been seen. This means that the $addToSet would not really be used as long as people don't start modifying this default. 

Nonetheless it does not need to check it's written twice so let me just change it to $push

darrylmasson: Rechunking is a tricky subject in the context of the Mongo interface, I think we can make a coherent argument for prohibiting it for this saver. If users are overriding `allow_rechunk=False` for this saver then they probably know what they're doing (or we can at least hope).

jorana:
@WenzDaniel , please don't squash merge this PR as that does not really make sense to bookkeep it as a single commit

PR #345, opened 2020-10-30
ignore checks for savenever plugins

### LINKED ISSUE #247: Time range selection fails if intersecting with some objects
You may get a RuntimeError while loading data with a time range, such as
```
~/software/strax/strax/plugin.py in _fetch_chunk(d, hope_to_see)
    259                         and self.input_buffer[d].end < hope_to_see):
    260                     raise RuntimeError(
--> 261                         f"Tried to get data until {hope_to_see}, but {d} "
    262                         f"ended prematurely at {self.input_buffer[d].end}")
    263                 return False

RuntimeError: Tried to get data until 1518690646999999000, but merged_s2s ended prematurely at 1518690592151168512
```
This happens because the plugin code currently assumes all the dependencies' data streams starts and ends at the same times. The loader's time range selection code cannot guarantee this when the time range intersects some of the objects you are loading.

If you hit this, you can try a slightly larger range until this is fixed.

jorana:
Fix  #247, https://github.com/XENONnT/straxen/issues/149 and prevent an implementation such as https://github.com/XENONnT/straxen/pull/261.

The inconsistency in time ranges may occur when we are loading plugins that are not saved themselves. In https://github.com/XENONnT/straxen/pull/261 I proposed to solve this by making sure that all data was stored. However, @JelleAalbers pointed out that it might be actually nicer to fix this on the strax side.

It took quite some trial and error to get it working. I propose to keep some of the associated Errors as Userwarnings to monitor if we are actually not making any edge case not working. I've tried whatever I could come up with but so far the data returned is exactly the same as the one I would get if we would store all the `save_when.NEVER` plugins (`Peaks` in this case).

**How to replicate**
I just ran the wf plot as in  https://github.com/XENONnT/straxen/issues/149  just until I hit this error, run the following two commands and you will see that you are now able to actually plot the wf:

```python
# where time_range=(9343216270, 9343224420) is a time range in simulated data 
# where from I figured out that I could run into this bug
st.get_array(run_id, 'peaks', time_range=(9343216270, 9343224420)) 
```


WenzDaniel (approved):
I think it is fine, but I think we should monitor this change a bit.

strax/plugin.py
```diff
@@ -409,8 +417,10 @@ def cleanup(self,
         # Check the input buffer is empty
         for d, buffer in self.input_buffer.items():
             if buffer is not None and len(buffer):
-                raise RuntimeError(
-                    f"Plugin {d} terminated with leftover {d}: {buffer}")
+                # This can happen especially in time range selections
+                if int(self.save_when) != strax.SaveWhen.NEVER:
```
JelleAalbers, line 421: Yes, nice, I think this is the key fix. (You could move the if condition up above the for loop for clarity, since the loop doesn't need to run at all for never-save plugins)

strax/plugin.py
```diff
@@ -345,13 +345,21 @@ def iter(self, iters, executor=None):
...
                 max_passes_left = 10
                 while max_passes_left > 0:
+                    if int(self.save_when) == strax.SaveWhen.NEVER:
+                        # So this is a bit funny, this is the only time we may
+                        # be actually running a plugin that is not saved with
+                        # the time range. This means that we might be asking
+                        # for a time range that is inconsistent with the
+                        # chunking of the inputs. See:
+                        # https://github.com/AxFoundation/strax/issues/247
+                        break
```
JelleAalbers, line 361: I think this triggers even during regular processing of a run, without a time range selection. Are you sure it is harmless? Maybe I'm not understanding why this break is needed; did we get the "ten passes" error message when plotting waveforms?

JelleAalbers (commented):
Thanks for the fix Joran! I can remove the item "Fix the strax time range bug, it's only peaks, so just disable exhaustion check" from my todo list :-) I didn't understand the break in the input consolidation section yet.

jorana:
Checkout https://github.com/AxFoundation/strax/blob/d07cc14c32a6673f5365c32085b1362d801ce013/tests/test_loop_plugins.py#L117 when reopening

jorana:
This PR has been validated in https://github.com/XENONnT/analysiscode/blob/master/StraxTests/Fix_time_for_savenever.ipynb and https://github.com/XENONnT/analysiscode/blob/master/StraxTests/Fix_time_for_savenever_validation.ipynb. I'm confident and happy that we can close this extremely long outstanding issue.

strax/plugin.py
jorana: Thanks, this was actually the first place I started looking. This was indeed not necessary and also felt slightly out of place. Fun fact, the input times were inconsistent (not the matter here). 

I removed it now.

strax/plugin.py
```diff
@@ -454,10 +455,27 @@ def do_compute(self, chunk_i=None, **kwargs):
...
             tranges = {k: (v.start, v.end) for k, v in kwargs.items()}
-            if len(set(tranges.values())) != 1:
-                raise ValueError(f"{self.__class__.__name__} got inconsistent "
-                                 f"time ranges of inputs: {tranges}")
             start, end = list(tranges.values())[0]
+
+            # For non-saving plugins, don't be strict, just take whatever
+            # endtimes are available and don't check time-consistency
+            if int(self.save_when) == strax.SaveWhen.NEVER:
+                # </start>This warning/check will be deleted, see UserWarning
```
WenzDaniel, line 463: I assume you mean in a future upcoming release?

WenzDaniel: And you will just keep the `end = max([v.end ....])` part right?

WenzDaniel (approved):
Looks fine and I looked through your tests as well. It looks like you spend a lot of time in setting them up. So I was wondering, do you think it would be possible to boil down your "offending" peak tests into some unity test or similar? Maybe you have already thought about it.

jorana:
> Looks fine and I looked through your tests as well. It looks like you spend a lot of time in setting them up. So I was wondering, do you think it would be possible to boil down your "offending" peak tests into some unity test or similar? Maybe you have already thought about it.

Thanks, actually, this should be hard because you basically need straxen. If I'd write a test it would take a lot of time because you'd be needing to copy over the chunks and set of a similar plugin-structure as in straxen but than with much more abstraction. In the end, the test would not be flexible. 

As a trade-off I can try to see if I can somehow replicate with the demo data. If this is possible I would just move the test to straxen saving myself a lot of time.

strax/plugin.py
jorana: yes, that is true. I just added this warning to give a handle in case we start seeing shenanigans which I highly doubt at this point due to the tests I did.

JelleAalbers (approved):
Very nice work!

PR #347, opened 2020-11-06
Keep self.database or self.client, not both

jorana:
Keep either self.client or self.database so as not to confuse which is used.


ahiguera-mx (approved):
looks good

Issue #346, opened 2020-11-05
Mongo registry may become large


jorana:
In commit https://github.com/AxFoundation/strax/commit/107b2b8c92da455544d04fd517d91f0c8339c9f7 we have changed the building of the plugin class registry. With this we offload queries to the database because the data is only queried once (and then stored in the `chunks_registry`.

Potentially, this might start taking a lot of data as one is querying all of the runs in the collection. We could opt to clear this `chunks_registry` if it exceeds a max number by clearing it if 
```python
sys.getsizeof(self.chunks_registry)/1e6>MAX_BUFF_MB
```

I don't think this is much of an issue as the data itself is very small but we could keep this in mind. @darrylmasson , strong opinions?

darrylmasson:
I see a couple of potential frustrations for an analyst, mainly around how many runs would fit into `MAX_BUFF_MB`. If this is "small", then they might have to frequently wait while the database is queried and the registry rebuilt, if they don't have decent caching in their notebook. Dictionaries remember the order in which keys are inserted, correct? What if the cache isn't strictly cleared when reaching this size threshold, but we start by clearing the earliest entries?

jorana:
So in principle, it's not at all slow if we rebuild the ``chunks_registry`` taking ~a few seconds per run (more or less implied by the 16 MB doc-limit). But yeah, it should be at least larger than the size of a single run, otherwise we might be rebuilding it more than once for a given run which would utterly slow things down.

I like the idea of ` we start by clearing the earliest entries`. We can set a default value of e.g. 500 MB (i.e. just over 2 chunks of other regular data). At the moment a single run Online Monitor -data is ~45 MB for a one hour run. As such we would be able to cache 12 runs.

darrylmasson:
Here's a slightly wilder idea: what if we set the cache to a certain number of runs, rather than MB?

jorana:
Possible, but imagine we are loading the veto_intervals which is only ~O(kB). Here you could cache all of the data that is stored in the collection without any memory issues. The number of runs is not per se a factor that hampers your notebook from not dying, it's the memory you are using. 

A combination of both is of course also possible: first check if we have exceeded 500 MB, then, if there are more than 5 runs cashed in the _backend_, clear the buffer. This of course does not hamper the analyst from using the data (one has obtained this already in st.get_array/df). It's just the backend-buffer that is sitting otherwise idle in the background eating your RAM that needs clearing.

darrylmasson:
[...]

PR #349, opened 2020-11-12
limit buffer for mongo backend

jorana:
Address #346 

Keep track of how many runs (datakinds) are stored in the buffer. If to large, dump some.

**MWE**
```python
for r in range(10165, 10165+10):
    if st.is_stored(f'{r:06}', 'online_peak_monitor'):
        st.get_array(f'{r:06}', 'online_peak_monitor')
        print(st.storage[-1].backends[0]._buffered_backend_keys)
        print(sys.getsizeof(st.storage[-1].backends[0].chunks_registry)/1e6)
```

### DIFF
--- a/strax/storage/mongo.py
+++ b/strax/storage/mongo.py
@@ -15,7 +15,13 @@ from strax import StorageFrontend, StorageBackend, Saver
 from datetime import datetime
 from pytz import utc as py_utc
 from warnings import warn
+from sys import getsizeof
 export, __all__ = strax.exporter()
+
+# Some data is stored in the buffer. Delete when either of these values
+# are exceeded
+DEFAULT_MONGO_BACKEND_BUFFER_MB = 200
+DEFAULT_MONGO_BACKEND_BUFFER_NRUNS = 5
 
 
 @export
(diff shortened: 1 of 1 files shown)


strax/storage/mongo.py
```diff
@@ -115,6 +126,27 @@ def _build_chunk_registry(self, backend_key):
...
+    def _clean_first_key_from_registry(self):
+        """
+        Remove the first item in the self.buffered_keys and all the
+        associated keys in the self.chunks_registry to limit RAM-usage
+        """
+        # only clean the first entry from the list
+        to_clean = self._buffered_backend_keys[0]
+        for registry_key in list(self.chunks_registry.keys()):
+            if to_clean in registry_key:
+                del self.chunks_registry[registry_key]
```
darrylmasson, line 148: Python's complained at me if I modified a dictionary while iterating over it. List-casting the keys probably helps with this, but I got around it by storing the keys to delete in a second list.

darrylmasson (approved):
Nesting a for-loop inside a while-loop might be really inefficient, but probably won't be a real issue unless someone decides to change the defaults.

strax/storage/mongo.py
jorana: you really have a good memory for errors, got to give you that. This is true. 

However if you look carefully we are making a copy of the keys by iterating them when converting to a list. Therefore you are not iterating over the keys but over the list of strings that are extracted

PR #351, opened 2020-11-17
copy data from frontend to frontend

jorana:
Some frontends are harder to interface with than others. Using this new function, one is able to copy data from one frontend to another. This is especially useful in at least these two concrete examples:

1. Querying the database is in Mongo.py is expensive. Better save it to disk as that is quicker and prevents people from overloading the database.
2. The mongo backend is a nightmare to use outside of dali, it's much easier to transfer when backcasted to a normal DataDirectory frontend.

Strax magic! I wrote it such that it can be used either trying to write to all the other frontends or to a specific one (

```python
for sf in st.storage[1:]:
    print(sf)
    print(st._is_stored_in_sf(run_id, target, sf))
st._copy_to_frontend(run_id, target, target_frontend_id=3)
for sf in st.storage[1:]:
    print(sf)
    print(st._is_stored_in_sf(run_id, target, sf))
## output
strax.storage.files.DataDirectory, path: /data/real_data/
True
strax.storage.files.DataDirectory, path: /data/_tests8/
False
strax.storage.files.DataDirectory, path: /data/_tests9/
False
# After copy
strax.storage.files.DataDirectory, path: /data/real_data/
True
strax.storage.files.DataDirectory, path: /data/_tests8/
False
strax.storage.files.DataDirectory, path: /data/_tests9/
True
```

**Future requests**
One might think that we may just as well then erase the original copy. This is is however not per se trivial in strax (yet), I might think about if we need this at some point.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1323,13 +1323,10 @@ class Context:
             # noinspection PyMethodFirstArgAssignment
             self = self.new_context(**kwargs)
 
-        key = self.key_for(run_id, target)
         for sf in self.storage:
-            try:
-                sf.find(key, **self._find_options)
+            if self._is_stored_in_sf(run_id, target, sf):
                 return True
-            except strax.DataNotAvailable:
-                continue
+        # None of the frontends has the data
         return False
 
     def _check_forbidden(self):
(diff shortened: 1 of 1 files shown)


strax/context.py
```diff
@@ -1361,6 +1358,106 @@ def _apply_function(self, data, targets):
             data = function(data, targets)
         return data
 
+    def _copy_to_frontend(self, run_id, target,
```
jorana, line 1361: Perhaps this should not be `_`-ed. If the reviewer has an opinion I'm interested

strax/context.py
```diff
@@ -1361,6 +1358,106 @@ def _apply_function(self, data, targets):
...
+        :param target: target datakind
+        :param target_frontend_id: index of the frontend that the data should go to
+        in context.storage. If no index is specified, try all.
+        :param rechunk: allow re-chunking for saving
+        """
+        if not self.is_stored(run_id, target):
+            raise strax.DataNotAvailable(f'Cannot copy {run_id} {target} since it '
+                                         f'does not exist')
+        if len(strax.to_str_tuple(target)) > 1:
+            raise ValueError('_copy_to_frontend only works for ')
```
WenzDaniel, line 1375: The message is missing something

strax/context.py
```diff
@@ -1361,6 +1358,106 @@ def _apply_function(self, data, targets):
...
+        if target_frontend_id is None:
+            target_sf = self.storage
+        elif len(self.storage) > target_frontend_id:
+            # only write to selected other frontend
+            target_sf = [self.storage[target_frontend_id]]
```
WenzDaniel, line 1380: I am wondering if it would not be a bit neater to add a `fornt_end_id` to the properties of every [forntned](https://github.com/AxFoundation/strax/blob/3dece573fdfd44955cc69d95dbdd790451f3d933/strax/storage/common.py#L120). In that case one could precisely distinguish between frontends. However, this would also require some additional checks, e.g. that the same id is not specified twice.

strax/context.py
WenzDaniel: On the one hand I agree with you that this function should just be used by a very limited amount of people who know what they are doing and hence make the function not "visible". On the other hand I think it is very unlikely that anyone will use this function except for the experts. My/our concerns are very limited, and the underscore usually indicates that the function is a private member, I agree with your statement to remove it.

WenzDaniel (approved):
Look good. Thanks a lot.

strax/context.py
jorana: whoop, thanks will add

strax/context.py
jorana: What would be the gain? Now it's just the index in st.storage which is a list. Using a number is quite simple. 

Could try to match the __repr__ or the input but that sounds more convoluted. Could do though because you than don't by accident fill the wrong sf.

WenzDaniel: I agree, I think it is not really needed and I am fine with the PR as is. 

> Could do though because you than don't by accident fill the wrong sf.

This was also my concern, but I am not sure if the "risk" justifies the extra work.

darrylmasson (approved):
Daniel beat me to a few comments

Issue #348, opened 2020-11-07
Unsafe division in hitlets.py


jorana:
In https://github.com/AxFoundation/strax/blob/master/strax/processing/hitlets.py#L405 and line 418 there are divisions by `m` who's value can be 0. 

This causes crashes which may disappear in the numba magic. Perhaps @WenzDaniel can have a look.

PR #355, opened 2020-11-20
Muveto + hitlet fixes

WenzDaniel:
This PR changes two things: 

1. It adds the possibility to change the settings ending of a child compared to the parent. Example `threshold_parentending` can now be changed into  `threshold_childending` instead of  `threshld_parentending_childending`. I also added some additional comments in the code to make things a bit easier to understandable. 
2. In this PR we are making some bug fixes to hitlets.py. The fixes address the `divided by zero` bug in `fwxm` which can occur for some  irregular data. The corresponding test is also updated. Further, I restructured the function a bit for better readability/logic.


jorana:
@WenzDaniel , let me know when ready. I would appreciate if you un-draft when this is the case.

strax/context.py
```diff
@@ -474,7 +480,18 @@ def get_plugin(data_kind):
                 # drop the parents config from the lineage
                 configs = {}
                 for q, v in p.config.items():
-                    if q + p.child_ends_with in p.takes_config:
+                    # Looping over all settings, q is either the option name of the
```
jorana, line 483: perhaps you can juse parent_option_name instead of q. Makes the code much more readable

strax/processing/hitlets.py
```diff
@@ -244,6 +244,7 @@ def get_single_hitlet_data(hitlet, records, prev_r, next_r):
             r = records[next_r_i]
             data, (start, p_end_i) = _get_thing_data(hitlet, r)
             if not start:
+                print(data, start, p_end_i)
```
jorana, line 247: Intended?

strax/processing/hitlets.py
```diff
@@ -343,29 +344,31 @@ def hitlet_properties(hitlets):
     Computes additional hitlet properties such as amplitude, FHWM, etc.
     """
     for h in hitlets:
-
+
```
jorana, line 347: nano small, perhaps delete

strax/processing/hitlets.py
```diff
@@ -388,31 +391,36 @@ def get_fwxm(hitlet, fraction=0.5):
...
+    if len(pre_max) and np.any(pre_max <= max_val):
+        # First the left edge:
+
+        lbi, lbs = _get_fwxm_boundary(pre_max[::-1], max_val)  # Reversing data starting at sample
+        # before maximum and go left
+        lbi = (index_maximum - 1) - lbi  # start sample minus samples we went to the left
         m = data[lbi + 1] - lbs  # divided by 1 sample
-        left_edge = lbi + (max_val - lbs) / m + 0.5  # .5 to start from bin center
+        left_edge = lbi + (max_val - lbs) / m + 0.5
```
jorana, line 411: I don't see you addressed https://github.com/AxFoundation/strax/issues/348 or maybe I missed it. Can you also point me to the test that should be able to catch it in the future?

strax/processing/hitlets.py
```diff
@@ -388,31 +391,36 @@ def get_fwxm(hitlet, fraction=0.5):
...
 
+    if len(post_max) and np.any(post_max <= max_val):
         # Now the right edge:
-    rbi, rbs = _get_fwxm_boundary(post_max[::-1], max_val)  # coming from the right
-    if rbi == NO_FWXM:
-        right_edge = len(data)
+        rbi, rbs = _get_fwxm_boundary(post_max, max_val)  # Starting after maximum and go right
+        rbi += 1 + index_maximum  # sample to the right plus start
+        m = data[rbi - 1] - rbs
+        right_edge = rbi - (max_val - rbs) / m + 0.5
```
jorana, line 421: same here

jorana (approved):
I think I'm fine with this PR. The `overwrite_parents_end` feels a bit patchy and also don't know if such an option would be useful for other experiments. I think another way around would be to have the neutron veto plugins `child_ends_with = _nv` and have the plugin recognize that we we should replace that with `_mv`.

Haven't tested the code itself, let me know if you think I need to try to break it.

strax/context.py
WenzDaniel: Sure I will change it. I just stick to the previous notation.

strax/processing/hitlets.py
WenzDaniel: :D I will change it

strax/processing/hitlets.py
WenzDaniel: Yes, but I agree I should put some additional text.

strax/processing/hitlets.py
WenzDaniel: Okay sorry, valid point I should have elaborated this a little bit more. I changed a bit the way how the the values for `lbi`and `lbs` are found. In the previous version it was possible  that  `data[lbi + 1] == lbs` are the same . This is now excluded, partially due to a change on how these values are found and partially due to some additional constraints to the data in case the FWHM is not defined.

WenzDaniel:
> The overwrite_parents_end feels a bit patchy and also don't know if such an option would be useful for other experiments.

I totally agree on that point and I thought the very same. Let me give it a second thought maybe I come up with a better solution. The main issue here, the child plugin does not know anything about the naming convention of the parent. And I did not want to put any constrained on the naming of the different options (e.g. double underscore indicates an ending etc.). 


But let me try your suggestion, maybe I get another idea.

strax/processing/hitlets.py
jorana: Thanks, are they tested against explicitly?

jorana:
Thanks for the changes, haven't looked at them very detailedly but I think this will require some changes on straxen. Can you explain what we will need to change, will e.g. _he channels now need both attributes? When will you need to update which?

Also, perhaps this is good to add to the documentation.

Finally, I think the tests are also not using the convention used here. I can imagine that a deprecation-phase might make sense as otherwise everything starts breaking at once.

WenzDaniel:
Sorry my fault I should have turned it into a draft again. You are right, it requires some more modifications also in https://github.com/XENONnT/straxen/pull/287

strax/processing/hitlets.py
WenzDaniel: Now it is tested explicitly.

WenzDaniel:
Hej Joran, this morning I had an idea how to simplify the whole child_plugin business. I removed the whole thing about `ends_with` and replaced it with the following three things:

Plugin side:
A plugin which inherits the compute statement from another plugin has to be specified as a child plugin (`child_plugin=True`).
Option side:
Since we cannot change the keys of `self.config['key']` in the `super.compute()` statement we have to overwrite the corresponding options according to the new options  specified by the child-plugin. This issue is the same as before. However, this is now done slightly differently on the option side. All options of a child plugin which should overwrite the option of the parent plugin has to have the following two attributes: 

* `child_option=True` showing that the option is a child option which overwrites a parent  
* `parent_option_name=string` Name of the parent config to find it and overwrite it 

Although, `child_option` is a redundancy I kept it since it adds an additional layer of security fort he user.  The lineage of the child is the same as in previous versions. All straxen plugins have been updated accordingly in https://github.com/XENONnT/straxen/pull/287.

I also updated the documentation and tests. Further, I addressed your request of adding a test for the `divided by zero`-error in `get_fwxm`.

jorana:
Thanks Daniel! This looks much cleaner, thanks a lot! To put the icing on the cake, perhaps can you also update the https://strax.readthedocs.io/en/latest/advanced/plugin_dev.html?

I think it deserves a separate heading as it's quite different than all the other types. In fact there is no such thing as a `strax.ChildPlugin`. You would really make my day if you can just code a minimal working example there!

The travis failing is odd on 3.6, it claims numpy incompatibility between pandas an py36. Don't know if it will persist. Perhaps we need to fix numpy before pandas as a requirement.

WenzDaniel:
Okay, I added an example and some more words.

jorana (approved):
Thanks a lot. It's really nice that you took the effort to write such a clear example. 

Would you say that it makes more sense to put `Plugin types` before `Plugin inheritance`. I would be inclined to argue so but its up to you, feel free to merge.

WenzDaniel:
I was thinking the very same to be honest :D But was not sure either. I will change it

PR #356, opened 2020-11-23
use utilix for client

jorana:
Change the way we init the pymongo client for the corrections. NB: the init is now done on the `straxen` side (makes more sense to do it in the nT specific package).

Given recent changes (https://github.com/XENONnT/straxen/pull/163), it makes more sense to use a centralized initiation for the pymongo client (i.e. `utilix`). 

All should work just as before. It's only when someone uses this script outside of the straxen environment that one should initialize a client (see https://github.com/XENONnT/straxen/pull/288).

### DIFF
--- a/strax/corrections.py
+++ b/strax/corrections.py
@@ -28,19 +28,20 @@ class CorrectionsInterface:
     this means a unique set of correction maps.
     """
 
-    def __init__(self, client=None,
-                 host=None, username=None, password=None,
-                 database_name='corrections'):
+    def __init__(self, database_name='corrections', client=None,
+                 host=None, username=None, password=None,):
         """
         Start the CorrectionsInterface. To initialize you need either:
-         - a pymongo.MongoClient (add as argument client) OR
-         - the credentials and url to connect to the pymongo instance
-         one wants to be using.
+            - a pymongo.MongoClient (add as argument client) OR
+            - the credentials and url to connect to the pymongo instance
+            one wants to be using.
+        :param database_name: Database name
         :param client: pymongo client, a pymongo.MongoClient object
+
+        (optional if client is not provided)
         :param host: DB host or IP address e.g. "127.0.0.1"
-        :param username: DB username
-        :param password: DB password
-        :param database_name: DB name
+        :param username: Database username
+        :param password: Database password
         """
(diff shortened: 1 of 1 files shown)


jorana:
Also trying to tag Evan here but cannot do because he is not part of the Axfoundation

ahiguera-mx:
I'm fine with this change. Just want to express some small comments.  To me it is more transparent to the user specify every object individually and since utilix is a straxen specific not sure this goes along with strax's philosophy of being experiment agnostic

jorana:
So if you just know about strax, the `client` can be anyone's, a client you get from admix, a client you construct yourself, whatever so it's not experiment specific at all.

If you think this too abstract, I can also propose this:
```python
    def __init__(self, client=None, host=None, username=None, password=None, database_name='corrections'):
        """
        :param client: DB pymongo client
        <update>
        :param database_name: DB name
        """
        if (client is not None) and (host is None and username is None and password is None):
             self.client = client
        elif (client is None) and (host is not None and username is not None and password is not None):
             self.client = pymongo.MongoClient(host=host, username=username, password=password)
        else:
              raise ValueError('Can only init using *either* the client or the combination of'
                               'host+username+password, not both')
              
        self.database_name = database_name
```
if this makes more sense to you.

WenzDaniel:
Since you have it already available, maybe the second option is a bit more flexible while not much more complex (I would also have been fine with the first approach). May I just ask you to be a bit more explicit with the description of the "client" argument. Because we are very explicitly expecting here an instance ala `pymongo.MongoClient`. I am not sure if there are any other pymongo libraries which could lead to misunderstandings, but more documentation never hurts.

ershockley:
Thanks for the PR and the invite to the organization @jorana. I personally like the simplicity of the original PR, though can see the merits of having an either/or implementation. I agree with Daniel on the documentation.

jorana:
Thanks @ahiguera-mx @ershockley @WenzDaniel for the review. I changed the PR accordingly. Please add a positive review if you think the concerns have been appropriately addressed.

WenzDaniel (approved):
Looks nice, thanks a lot.

ahiguera-mx:
thanks, I'm fine with the changes

ahiguera-mx (approved):

PR #357, opened 2020-11-23
Allow loop_plugin for multi_output

jorana:
Allow loop-plugins to be used with multiple outputs.

I tried to write a new loop plugin, but this yielded quite an incomprehensible traceback that the dtype was wrong. We should at least make a better error. So I just modified the loop plugin to just work in this case. There is some copy pasting but given it's only 10 lines, I don't care too much.

The only downside is that, while solving this, I realized I don't actually need it.

Loop over assumed a single output, when a dict is returned for the different outputs, it got confused. Especially when setting up the results buffer.

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -650,22 +650,12 @@ class LoopPlugin(Plugin):
                                        f"should be {len(base)}!")
                 kwargs[k] = r
 
-        if not self.multi_output:
-            results = np.zeros(len(base), dtype=self.dtype)
-            deps_by_kind = self.dependencies_by_kind()
-
-            for i in range(len(base)):
-                r = self.compute_loop(base[i],
-                                      **{k: kwargs[k][i]
-                                         for k in deps_by_kind
-                                         if k != loop_over})
-
-                # Convert from dict to array row:
-                for k, v in r.items():
-                    results[i][k] = v
-        else:
-            # Just doing the same as above but this time we need to
-            # create a dict for the outputs.
+        if self.multi_output:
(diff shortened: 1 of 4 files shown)


strax/plugin.py
```diff
@@ -648,18 +648,35 @@ def compute(self, **kwargs):
...
-        for i in range(len(base)):
-            r = self.compute_loop(base[i],
-                                  **{k: kwargs[k][i]
-                                     for k in deps_by_kind
-                                     if k != loop_over})
-
-            # Convert from dict to array row:
-            for k, v in r.items():
-                results[i][k] = v
+        if not self.multi_output:
```
WenzDaniel, line 651: Minor comment can we revert the logic, `if self.multi_output:` is a nano bit simpler.

WenzDaniel (commented):
In general I think it is fine, though I think this is one of the fundamental changes which should be covered with a test. Given that this PR is not urgent may I ask you to add one, when there is time. I can also check and add something when I have a free minute this week.

WenzDaniel:
Btw, you reached the next level solving problems by accident without need :D

strax/plugin.py
jorana: Sure, if it makes you but a nano bit more happy, I'm completely fine with it. Just though most people read from top to bottom and no one has been having `self.multi_output` issues .

jorana:
The tests are a good point, will put this in the refrigerator for a bit and make the tests (and undraft)

jorana:
Am I missing something or is it really strange that the coverage does not increase after I just spend time on writing a unit test?

PR #358, opened 2020-11-30
remove s3

jorana:
S3 is not used and should therefore not be part of strax.

See #307


jorana:
Admin merging, I think Chris' approval is implied in #307

PR #362, opened 2020-12-07
More clear print statement.

WenzDaniel:
Update of print statement object ids can be confused with unix times.


jorana (commented):
I never considered it to be possible to be a timestamp and looked at this message a million times. On the other hand, it does not hurt to add this change.

jorana (approved):

Issue #361, opened 2020-12-07
NB breaks test should be reproducible


jorana:
The natural breaks test just failed in the base env. The test just provided a failed assertion statement where it should at least provide the failing example.

jorana:
Should be fixed in https://github.com/AxFoundation/strax/pull/363

PR #363, opened 2020-12-07
Fix reproducibility of peak split test

jorana:
Hotfix to fix issue https://github.com/AxFoundation/strax/issues/361. This PR solves two issues:
1. Makes the test reproducible
2. Disables check that is only working for local_minimum

This 2.nd issue was making some builds fail on travis. Did not catch that before.


WenzDaniel (approved):
Looks good. Now we have full control/feedback about the tested waveforms. Just out of curiosity a question though. Any specific reasons why you have not used the numpy support of hypothesis: https://hypothesis.readthedocs.io/en/latest/numpy.html#numpy

jorana:
> Looks good. Now we have full control/feedback about the tested waveforms. Just out of curiosity a question though. Any specific reasons why you have not used the numpy support of hypothesis: https://hypothesis.readthedocs.io/en/latest/numpy.html#numpy

Probably I spend to little time searching. Thanks for the link!

PR #367, opened 2020-12-17
Baseline pr

FaroutYLq:
We are always losing the fractional part of the baseline values for `record_i!=0`! 

To solve this, I modified line 130 of `/strax/processing/pulse_processing.py` from:
```
    last_bl_in = np.zeros((n_channels, 2), dtype=np.int16)
```
to
```
    last_bl_in = np.zeros((n_channels, 2), dtype=np.float32)
```
For more details like how I found the problem, please check the enclosed file. 
[baseline-pr description.pdf](https://github.com/AxFoundation/strax/files/5708175/baseline-pr.description.pdf)


WenzDaniel:
Hej @FaroutYLq thank you very much for this through check and nice catch. Your proposed change is very simple and easy to approve ;)

WenzDaniel (approved):

jorana:
@FaroutYLq , many thanks for the complete document. We will have to carefully implement this in straxen and up the version of the `pulse_processing` in straxen!

jorana (approved):

jorana (changes requested):
Hi @FaroutYLq , just noticed it now, can you remove the ` .DS_Store` file?

FaroutYLq:
Hi @jorana. Sorry for the .DS_Store stuff. I just committed another one. Is the problem solved now? (Sorry for the silly question...)

tunnell:
Nice catch.  Why float32?  As we have a lot of this data, could we use float16 for instance?  I'm just wondering if this PR increases our data volume.

(Also, @sophiaandaloro you may be interested in knowing this)

jorana:
@tunnell , no right? It's just used as an intermediate calculation variable and then cast into the `records`-`dtype`.

tunnell:
OK.  I didn't dig in beyond seeing it was a type change, but just wanted to check that this wasn't somehow saved somehow.  Good catch

jorana (approved):

PR #371, opened 2020-12-27
Speed up st.select_runs by factor ~100 with strict "_find" in files-storage

jorana:
**What is the problem / what does the code in this PR do & can you briefly describe how it works?**
You might have noticed that `st.select_runs` is getting slow lately. After some nice profiling, I traced back the issue to the operation I'm changing in this PR. For each run we could not find in the rundb (rucio) we used the [strax.DataDirectory](https://github.com/XENONnT/straxen/blob/master/straxen/contexts.py#L96) in the context. These in turn look for the keys not found yet. However, there is a very intensive check in `DataDirectory._find` that loops over all the files in the `DataDirectory` **for each of the missing keys!** This makes everything very very slow and without having to do so. This check makes sense only if one does a fuzzy search. Adding the requirement that this is actually a fuzzy search speeds up things a lot, see the comparison below.

E.g. due to a recent version bump in `PulseProcessing` we noticed that it takes forever to execute `st.select_runs` (not much available if you version bump at low level.) This meant a lot of redundant looping over files that were clearly not the file that was requested.

*before*:
afbeelding

*after*
afbeelding

### DIFF
--- a/strax/storage/files.py
+++ b/strax/storage/files.py
@@ -12,7 +12,6 @@ from .common import StorageFrontend
 
 export, __all__ = strax.exporter()
 
-
 RUN_METADATA_PATTERN = '%s-metadata.json'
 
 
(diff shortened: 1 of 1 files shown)


WenzDaniel (approved):
A nice Christmas present. Thanks a lot Joran.

Issue #372, opened 2020-12-29
Key error for multiprocessing if not making raw_records


jorana:
There is an annoying key error in multiprocessing when raw_records, is already stored. See the example below:

```python
>>>st_mp.make(run_id, 'event_info_double', max_workers = 10)
Removing old incomplete data in ./strax_data_mp/009104-veto_regions-pamvninlop
Removing old incomplete data in ./strax_data_mp/009104-records-pamvninlop
Removing old incomplete data in ./strax_data_mp/009104-pulse_counts-pamvninlop
Removing old incomplete data in ./strax_data_mp/009104-peaklets-ioudwhzz4n
Removing old incomplete data in ./strax_data_mp/009104-lone_hits-ioudwhzz4n
Removing old incomplete data in ./strax_data_mp/009104-peaklet_classification-nnlrqza4wr
Removing old incomplete data in ./strax_data_mp/009104-merged_s2s-4os2w32txl
Removing old incomplete data in ./strax_data_mp/009104-peak_basics-3k6xvmmrww
Removing old incomplete data in ./strax_data_mp/009104-peak_proximity-npglcajfmc
Removing old incomplete data in ./strax_data_mp/009104-events-zf5ybclhvk
Removing old incomplete data in ./strax_data_mp/009104-peak_positions-5ihezjokah
Removing old incomplete data in ./strax_data_mp/009104-event_basics-ahgsk5qm5w
Removing old incomplete data in ./strax_data_mp/009104-event_positions-dmugiittv4
Removing old incomplete data in ./strax_data_mp/009104-corrected_areas-nkrgct37wh
Removing old incomplete data in ./strax_data_mp/009104-energy_estimates-pztdnyajof
Removing old incomplete data in ./strax_data_mp/009104-event_info-j6jc4vuyjl
Removing old incomplete data in ./strax_data_mp/009104-distinct_channels-wj55i6dlbb

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-19-d0db1509c592> in <module>
----> 1 st_mp.make(run_id, 'event_info_double', max_workers = 10)

/mnt/d/Google_Drive/PhD-master/ubuntu-storage/ubuntu-windows/software/strax_official/strax/context.py in make(self, run_id, targets, save, max_workers, progress_bar, _skip_if_built, **kwargs)
   1097             return
   1098 
-> 1099         for _ in self.get_iter(run_ids[0], targets,
[...]

PR #376, opened 2021-01-04
fix key error for non rr multiprocess

jorana:
See issue #372 for the problem.

When we create a `ParallelSourcePlugin`, there is some error in finding the `dtype` for a requested data type. This yields the error as in #372.

Usually this is not much of a problem but if we already have raw_records registered, this becomes a mess. I now just iterate over the known plugins to the `ParallelSourcePlugin` to find the right dtype. Given that we have been running like this for a long time illustrates that there are probably few people actually using strax to it's max performance and few will notice.

I'll upload a `multiprocess problem.ipynb` to the StraxTests on the analysiscode repo.

**Tests?**
Nah don't think such a think is likely to occur again any time soon.

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -886,8 +886,19 @@ class ParallelSourcePlugin(Plugin):
         p.sub_savers = sub_savers
         p.start_from = start_from
         if p.multi_output:
-            p.dtype = {d: p.sub_plugins[d].dtype_for(d)
-                       for d in outputs_to_send}
+            p.dtype = {}
+            for d in outputs_to_send:
+                if d in p.sub_plugins:
+                    p.dtype[d] = p.sub_plugins[d].dtype_for(d)
+                else:
+                    log.debug(f'Finding plugin that provides {d}')
+                    # Need to do some more work to get the plugin that
+                    # provides this data-type.
+                    for sp in p.sub_plugins.values():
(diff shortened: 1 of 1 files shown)


strax/plugin.py
```diff
@@ -886,8 +886,19 @@ def inline_plugins(cls, components, start_from, log):
...
+                    p.dtype[d] = p.sub_plugins[d].dtype_for(d)
+                else:
+                    log.debug(f'Finding plugin that provides {d}')
+                    # Need to do some more work to get the plugin that
+                    # provides this data-type.
+                    for sp in p.sub_plugins.values():
+                        if d in sp.provides:
+                            log.debug(f'{sp} provides {d}')
+                            p.dtype[d] = sp.dtype_for(d)
+                            break
```
WenzDaniel, line 901: Have you checked what happens if you cannot find `d` in `sp.provides.values()`? Is the traceback already enough or shall we just add a small raise condition after the for-loop?

WenzDaniel (approved):
Looks fine, thanks

strax/plugin.py
jorana: If that happens, there is some deeper bug in strax because it asks you to produce a datatype that is not provided by any of the multi-process plugins, that is troublesome.

Issue #238, opened 2020-03-04
Bug? plugin return with a single key dict leads to ValueError:


WenzDaniel:
Hej I just noticed that if a plugin provides only a single output as a one key dict instead of a simple return, `strax.utils.dict_to_rec()` throws the following error:

```
/strax/strax/utils.py in dict_to_rec(x, dtype)
    404     for k, v in x.items():
--> 405         r[k] = v
    405     return r
    407 

ValueError: no field of name nveto_records
```
in this case the plugin should have provided nveto_records. Not sure if this is intended.

JelleAalbers:
Yeah, this is confusing: returning dictionaries for a multi-output plugin is mandatory (to separate the outputs), for a single-output plugin a dictionary output is seen a shorthand to specify a record array. See e.g. https://github.com/XENONnT/straxen/blob/master/straxen/plugins/event_processing.py#L283.

We should certainly throw a better error message if this situation happens.

jorana:
OK, so you should wrote something like (pay attention to the `compute`-`return`):
```python
class PeaksProxProxy(strax.Plugin):
    depends_on = 'peak_proximity'
    provides = 'peak_proximity_prox'
    __version__ = '0.0.0'
    
    def infer_dtype(self):
        prox_dtype = [straxen.plugins.PeakProximity.dtype[0] ,
                      straxen.plugins.PeakProximity.dtype[-1],
                      straxen.plugins.PeakProximity.dtype[-2] ]
        dtype = prox_dtype
        return dtype
    
    def compute(self, peaks):
        res = np.zeros(len(peaks), self.dtype)
        for k in self.dtype.names:
            res[k] = peaks[k]
        ###
         # wrong
        return {peak_proximity_prox:res}
        ###
```
This should have been something like:
```python
class PeaksProxProxy(strax.Plugin):
    depends_on = 'peak_proximity'
    provides = 'peak_proximity_prox'
    __version__ = '0.0.0'
    
    def infer_dtype(self):
        prox_dtype = [straxen.plugins.PeakProximity.dtype[0] ,
                      straxen.plugins.PeakProximity.dtype[-1],
                      straxen.plugins.PeakProximity.dtype[-2] ]
        dtype = prox_dtype
        return dtype
    
    def compute(self, peaks):
        res = np.zeros(len(peaks), self.dtype)
        for k in self.dtype.names:
            res[k] = peaks[k]
        ### 
        # correct
        return res
        ###
```

Closing this issue since it is fixed in #377.

PR #377, opened 2021-01-04
fix issue #238 for single key-word returns

jorana:
Make a clear error message to notify the user. I also once or twice ran into this error so better to solve it.

See issue #238 for examples


WenzDaniel (approved):
Thanks

PR #382, opened 2021-01-18
Added keep_columns into docstring

WenzDaniel:
Adds keep_columns to iter/get_array doc strings. I am not sure if I am recalling it correclty, but I think we had this doc-string already before and lost it at some point. But, I wa snot able to track down the commit in which it happened.


jorana (approved):
Sure why not.

Issue #229, opened 2020-02-07
Deterministic/pinned test environment


JelleAalbers:
Strax is a library that should support a wide range of dependency versions. Still, it would be good to run the integration tests against a specific set of pinned dependencies. This gives a known-good configuration and prevents sudden build failures due to breaking updates upstream.

@tunnell implemented this a while ago, with PyUP keeping the dependencies up to date. Since then some things have changed: PyUP seems to have died, and we made a new repository [base_environment](https://github.com/XENONnT/base_environment) to host a deterministic environment for XENON.

We have a few options:
  * Switch to [dependabot](https://dependabot.com/) just like [base-environment](https://github.com/XENONnT/base_environment/pull/48). Unfortunately, that means we'll get two sets of 5-10 pull request updates a week. If we do the same for other repos like straxen and wfsim we'll have three or four sets of them.
  * Fetch the pins from base_environment's [requirements.txt](https://github.com/XENONnT/base_environment/blob/master/requirements.txt) during continuous integration.
  * Rethink thinks more fundamentally and investigate the new generation of python packaging and dependency resolution tools, like pipfile locks, pipenv, poetry, etc.

jorana:
Hi Jelle,
I think the first option has also fundamental issues that it will prohibit the  base_env from actually using dependabot successfully. If we fix everything here, it will become difficult for the base_env to fix the more complex dependencies.

JelleAalbers:
Good point Joran, I guess this relates to your question at https://github.com/XENONnT/base_environment/issues/404. 

Maybe the easiest fix is the second option then, i.e. have a small script run during strax's travis build:
  * wget/curl https://raw.githubusercontent.com/XENONnT/base_environment/master/requirements.txt,
  * parse it with e.g. https://pypi.org/project/requirements-parser
  * same for strax's requirements.txt,
  * Output a new requirements.txt with strax's requirements and the version pins from base_environment (if any, and if the strax branch being built specifies a version pin in the requirements.txt, that should take precedence) 
  * Then you'd install the new requirements.txt here https://github.com/AxFoundation/strax/blob/master/.travis.yml#L46 and proceed with the build as normal

jorana:
Hi Jelle, yeah thanks I was thinking something similar, perhaps slightly less elegant but brutely simple:
```bash
# Get requirements from the base_env
wget -O pre_requirements.txt https://raw.githubusercontent.com/XENONnT/base_environment/master/requirements.txt
[...]

PR #383, opened 2021-01-19
Parse requirements for testing

jorana:
Address https://github.com/AxFoundation/strax/issues/229

Rather than figuring out e.g. numpy dependencies and compatibility in the requirements for strax, leave that to the base-environment (unless something is truly incompatible). Parse the requirements as set there (are tested against strax and straxen) such that we don't have to do it here. 

The way we set it up now is cyclically unlogical, causing failures like: https://github.com/XENONnT/base_environment/pull/444

This approach has been working perfectly for straxen:
https://github.com/XENONnT/straxen/commit/bde58e97fe2487a6abc5aa62879d344630a4a4e1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,8 +42,17 @@ install:
   - source activate strax
   - conda install -c conda-forge python-blosc      # Recommended way to install blosc
 
-  - pip install coveralls --use-feature=2020-resolver
-  - pip install -r requirements.txt --use-feature=2020-resolver
+  - echo "download requirements from base_environment"
+  - wget -O pre_requirements.txt https://raw.githubusercontent.com/XENONnT/base_environment/master/requirements.txt &> /dev/null
+  - echo "select important dependencies for strax(en)"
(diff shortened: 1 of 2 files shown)


WenzDaniel:
I do not remember were the pandas and numpy requirements imposed by straxen?

jorana:
Nope, numpy was imposed since pip installed a pre-release of 1.20 which is incompatible with py36. Pandas was imposed in a similar way. There was a whole train of commits related to this in strax and straxen, which fueled my frustration: https://github.com/XENONnT/base_environment/issues/404

This PR should put an end to this tuning of (in)compatibilities and leave such decisions to the user or dedicated software: https://github.com/XENONnT/base_environment/

We should here (in strax' requirements) just concern ourself with figuring out which version is compatible with the *code* rather than with what package and what not each dependency is compatible with.

WenzDaniel:
> We should here (in strax' requirements) just concern ourself with figuring out which version is compatible with the code rather than with what package and what not each dependency is compatible with.

Agreed, this is why I was asking.

PR #384, opened 2021-01-19
Nveto changes

WenzDaniel:
In this PR we make some optimizations to the hitlet data_type class. 

* We roughly halve the amount of Bytes stored per hitlet:
* We reduced the area width and area decile fields to a 50% and 80% width and 10% and 80% decile. This turns out to be sufficient and allows to define rise and fall time
* We added a new type of width which is estimated based on the [highest density region](https://stats.stackexchange.com/questions/148439/what-is-a-highest-density-region-hdr) of a distribution/signal. Since there was not any suited function I added one in prcoessing/statistics.  Also here we keep always the left edge and width (50 % and 80% coverage) to compute rise and fall time if needed. See also the corresponding [wikinote](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:wenz:comissioning:nveto:hitlet_width)
    
This PR goes together with: https://github.com/XENONnT/straxen/pull/319


jorana:
> **What is the problem / what does the code in this PR do**
> In this PR we make some optimizations to the hitlet data_type class.
>     * We added a new type of width which is estimated based on the [highest density region](https://stats.stackexchange.com/questions/148439/what-is-a-highest-density-region-hdr) of a distribution/signal. Since there was not any suited function I added one in prcoessing/statistics.  Also here we keep always the left edge and width (50 % and 80% coverage) to compute rise and fall time if needed. See also the corresponding [wikinote](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:wenz:comissioning:nveto:hitlet_width)

Hi Daniel, thanks for the note and the explanation. Like you also mention, it's nice to have somewhat consistent usage among NV & TPC, the downside for noisy wfs is granted but wouldn't you want to cut those anyway. I'm asking because I'd hate if in the end no one understand why there are two different widths in the hitlets.

Sorry for also asking it, but we should also really add tests for this. It's super easy to make a little mistake in abstract code.

strax/processing/statistics.py
```diff
@@ -0,0 +1,101 @@
...
+
+        # Check if this height step exceeded at least one of the desired
+        # fractions
+        m = fractions_desired[fi:] <= f
+        if np.any(m):
+            for fd in fractions_desired[fi:fi + np.sum(m)]:
+                # Compute correct height for which we reach desired
+                # fraction and store:
+                g = fd / f
+                h = (1 - g) * np.sum(d) / j + g * lowest_point
```
jorana, line 61: h = (1 - g) * np.sum(d) / j + g * lowest_point

at this point, I've lost what I'm reviewing. Perhaps using less indents and a little clearer variable names helps

WenzDaniel:
Why should not we allow different parameters for a different type of signal/detector? I do not really understand this argument. In the end we have to see which parameter gives the best results/discrimination power and based on what I have seen so far the area deciles did not perform so well on our SPE signals. 

> Sorry for also asking it, but we should also really add tests for this. It's super easy to make a little mistake in abstract code.

Valid point, but I simply had not any time yet to add those.

jorana (commented):
Hi Daniel, thanks, it's very tarse material to review. I would argue it cannot do without tests but at this stage I'm not finding any obvious mistakes.

strax/processing/statistics.py
WenzDaniel: Sure can change it

jorana:
> Why should not we allow different parameters for a different type of signal/detector? I do not really understand this argument. 
Sorry, this was not what I was trying to say. Was just wondering about the noisiness you describe in your note etc. I think it's really nice if we can have multiple algorithms in strax!

WenzDaniel:
Hej, as requested I added more tests to the function which indeed also captured some nasty bugs. Further, I extended the documentation of the algorithm in https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:wenz:comissioning:nveto:hitlet_width#explanation_of_the_algorithm as well as in the code itself.

jorana (approved):
Thanks Daniel!

PR #386, opened 2021-01-24
very small simplification of pbar

jorana:
Very very small tweak to progress bar. Was working on something related and recognized it might also work here, seems like it does. Could have written an issue but would have been as much work as this small tweak. Daniel did all the actual work 😉 .

Reduce the `get_iter` complexity by allowing a disabled progress bar.

**Extra feature**
Because we slightly changed how the pbar is set up, it now also works for `st.make` (at least, I thought this did not work before) :
afbeelding

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -926,78 +926,48 @@ class Context:
                 allow_lazy=self.context_config['allow_lazy'],
                 max_messages=self.context_config['max_messages'],
                 timeout=self.context_config['timeout']).iter()
-        
+
+        # If no selection is specified we have to get the last end_time:
+        start_time = None
+        end_time = None
+
         # Defining time ranges for the progress bar:
         if time_range:
             # user specified a time selection
             start_time, end_time = time_range
-        else:
-            # If no selection is specified we have to get the last end_time:
-            start_time = 0
-            end_time = float('inf')
-
-        if progress_bar:
+        elif progress_bar:
+            # It is off unless we find some start and end times to work with
+            progress_bar = False
             for t in strax.to_str_tuple(targets):
                 try:
-                    # Sometimes some metadata might be missing e.g. during tests.
+                    # Sometimes some metadata might be missing e.g.
+                    # during tests. Just get one plugin, should be good enough.
                     chunks = self.get_meta(run_id, t)['chunks']
-                    start_time = max(start_time, chunks[0]['start'])
-                    end_time = min(end_time, chunks[-1]['end'])
+                    start_time = chunks[0]['start']
+                    end_time = chunks[-1]['end']
+                    progress_bar = True
+                    break
                 except (strax.DataNotAvailable, KeyError, IndexError):
-                    # IndexError caused for only empty chunks when data not
-                    # available.
-                    # Maybe at least one target had some metadata.
-                    start_time = max(start_time, 0)
-                    end_time = min(end_time, float('inf'))
-
-            # Define nice progressbar format:
-            bar_format = "{desc}: |{bar}| {percentage:.2f} % [{elapsed}<{remaining}],"\
-                         " {postfix[0]} {postfix[1][spc]:.2f} s/chunk,"\
-                         " #chunks processed: {postfix[1][n]}"
-            sec_per_chunk = np.nan  # Have not computed any chunk yet.
-            post_fix = ['Rate last Chunk:', {'spc': sec_per_chunk, 'n': 0}]
+                    pass
 
         try:
-            with contextlib.ExitStack() as stack:
-                if progress_bar and (start_time != 0 and end_time != float('inf')):
-                    # Get initial time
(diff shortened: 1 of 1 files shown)


strax/context.py
```diff
@@ -1014,6 +965,50 @@ def get_iter(self, run_id: str,
...
+        :param progress_bar: Bool weither or not to display the progress bar
+        :return: progress bar, t_start (run) and t_end (run)
+        """
+        try:
+            t_start, t_end, = self.estimate_run_start_and_end(run_id, targets)
+        except (AttributeError, KeyError, IndexError):
+            # During testing some thing remain a secret
+            t_start, t_end, = 0, float('inf')
+        if t_end == float('inf'):
+            progress_bar = False
```
WenzDaniel, line 981: I do no think that `self.estimate_run_start_and_end(run_id, targets)` can return `inf` can it? You could remove the if-statement.

strax/context.py
```diff
@@ -977,27 +944,11 @@ def get_iter(self, run_id: str,
...
-                        sec_per_chunk = pbar.last_print_t - last_time
-                        pbar.postfix[1]['spc'] = sec_per_chunk
-                        pbar.postfix[1]['n'] = n_chunks
-                        pbar.refresh()
-                        last_time = pbar.last_print_t
+                    self._update_progress_bar(
+                        pbar, t_start, t_end, n_chunks, result.end)
 
                     yield result
+            _p.close()
```
WenzDaniel, line 951: Needed? Should be close by the `with`-statement is not it?

strax/context.py
```diff
@@ -1014,6 +965,50 @@ def get_iter(self, run_id: str,
...
+                    bar_format=bar_format,
+                    leave=True,
+                    disable=not progress_bar)
+        return pbar, t_start, t_end
+
+    @staticmethod
+    def _update_progress_bar(pbar, t_start, t_end, n_chunks,  chunk_end):
+        """Do some tqdm voodoo to get the progress bar for st.get_iter"""
+        if t_end - t_start > 0:
+            pbar.n += (chunk_end - t_start) / (t_end - t_start)/n_chunks
```
WenzDaniel, line 1001: Is there a particular reason why you changed `pbar.n = (chunk_end - t_start) / (t_end - t_start)`  into `pbar.n += (chunk_end - t_start) / (t_end - t_start)/n_chunks`? For me both is fine, I am just curious.

strax/context.py
```diff
@@ -1014,6 +965,50 @@ def get_iter(self, run_id: str,
...
+                      "|{bar}| "  # Bar that is being filled
+                      "{percentage:.2f} % "  # Percentage
+                      "[{elapsed}<{remaining}]"  # Time estimate
+                      "{postfix}"  # Extra info
+                      )
+        pbar = tqdm(total=1,
+                    desc=f'Loading {targets}',
+                    bar_format=bar_format,
+                    leave=True,
+                    disable=not progress_bar)
```
WenzDaniel, line 994: Nice, is this kexword-option new? It definitely makes things simpler. I may have overlooked it.

WenzDaniel (approved):
Nice work, thanks. Now it is much cleaner. I did not notice or maybe forgot that it does not work for `st.make`, from what I can recall I thought we just decided to switch it off by default.

strax/context.py
jorana: Yeah so if it cannot find anything, that is where it defaults to:
https://github.com/jorana/strax/blob/eab44a363c2ee83eb5dbe5137a395e5eca2495b2/strax/context.py#L822

strax/context.py
jorana: It's the tqdm recommended way according to their documentation. Felt like it did not hurt to do so:
https://github.com/tqdm/tqdm#manual

strax/context.py
jorana: Uhm, now that I look at it, I don't think there is a particular reason. I unfixed things, then noticed I was again (uhg) using a pip installed strax version. Then I noticed I made a mess and had to refix it. So in the end, I think there is no reason.

strax/context.py
jorana: I'm not sure, but indeed it's what made the biggest difference

jorana:
> 
> 
> Nice work, thanks. Now it is much cleaner. I did not notice or maybe forgot that it does not work for `st.make`, from what I can recall I thought we just decided to switch it off by default.

The thing was the start and end time calculation. There was some overlap with `self.estimate_run_start_and_end(run_id, targets)`  but the progress bar piece of the code was looking at the `targets` only. If you look at the targets, and still have to make them, you cannot read the start and endtime on the basis of the metadata.

strax/context.py
jorana: We by the way can safely agree on saying that tqdm is a horribly confusingly documented package. There are so many incompatible statements on this package that it's sometimes complete voodoo.

WenzDaniel: Yes, voodoo is the correct word I guess :D

PR #392, opened 2021-02-04
Allow option to be overwritten to allow subclassing

jorana:
Not allowing options to be overwritten can lead to less clear code than the other way around. See: https://github.com/XENONnT/cutax/issues/3

Allow an option to be overwritten with the expert flag `overwrite=True`.

Even better! I have a pull request:
https://github.com/XENONnT/cutax/pull/4


strax/config.py
```diff
@@ -37,7 +37,7 @@ def wrapped(plugin_class):
             # Already have some options set, e.g. because of subclassing
             # where both child and parent have a takes_config decorator
             for opt in result.values():
-                if opt.name in plugin_class.takes_config:
+                if opt.name in plugin_class.takes_config and not opt.overwrite:
                     raise RuntimeError(
                         f"Attempt to specify option {opt.name} twice")
             plugin_class.takes_config = immutabledict({
```
WenzDaniel, line 43: Hej thanks for the changes, one question though. I do not clearly see the overwrite part I have to admit. To me it looks like both options are now in `plugin_class.takes_config`, but maybe I am wrong. Have you checked the lineage and that the option is propagated to the plugin correctly?

jorana: Hi @WenzDaniel , yes I checked, nevertheless not bad to just add a test: 5fc6f87

WenzDaniel (approved):
Wow that was quick thanks

PR #393, opened 2021-02-05
Seperate top and bottom waveforms

petergaemers:
For waveform analysis we need the waveforms of the top and bottom array seperatly. Since nT is so big there is a significant time difference between photon arrival times between the two arrays
In sum_waveform I add two buffer arrays for each array which are filled, downsampled and stored with the total waveform
For this the function needs to number of pmts in the top array (and some splitting functions aswell)
Use together with https://github.com/XENONnT/straxen/pull/359

Waveforms

Performance wise this doesn't change anything. On 3 chunks with my laptop the old time is:
```
CPU times: user 1min 32s, sys: 51.6 s, total: 2min 24s
Wall time: 2min 52s
```

With these changes:
```
CPU times: user 1min 43s, sys: 46.2 s, total: 2min 29s
Wall time: 2min 42s
```

Storage wise it's a different story. Per chunk disk space for peaklets is increased from 50mb to 100 mb
Ps I made some mistake with chunk.py


jorana:
Hi Peter, many thanks! I think a lot of people are eagerly looking forward to this.

~Three~Two questions I have already:

- ~NICE that you already speed checked it~
- What is the effect on storage?
- Related to the second; I liked your idea of having the total, and the top, then you can get the bottom as the difference, have you decided against it?

jorana:
Also (sorry for being the one to ask) I do think we should make this optional. For xenon100 like experiments, XAMS and alike, we don't really need it per se right?

petergaemers:
I accidentally deleted my data so I needed to remake it.
Storage wise is suprizingly bad. Peaklets per chunk is almost doubled. From ~50mb to 100mb per chunk

For XAMS we actually really need this since we have diffent sensors in the top and bottom array. Their waveform shape is not the same

strax/mailbox.py
```diff
@@ -203,7 +203,7 @@ def cleanup(self):
         for t in self._threads:
             t.join(timeout=self.timeout)
             if t.is_alive():
-                raise RuntimeError("Thread %s did not terminate!" % t.name)
+                raise RuntimeError("Thread %s did not terminate! %s" % (t.name,self.timeout))
```
WenzDaniel: Would you like to make me happy and change this also into a f-string? Also could you please add some additional info about self.timeout into the string?

strax/mailbox.py
```diff
@@ -360,7 +360,7 @@ def next_ready():
                     if not self._read_condition.wait_for(next_ready,
                                                          self.timeout):
                         raise MailboxReadTimeout(
-                            f"{self.name} did not get {next_number} in time.")
+                            f"{self.name} did not get {next_number} in time.{self.timeout}")
```
WenzDaniel: Also here add some info, please.

strax/processing/peak_building.py
```diff
@@ -104,8 +104,8 @@ def find_peaks(hits, adc_to_pe,
 
 
 @export
-@numba.jit(nopython=True, nogil=True, cache=True)
-def store_downsampled_waveform(p, wv_buffer):
+@numba.jit(nopython=True, nogil=True, cache=False)
```
WenzDaniel: Is there a reason why you changed the caching to False? Usually numba is quite good in bookkeeping when to recompile.

strax/processing/peak_building.py
```diff
@@ -125,19 +125,26 @@ def store_downsampled_waveform(p, wv_buffer):
     if downsample_factor > 1:
         # Compute peak length after downsampling.
         # Do not ceil: see docstring!
+        #Doing this is a loop is not allowed by numba
```
WenzDaniel, line 128: I guess you mean looping over the field names of the structured array right? Yes this is somewhat annoying...

strax/processing/peak_building.py
```diff
@@ -104,8 +104,8 @@ def find_peaks(hits, adc_to_pe,
 
 
 @export
-@numba.jit(nopython=True, nogil=True, cache=True)
-def store_downsampled_waveform(p, wv_buffer):
+@numba.jit(nopython=True, nogil=True, cache=False)
+def store_downsampled_waveform(p, wv_buffer, wv_top_buffer, wv_bottom_buffer):
```
WenzDaniel, line 108: Could, you please also update the doc string? Something like "same as ... but for ..." would be enough.

strax/processing/peak_building.py
```diff
@@ -157,10 +164,10 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
         return
     dt = records[0]['dt']
 
-    # Big buffer to hold even largest sum waveforms
-    # Need a little more even for downsampling..
```
WenzDaniel, line 161: Specific reason to remove the comment? If still true, could you please add it again?

strax/processing/peak_splitting.py
```diff
@@ -6,7 +6,7 @@
 
 
 @export
-def split_peaks(peaks, records, to_pe, algorithm='local_minimum',
+def split_peaks(peaks, records, to_pe, n_top_pmts, algorithm='local_minimum',
```
WenzDaniel, line 9: Also here, please update the doc string.

WenzDaniel (commented):
Hi Peter thanks a lot for your nice work. Beside a few small comments, I think we should discuss if we really need both top and bottom waveforms. What do you think @jorana?

WenzDaniel:
Okay I was too slow :D

strax/processing/peak_building.py
petergaemers: Yes, making 1 bigger buffer to hold all 3 does not work

petergaemers:
Daniel I think you were commenting at the same time as I was changing most. Regarding the mailbox things.. I don't know where they come from actually...

strax/processing/peak_building.py
petergaemers: Check

strax/processing/peak_splitting.py
petergaemers: Check

jorana (commented):
Hi Peter, before going into the details I think it would make sense to write it in such a way that it's optional. Also taking twice as much storage is something to consider carefully, therefore not storing 3 when we can do with 2 arrays is something we should do.

If we agree on this and the tests are passing, I can provide a more detailed review.

petergaemers:
The test fails because I need to update the test.

For optional storing.. Sure, I don't really care. Do note that then people still make it for themselves and it will be harder to track what data is where made by who.

jorana:
Of course but this is the same with our option to do local-minima splitting. The people that change this themselves are usually skilled enough not to store their test data in the wrong place.

petergaemers:
Oke so now it's top array only with a option in peaklet processing to store or not. The field is added to the dtype anyway, els numba crashes. If it's not filled it's compressed no increase in storage space required.

Test is also updated

WenzDaniel:
Hej @petergaemers  Travis is complaining about some check: 

```
    @settings(deadline=None)

    @given(several_fake_records,

           st.integers(min_value=0, max_value=100),

           st.integers(min_value=1, max_value=100)

           )

    def test_sum_waveform(records, peak_left, peak_length):

        # Make a single big peak to contain all the records

        n_top_pmts=256

        store_top_waveform=True

        n_ch = 100

>       peaks = np.zeros(1, strax.peak_dtype(n_ch, n_sum_wv_samples=200) + [(('data_top'), np.float32, 200)])

E       ValueError: field 'data_top' occurs more than once
```
Could you look into this? To me it looks like you updated the peaks data_type but forgot to remove the mentioned line in the test.

petergaemers:
Whoops, thanks Daniel

jorana (changes requested):
Hi Peter, thanks for the changes. 

I don't understand why we cannot make the top array optional though in the dtype. We also change the dtype of the raw_records by an argument. If we always keep the dtype having both top and full we are taking twice the ram which is not desirable.

jorana:
What is codefactor complaining about now 🤦 ? Sometimes it's more of a curse than a blessing especially when it starts complaining about stuff that you haven't worked on...

strax/processing/peak_building.py
```diff
@@ -125,19 +126,24 @@ def store_downsampled_waveform(p, wv_buffer):
...
     else:
         p['data'][:p['length']] = wv_buffer[:p['length']]
-
+        if store_top_waveform:
+            p['data_top'][:p['length']] = wv_buffer_top[:p['length']]
 
 @export
 @numba.jit(nopython=True, nogil=True, cache=True)
-def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
+def sum_waveform(peaks, records, adc_to_pe, n_pmts_top, store_top_waveform=False, select_peaks_indices=None):
```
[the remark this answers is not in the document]
petergaemers: Yeah thats possible. I don't see any profit in doing that. It will be less explicit

strax/processing/peak_building.py
```diff
@@ -157,10 +163,11 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
         return
     dt = records[0]['dt']
 
-    # Big buffer to hold even largest sum waveforms
-    # Need a little more even for downsampling..
+    # Big buffer to hold even largest sum waveforms		
+     # Need a little more even for downsampling..
     swv_buffer = np.zeros(peaks['length'].max() * 2, dtype=np.float32)
-
+    swv_buffer_top = np.zeros(peaks['length'].max() * 2, dtype=np.float32)
```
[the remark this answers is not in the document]
petergaemers: You'll need something to pass around, els it becomes a mess calling functions. I'll make it smaller

strax/processing/peak_building.py
```diff
@@ -157,10 +163,11 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
         return
     dt = records[0]['dt']
 
-    # Big buffer to hold even largest sum waveforms
-    # Need a little more even for downsampling..
+    # Big buffer to hold even largest sum waveforms
```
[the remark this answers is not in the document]
petergaemers: I don't understand what you're refering to

petergaemers:
The change in raw_records is the length of the array. In this case it's the existance of the field.

WenzDaniel:
But also the extension of a field should work. Something similar to: 

```
def peak_dtype(store_top):
    create_list_of_dtypes
    if  store_top:
        create_list_of_dtypes += [(top bla)]
```

And then in the plugin you can do in the infer_dtype step: 

```
dtype = peak_dtype(self.config[bla])
```
This should work,

terliuk:
Why cannot we simply store it for all data? i don't see why we need to over-complicate things so much for such a trivial thing?

WenzDaniel:
Because of the data-size which increases quite a lot. In this regard this change is not trivial at all.

terliuk:
can we quantify a lot? and "a lot" of which data type, all data, records etc?

WenzDaniel:
I do not remember the exact value but I think the overall memory and storage consumption was increased by about 25% - 30% for peaks.

jorana:
> bad

So storage wise it's most likely indeed 50% if it was 100% before Peter. Also memory is ~x 2 would be my guess. Also we need to fully reprocess all the data without the possibility to first validate it if we merge this to strax master. This a change we should coordinate carefully @terliuk . By making it optional we could also do this while already in strax' master.

terliuk:
basically, this means that we do not transfer records/raw records, but peaklets is the lowest available level, right? that's the only scenario where having an extra waveform would make such a huge difference in memory

WenzDaniel:
No, in peaklets/peaks we store the summed waveform for each peak. Because, you do not want to make peaks always from scratch since it is computationally more heavy then just recreating event_basics. This is independent of raw_records/records.

terliuk:
One of the aspects that we have to keep in mind with a switch of top/bottom waveforms - we will need to propagate that switch all the way up to peaks/peak_basics/etc. As one of the main goals is to have timing parameters (center time/rise time/width etc) separately for top and bottom arrays (next step, after waveforms are available).

WenzDaniel:
> propagate that switch all the way up to peaks/peak_basics/etc
I think there is a conceptional miss understanding. The summed waveform only lives in peaklets and peaks. It has nothing to do with the data-kinds below.

terliuk:
> I think there is a conceptional miss understanding. The summed waveform only lives in peaklets and peaks. It has nothing to do with the data-kinds below.

I totally agree, but if we want to have top/bottom timing parameters available at higher levels, other modules should know that they should calculate them only if the corresponding waveforms are stored.

strax/dtypes.py
```diff
@@ -186,6 +186,10 @@ def peak_dtype(n_channels=100, n_sum_wv_samples=200, n_widths=11):
           'n_hits'), np.int32),
         (('Waveform data in PE/sample (not PE/ns!)',
           'data'), np.float32, n_sum_wv_samples),
+        (('Waveform data from the top array in PE/sample (not PE/ns!)',
+          'data_top'), np.float32, n_sum_wv_samples),
+        (('Waveform data from the bottom array in PE/sample (not PE/ns!)',
+          'data_bottom'), np.float32, n_sum_wv_samples),
         (('Peak widths in range of central area fraction [ns]',
```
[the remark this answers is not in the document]
kdund: In the cases top/bottom are redundant, would the compression studies alleviate the memory size issue here?

JelleAalbers:
Maybe you could store the 'area fraction top waveform' in a uint8 array? That is, store a uint8 array of `round(255 * top/total)` for each sample. With that and the usual sum waveform, you can reconstruct the top waveform to ~0.5% accuracy at each sample (surely enough if you're just after timing info), but you need only 1/4th the extra storage.

terliuk:
That might be an option, but we have to keep in mind that in this case AFT can be negative, as we eventually have negative areas in waveforms, so we need at least signed int.

strax/processing/peak_building.py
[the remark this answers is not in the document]
jorana: It's not really explicit if the argument is only used if another kwarg is changed right?

strax/processing/peak_building.py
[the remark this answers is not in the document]
jorana: You somehow changed these two lines but didn't actually change them, you can see in https://github.com/AxFoundation/strax/pull/393/files.

JelleAalbers:
Yes, with negative areas, AFT could be any real number theoretically. You'd need a float to store that, but 1-byte floats are really imprecise (and not even supported by numpy).

However, if the negative areas are just due to noise, you should be able to clip AFT to [0,1] without losing any significant timing information. It's something to test though.

WenzDaniel:
When we add the summed waveform for the top array, we would not remove the area_per_channel-field. Hence, AFT  should be unaffected by Jelle's proposal. I think this could be a very good compromise, which has to be communicated well to avoid any miss-understandings.

terliuk:
Is it possible to make a quick test on the effect of this approach on timing parameters for top and bottom arrays? I think this is one of the most important characteristics. I would also try to store a sign bit/bool. Also, i am not sure whether keeping it capped at one is a good idea. There are some weird values of AFT, where top area goes to very unusual values. It doesn't happen often, but here is an example in my plots (I actually don't remember already what settings/data did I use here, as it was a while ago). 
Negative_top_AFT

terliuk:
Hi, can we go with the first option for now and make top/bottom waveforms optional, but stored as a separate ones? Given that total = top+bottom, we need only one.

PR #394, opened 2021-02-06
specify saver timeout at context level

jorana:
Allow the timeout for the saver to be specified at the context level.

When storing files [with high compression](https://github.com/XENONnT/straxen/pull/358), one might change the context to allow also slower savers. However at the moment this is not such a trivial task. Using the context_config, we can change this number to some different value without having to change the strax source.


jorana:
Might need this for https://github.com/XENONnT/straxen/pull/358

WenzDaniel (approved):
Thanks a lot. Looks fine.

PR #395, opened 2021-02-06
allow dynamic function creation for dtype copy

jorana:
Add a function to more easily numbafy the copy from dtype to dtype. Also see https://github.com/XENONnT/straxen/issues/353

Numba magic 🧙 

Rather than writing each time a cumbersome function that has to get the dtype write, why not code this. The annoying thing about numba is that it does not allow you to use strings to loop over. This is a workaround.

```python
r = np.ones(1000, dtype=strax.raw_record_dtype()
def test_raw_to_records(r):
    buffer = np.zeros(len(r), r.dtype)
    strax.copy_to_buffer(r, buffer, "_test_r_to_buffer")
    if len(r):
        assert np.all(buffer == r)
```

**Special thanks to**
Daniel for helping writing this up nicely and Jim Pivarski for showing such functionality in numba.

### DIFF
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -5,10 +5,14 @@ there is an existing numba cache. Clear __pycache__ and restart.
 TODO: file numba issue.
 """
 import numpy as np
+import typing as ty
+import numba
+
 
 __all__ = ('interval_dtype raw_record_dtype record_dtype hit_dtype peak_dtype '
            'DIGITAL_SUM_WAVEFORM_CHANNEL DEFAULT_RECORD_LENGTH '
-           'time_fields time_dt_fields hitlet_dtype hitlet_with_data_dtype').split()
+           'time_fields time_dt_fields hitlet_dtype hitlet_with_data_dtype '
+           'copy_to_buffer').split()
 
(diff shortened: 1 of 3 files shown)


strax/dtypes.py
```diff
@@ -201,3 +205,61 @@ def peak_dtype(n_channels=100, n_sum_wv_samples=200, n_widths=11):
...
+    if func_name not in globals():
+        # Create a numba function for us
+        _create_copy_function(buffer.dtype, field_names, func_name)
+
+    globals()[func_name](source, buffer)
+
+
+def _create_copy_function(res_dtype, field_names, func_name):
+    """Write out a numba-njitted function to copy data"""
+    # Cannot cache = True since we are creating the function dynamically
```
WenzDaniel, line 249: I think the default of cache is True in numba is not it?

WenzDaniel (approved):
Thanks a lot Joran for this awesome idea. It makes copying data much less tedious.

strax/dtypes.py
jorana: Not sure but setting it to cache = True did raise problems. I think it's because there is no function for numba to inspect

Issue #398, opened 2021-02-11
Round progresbar


jorana:
![afbeelding](https://user-images.githubusercontent.com/22295914/107674679-7d1f5900-6c97-11eb-9e74-aa47a31fee10.png)

As pointed out by @dpserena this is confusing. We should patch this to 100%, To do this, we need to fix the off by one error and clip between 0,1.

jorana:
fixed in  #399

PR #399, opened 2021-02-21
clip progress progressbar

jorana:
Fix issue #398 for progress bars that appear to be unfinished.

There can be some jitter at the end of a run that appears to say you are are 99.x % or 100.x % whereas you are simply done, and we should set to progress to exactly 100%.


WenzDaniel (approved):
Thanks

PR #402, opened 2021-03-02
Numbafy split by containment

WenzDaniel:
In this PR we numbafy `split_by_containment` and change its return value into a numba.typed.List instead of a normal Python list. This has the advantage that the result of `split_by_containment` can be used in numba njitted functions which enhances there speed. Further, this PR increases the performance of  `split_by_containment` by a factor of x2 to x4 depending on its inputs. This performance increase will directly manifest in a performance increase of LoopPlugins. 
If the numba.typed.List output is used instead of a LoopPlugin one can gain another factor x10 boost if needed. This enables maybe new usages of `split_by_containment`. 
In addition to the function changes I added small tests to ensure, that the required replacement functions behave as their numpy counterparts. 

A simple comparison notebook can be found [here](https://github.com/XENONnT/analysiscode/blob/master/StraxTests/NumbafiedSplitByContainment.ipynb).

### DIFF
--- a/strax/processing/general.py
+++ b/strax/processing/general.py
@@ -144,7 +144,12 @@ def split_by_containment(things, containers):
     Assumes everything is sorted, and containers are non-overlapping.
     """
     if not len(containers):
-        return []
+        # No containers so return empty numba.typed.List
+        empty_list = List()
+        # Small hack to define its type before returning it:
+        empty_list.append(np.zeros(0, dtype=things.dtype))
+        empty_list = empty_list[:0]
+        return empty_list
 
     return _split_by_containment(things, containers)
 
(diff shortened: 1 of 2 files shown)


jorana (approved):
Looks like absolutely great work Daniel.

Anything you want me to review specifically, I looked over it and it's clear you did your work with diligence.

I think I need to start paying more attention to the way you write the tests, it looks great, can you perhaps help me with even adding some comments there.

strax/processing/general.py
```diff
@@ -135,13 +137,20 @@ def _fc_in(a_starts, b_starts, a_ends, b_ends, result):
...
-    """Return list of thing-arrays contained in each container
+    """
+    Return list of thing-arrays contained in each container. Result is
+    returned as a numba.typed.List or list if containers are empty.
 
-    Assumes everything is sorted, and containers are nonoverlapping
+    Assumes everything is sorted, and containers are non-overlapping.
     """
     if not len(containers):
         return []
```
jorana, line 147: Should this be a typed empy list?

WenzDaniel: Yes, thanks for spotting. It took a small hack to define the list type.

WenzDaniel:
Thanks a lot. I think I addressed all your comments.

PR #403, opened 2021-03-03
Use int32 for peak dt, fix #397

### LINKED ISSUE #397: dt integer overflow
We should add some test that the dt field can be stored in the int16. We either need to change the int16->int32 for everything or need to come up with some other smart solution. Effectively we cannot have peaks > 7ms due to this.

jorana:
Fix issue #397 , this way, we could create much much longer waveforms to never run into such issues again.

Use a signed int for the dt field in peaks as they may well get very long

**What is needed on the straxen-side**
Full reprocessing of everything with `peak_interval_dtype` by incrementing the version of the affected plugins.

**Have you tested this on the known culprit?**
Not yet, let's first discuss if we want to go down this road

**Why not uint16?**
Could also do that. I actually think np.int32 might be even better so as to stay well away from the boundary, also, the biggest cost (i.e. reprocessing) stays the same for either option.

**Why not cap the WF to a max duration**
See #397


WenzDaniel:
Hmm and raw_records? It is also an interval dtype which would be affected.

darrylmasson:
I think there are pros and cons to both `uint16` and `int32`. `int32` is future-proof, because we will never have peaks with duration 2s, however this would a) invalidate all currently stored raw_records, and b) require a corresponding patch in redax. We've already changed the lowest-level dtype once (#237), and right now doesn't seem like the best time to do it again. `uint16` only buys us a factor of 2, but this doesn't invalidate raw_records because the digitizers can't store enough data to be affected by this, so we wouldn't have to up-convert any of the massive data.

darrylmasson:
Also, for all records-type data, this value is either 2 or 10, because there is no down-sampling at the readout level.

jorana:
@darrylmasson @WenzDaniel , you are both right, I was a bit too quick. I've changed it now only for peaks, which should only require a version bump of peaklets and a change of the Merged S2 dtype.

Reprocessing will happen sooner or later, I think better do it within next three weeks. Might want to do this when peak-classification is changed for example.

darrylmasson (approved):

WenzDaniel (approved):
No comments.

jorana:
Okay, then I will keep this PR in the fridge until the opportune moment presents itself

jorana:
For bookkeeping, see https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:commissioning:telecons:20210126#updates_on_daq

PR #405, opened 2021-03-08
Infer start/stop from any dtype

jorana:
We determine the start/stop of a run from the metadata. If that fails, we try getting it from the target currently requested. However, if we are making that target, this of course doesn't work. Therefore, it makes a lot more sense to check all the datatypes if they are stored and determine the start and stop based on the datatype-metadata.

This is helpful e.g. when raw-records is stored but we want to make peaks. (And don't have access to the runs-metadata).

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -815,8 +815,7 @@ class Context:
                     pass
         self.log.warning(
             "Could not estimate run start and end time from "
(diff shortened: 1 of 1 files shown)


WenzDaniel (approved):
Looks good, thanks.

PR #407, opened 2021-03-10
Allow re-compression at copy to frontend

jorana:
Allow re-compression when transferring from one frontend to the other.

If the optional argument for a compressor is specified, use that as the compression algorithm by changing the metadata.

https://github.com/XENONnT/analysiscode/blob/master/StraxTests/Copy_to_frontend_and_recompress.ipynb


strax/context.py
```diff
@@ -1404,14 +1404,19 @@ def _apply_function(self, data, targets):
             data = function(data, targets)
         return data
 
-    def copy_to_frontend(self, run_id, target,
-                         target_frontend_id=None, rechunk=False):
+    def copy_to_frontend(self,
+                         run_id: str,
+                         target: str,
+                         target_frontend_id: int = None,
+                         target_compressor: str = None,
```
darrylmasson, line 1411: Is it worth considering passing non-null but type-correct "empty" default args here? So an empty string for the target_compressor and something equivalent for target_frontend_id?

darrylmasson (approved):

strax/context.py
jorana: sure, let me change the typecasting to `typing.Optional[str]`

jorana: Alright, fixed in f8e0809 

See typing.Optional for the documentation: https://docs.python.org/3/library/typing.html#typing.Optional

PR #411, opened 2021-03-21
Patch concat hits

WenzDaniel:
Fixes two bugs which can cause a division by zero bug in hitlet_properties.

Added and refactored hitlet tests. I was not able to find a suitable test for the `yield offset` bug any suggestions are welcome.


tests/test_hitlet.py
```diff
@@ -257,9 +253,15 @@ def test_hitlet_properties(hits_n_data):
         h = hitlets[ind]
         h['data'][:h['length']] = d[:h['length']]
 
-    strax.hitlet_properties(hitlets)
+    # Step 3.: Add np.nan in data but outside of length:
+    for h in hitlets:
+        if h['length'] < len(h['data']):
+            h['data'][-1] = np.nan
+            # It is enough to test this for a single hitlet:
+            break
```
WenzDaniel, line 261: This test catches the highest_density_region bug

WenzDaniel:
We are doing to many tests :D The Travis error `The requested URL returned error: 429` is due to too many URL requests in a short time. That is why travis was not able to git clone strax,

jorana (approved):
Thanks Daniel

Issue #400, opened 2021-02-24
How specify option per run if not default


JoranAngevaare:
In straxen, we have a lot of [plugins with the default ](https://github.com/XENONnT/straxen/blob/b5433bdbd134bbf14a406844fe3089bc1ed163e3/straxen/plugins/event_processing.py#L293)per run for 1T. How can we overwrite those if we want to set the default to something else?
We cannot overwrite this nicely in the context config (e.g. the 1T), can we?

JoranAngevaare:
Fixed in #550

PR #413, opened 2021-03-28
Versioned validity intervals

jmosbacher:
This is a basic implementation of versioned interval of validity values meant to be used eg for config options that depend on time (or some other monotonic integer index) and need to be applied on a per time interval basis.
This will assist in generalizing the concept of "run_id" to a more general "interval index". Adding support for interval indexing will eliminate the need for the domain specific run_id field and code can be shared between chunking logic and run logic as simply indexing logic. When that is completed, chunking can be optimized to result in minimal time dependence of config parameters within a chunk.   

the VersionedValidityIntervals class tracks changes on a per interval basis, setting a value on a interval that has already been set creates a new version for the overlapping interval only. requesting the value at a point or list of points returns the latest version at every point by default but can be limited to a maximum version for repeating previous calculations after some values already changed.
 It currently uses an IntervalTree to manage intervals in memory but this has been abstracted as an IntervalEngine abstract class that can be easily implemented using mongodb or some other shared storage. 

```
class Config:
    some_config = VersionedValue(default=float("nan"))
    other_config = VersionedValue(default=np.arange(4))
config = Config()
now = int(time.time())
config.some_config[1:now] = 10 # intervals are open on the right so this will set the interval [1,now) to the value of 10.
config.other_config[now:now+3600] = np.array([3,2,1]) 

config.some_config[now] 
>>> nan
config.some_config[now-1]
>>>10
config.other_config[now]
>>>array([3,2,1]) 
config.other_config[now-1]
>>>array([1,2,3]) 
config.some_config[[0,now,now-10,now-20]] #passing iterables of indices also works
 >>>[nan,10,10,10]
config.some_config[now:now+10] = 15
config.some_config[[now-1,now, now+5]]
>>>[10,15,15] 
```
Visual example:
Screenshot_2021-04-01_19-02-10


jmosbacher:
This should help resolve issue #400

PR #414, opened 2021-03-30
Cleanup requirements for boto3

jorana:
# Cleanup boto3
Remove from requirements  as s3 backend is not used


jorana:
no idea why codefactor is having issues

PR #416, opened 2021-04-07
Check data availability for single run

jorana:
Imagine, you want to have a detailed report for a single run what is stored and what is not. This is chunky with current strax. Therefore, this PR adds a function that allows you to check what is stored for a single run with some simple regex matching.

### Example 1
```python
st.available_for_run('012275', 
                     include_targets = ('raw_records', 'peaklets'))
```
 target | is_stored
-- | --
peaklets | True
raw_records | False

### Example 2
```python
st.available_for_run('012275', 
                     include_targets = ('*record*'),
                     exclude_targets = ('*_mv', '*_nv', '*_he'))
```

target | is_stored
-- | --
records | False
raw_records | False
raw_records_aqmon | False

### Example 3

```python
st.available_for_run('012275', include_targets = ('*peak*')
```` 
target | is_stored
 -- | --
peaklet_classification | True
peaklet_classification_he | False
peaklets | True
peaklets_he | False
online_peak_monitor | False
peak_basics | True
peak_proximity | True
peak_positions_cnn | True
peak_positions_mlp | True
peak_positions_gcn | True
peak_basics_he | False

### Example 4
```python
st.available_for_run('012275')
```
target | is_stored
-- | --
corrected_areas | True
energy_estimates | False
event_basics | True
event_posrec_many | True
event_info | True
event_positions | True
events | True
distinct_channels | True
raw_records_coin_nv | False
lone_raw_records_nv | False
lone_raw_record_statistics_nv | False
records_mv | False
records_nv | False
hitlets_mv | False
hitlets_nv | False
event_positions_nv | False
events_nv | False
aqmon_hits | False
veto_intervals | False
records | False
veto_regions | True
pulse_counts | True
records_he | False
pulse_counts_he | False
merged_s2s | True
merged_s2s_he | False
peaklet_classification | True
peaklet_classification_he | False
peaklets | True
lone_hits | True
peaklets_he | False
online_event_monitor | False
online_peak_monitor | False
peak_basics | True
peak_proximity | True
peak_positions_cnn | True
peak_positions_mlp | True
peak_positions_gcn | True
peak_basics_he | False
raw_records | False
raw_records_he | False
raw_records_aqmon | False
raw_records_nv | False
raw_records_aqmon_nv | False
raw_records_aux_mv | False
raw_records_mv | False
led_calibration | False

### DIFF
--- a/strax/run_selection.py
+++ b/strax/run_selection.py
@@ -299,7 +299,8 @@ def available_for_run(self: strax.Context,
     :param exclude_targets: targets to exclude e.g. raw_records,
         raw_records* or *_nv. If multiple targets (e.g. a list) is
         provided, the target should match none of the arguments!
-    :param pattern_type: either 'fnmatch' or 're'
+    :param pattern_type: either 'fnmatch' (Unix filename pattern
+        matching) or 're' (Regular expression operations).
     :return: Table of available data per target
     """
     if not isinstance(run_id, str):
--- a/strax/run_selection.py
+++ b/strax/run_selection.py
@@ -2,7 +2,7 @@
 import fnmatch
 import re
 import typing as ty
-
+from collections import defaultdict
 import numpy as np
 import pandas as pd
 from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


strax/run_selection.py
```diff
@@ -283,6 +283,66 @@ def define_run(self: strax.Context,
                            " run definition")
 
 
+@strax.Context.add_method
+def available_for_run(self: strax.Context,
+                      run_id: str,
+                      include_targets: ty.Union[None, list, tuple, str] = None,
+                      exclude_targets: ty.Union[None, list, tuple, str] = None,
+                      pattern_type: str = 'fnmatch') -> pd.DataFrame:
```
WenzDaniel, line 291: I also have to start using this, I am not at all used to this yet.

strax/run_selection.py
```diff
@@ -283,6 +283,66 @@ def define_run(self: strax.Context,
...
+    For a given single run, check all the targets if they are stored.
+        Excludes the target if never stored anyway.
+    :param run_id: requested run
+    :param include_targets: targets to include e.g. raw_records,
+        raw_records* or *_nv. If multiple targets (e.g. a list) is
+        provided, the target should match any of the arguments!
+    :param exclude_targets: targets to exclude e.g. raw_records,
+        raw_records* or *_nv. If multiple targets (e.g. a list) is
+        provided, the target should match none of the arguments!
+    :param pattern_type: either 'fnmatch' or 're'
```
WenzDaniel, line 302: What does `fnmatch` or `re` stand for?

strax/run_selection.py
```diff
@@ -283,6 +283,66 @@ def define_run(self: strax.Context,
...
+    """
+    For a given single run, check all the targets if they are stored.
+        Excludes the target if never stored anyway.
+    :param run_id: requested run
+    :param include_targets: targets to include e.g. raw_records,
+        raw_records* or *_nv. If multiple targets (e.g. a list) is
+        provided, the target should match any of the arguments!
+    :param exclude_targets: targets to exclude e.g. raw_records,
+        raw_records* or *_nv. If multiple targets (e.g. a list) is
+        provided, the target should match none of the arguments!
```
WenzDaniel, line 301: Maybe add a command that exclude > include. If I understand your function correct, `includ_targets='raw_records*', exclude_tragets='*records*'` would return nothing.

WenzDaniel (commented):
Only a minor comment regarding the documentation, but the rest is fine. Thanks a lot.

strax/run_selection.py
jorana: It's super useful, especially since your PyCharm also uses this type checking

strax/run_selection.py
jorana: Proper python constructs 😉 :
- https://docs.python.org/3/library/fnmatch.html
- https://docs.python.org/3/library/re.html

strax/run_selection.py
jorana: Exactly, I thought about that too, however, if someone is both excluding and including the same targets, I think they should think a bit harder. It is a bit hard to guess what a given user would have meant in this case, as such, I figured it was better to be unhelpful rather than trying to guess what a user might have meant.

strax/run_selection.py
jorana: added a sentence or two

WenzDaniel (approved):
Thanks a lot, it is super handy. Thought, I approved already, sorry.

PR #420, opened 2021-04-09
add failing dt test - and solve it with max peak duration

jorana:
Add a test that will push us into #397.

Make a big bunch of records that should be sufficient to break strax with integer overflows. To solve the issue, we should solve this test.

This is the example

### DIFF
--- a/tests/test_peak_processing.py
+++ b/tests/test_peak_processing.py
@@ -132,11 +132,9 @@ def test_peak_overflow(records,
 
     # Set this here, no need to test left and right independently
     left_extension = 0
-    p = np.zeros(0, dtype=strax.peak_dtype())
-    magic_overflow_time = np.iinfo(p.dtype['dt']).max * p.dtype['data'].shape[0]
-    del p
+    peak_dtype = np.zeros(0, strax.peak_dtype()).dtype
+    magic_overflow_time = np.iinfo(peak_dtype['dt']).max * peak_dtype['data'].shape[0]
 
-    # Make a single big peak to contain all the records
     def retrun_1(x):
         """
         Return 1 for all of the input that can be used as a parameter
@@ -154,18 +152,19 @@ def test_peak_overflow(records,
         return
 
     # Copy the pulse train of the records. We are going to copy the same
-    # set of records many times now
-    r_buffer = []
-    t_max = strax.endtime(r)
-    max_record_repetition_factor = 1_000_000
-    for i in range(max_record_repetition_factor):
-        r_copy = r.copy()
-        r_copy['time'] = r_copy['time'] + t_max * i * gap_factor
-        r_buffer.append(r_copy)
-        if r_copy['time'][-1] - r['time'][0] > 1.5 * magic_overflow_time:
-            # No need to go over and beyond. We should get int-overflow
-            break
-    r = np.concatenate(r_buffer)
+    # set of records many times now.
+    t_max = strax.endtime(r).max()
+    n_repeat = int(1.5 * magic_overflow_time + t_max * gap_factor) // int(t_max * gap_factor) + 1
+    time_offset = np.linspace(0,
+                              1.5 * magic_overflow_time + t_max * gap_factor,
+                              n_repeat,
+                              dtype=np.int64)
+    r_buffer = np.tile(r, n_repeat // len(r) + 1)[:len(time_offset)]
+    assert len(r_buffer) == len(time_offset)
+    r_buffer['time'] = r_buffer['time'] + time_offset
+    assert strax.endtime(r_buffer[-1]) - r_buffer['time'].min() > magic_overflow_time
+    r = r_buffer.copy()
+    del r_buffer
 
     # Do peak finding!
     hits = strax.find_hits(r, min_amplitude=0)
--- a/tests/test_peak_processing.py
+++ b/tests/test_peak_processing.py
@@ -5,87 +5,87 @@ import hypothesis.strategies as st
 
 import strax
 
-#
-# @given(fake_hits,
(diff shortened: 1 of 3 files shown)


jorana:
We must have been really desperate of tests:
afbeelding

tests/test_peak_processing.py
```diff
@@ -87,3 +86,161 @@ def test_sum_waveform(records, peak_left, peak_length):
...
+    # Make a single big peak to contain all the records
+    def retrun_1(x):
+        """
+        Return 1 for all of the input that can be used as a parameter
+            for the splitting in natural breaks
+        :param x: any type of array
+        :return: ones * len(array)
+        """
+        ret = np.ones(len(x))
+        return ret
```
WenzDaniel, line 148: Can we use a lambda function instead? This looks a bit odd.

tests/test_peak_processing.py
```diff
@@ -87,3 +86,161 @@ def test_sum_waveform(records, peak_left, peak_length):
...
+    t_max = strax.endtime(r)
+    max_record_repetition_factor = 1_000_000
+    for i in range(max_record_repetition_factor):
+        r_copy = r.copy()
+        r_copy['time'] = r_copy['time'] + t_max * i * gap_factor
+        r_buffer.append(r_copy)
+        if r_copy['time'][-1] - r['time'][0] > 1.5 * magic_overflow_time:
+            # No need to go over and beyond. We should get int-overflow
+            break
+    r = np.concatenate(r_buffer)
```
WenzDaniel, line 168: This is a bit confusing to me-. Are not you just creating equidistant intervals like: 
```
time = np.arange(time, 1,5* magic_overflow_time + t_max*gap_factor, t_max*gap_factor)
```
which would be much simpler?

tests/test_peak_processing.py
```diff
@@ -87,3 +86,161 @@ def test_sum_waveform(records, peak_left, peak_length):
...
+            split_low=True,
+            filter_wing_width=70,
+            min_area=0,
+            do_iterations=2)
+    except AssertionError as e:
+        if not left_extension + max_duration + right_extension < magic_overflow_time:
+            # Ending up here is the ultimate goal of the tests. This
+            # means we are hitting github.com/AxFoundation/strax/issues/397
+            print(f'Great, the test worked, we are getting the assertion '
+                  f'statement for the int overflow')
```
WenzDaniel, line 232: Would be very strange if we ever trigger this.

WenzDaniel (approved):
Just some minor comments, in general it is fine.

tests/test_peak_processing.py
```diff
@@ -87,3 +86,161 @@ def test_sum_waveform(records, peak_left, peak_length):
...
+    :param right_extension: option for strax.find_peaks
+    :param gap_threshold: option for strax.find_peaks
+    :return: None
+    """
+
+    # Set this here, no need to test left and right independently
+    left_extension = 0
+    p = np.zeros(0, dtype=strax.peak_dtype())
+    magic_overflow_time = np.iinfo(p.dtype['dt']).max * p.dtype['data'].shape[0]
+    del p
```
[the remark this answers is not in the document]
jorana: because I was lazy and used `p` for peaks again below

WenzDaniel:
One more thing I noticed that your changes are also present in: https://github.com/AxFoundation/strax/pull/403/files. We have to be careful with the commit history I think.

tests/test_peak_processing.py
jorana: indeed, this is exactly what it is doing but I also want to have some records data, if it were just the time fields I was interested in, you would be right.

tests/test_peak_processing.py
jorana: Yeah, this was the design of the tests as I first wrote the test and then implemented the solution.

jorana:
>One more thing I noticed that your changes are also present in: https://github.com/AxFoundation/strax/pull/403/files. We have to be careful with the commit history I think.

This is true, it's almost exactly the same so should be relatively easy to resolve the merge-conflicts.

tests/test_peak_processing.py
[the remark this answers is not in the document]
jorana: let me just change it

tests/test_peak_processing.py
WenzDaniel: But the data stays the same is not it. It is something like result['data'][:] = r['data']

tests/test_peak_processing.py
WenzDaniel: But the data stays the same is not it. It is something like result['data'][:] = r['data']

tests/test_peak_processing.py
jorana: lambda is very non pep8 right? What would your main motivation be for making this a lambda?

WenzDaniel: Hmm, I to my understand something like: 
```
function(a, b, c labda x: do_something_simple)
```
would be the default example for a lambda function. Like used in `filter`, `map` etc.   It is the only places where I would use them.

tests/test_peak_processing.py
jorana: okay, yeah it is a bit convoluted, I've changed to `np.tile` to do almost exactly the same, thanks for the suggestion

tests/test_peak_processing.py
jorana: @WenzDaniel , I just stumbled across the pep8 guidelines, and I must say, I would agree with their reasoning:
https://www.python.org/dev/peps/pep-0008/

PR #421, opened 2021-04-09
Fix dt integer overflow with max-peak length in peakfinding

jorana:
In #420, we showcase a test that fails under the current strax version. In this PR we add the solution to the failing tests. This is an alternative to #397. We might not really actually need to set the dt to an int16 if we simply not make pulses this long.

In the rules we take into account for merging, we also consider a max time which we at the moment set to 6.7 us. This should prevent int overflows


jorana:
github bug

PR #422, opened 2021-04-10
update apply function to data & test

jorana:
Move the `apply_data_function` logic from `st.get_array` to `st.get_iter` to avoid any malicious/accidental usage of `st.get_iter` thereby bypassing `apply_data_function`.

Additionally, we fix an issue we had for a while in straxen that this option was incompatible with selection strings:
https://github.com/XENONnT/straxen/blob/5c945e141bee0d5fa976a6e1a78516a0a172305a/straxen/common.py#L491-L493
Therefore, move the application of this function prior to the application of a selection string.

apply_data_function is applied just prior to returning the data to the user.

See the tests as simple examples of use cases.

### DIFF
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -73,7 +73,7 @@ def test_apply_drop_data():
     class Drop:
         """Small class to keep track of the number of dropped rows"""
 
-        dropped = 0
+        kept = []
 
         def drop(self, data, r, t):
             """Drop a random portion of the data"""
@@ -82,7 +82,7 @@ def test_apply_drop_data():
             keep = np.random.randint(0, 2, len(data)).astype(np.bool_)
 
             # Keep in mind that we do this on a per chunk basis!
-            self.kept = keep
+            self.kept += [keep]
             res = data.copy()[keep]
             return res
 
@@ -91,8 +91,7 @@ def test_apply_drop_data():
     r, r_changed = _apply_function_to_data(dropper.drop)
 
     # The number of records should e
-    assert np.all(r[dropper.kept] == r_changed)
-
+    assert np.all(r[np.concatenate(dropper.kept)] == r_changed)
 
 
 def test_accumulate():
--- a/tests/test_context.py
+++ b/tests/test_context.py
@@ -32,7 +32,8 @@ def _apply_function_to_data(function
 def test_apply_pass_to_data():
     """
     What happens if we apply a function that does nothing (well,
-        nothing hopefully)
+    nothing hopefully)
+
     :return: None
     """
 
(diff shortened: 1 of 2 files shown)


tests/test_context.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_apply_drop_data():
+    """
+    What if we drop random portions of the data, do we get the right results?
+    :return: None
+    """
+
+    class Drop:
+        """Small class to keep track of the number of dropped fields"""
```
WenzDaniel, line 73:
```suggestion
        """Small class to keep track of the number of dropped rows"""

```
First I associated fields with the field names of the structured array.

tests/test_context.py
```diff
@@ -0,0 +1,108 @@
...
+        dropped = 0
+
+        def drop(self, data, r, t):
+            """Drop a random portion of the data"""
+            # I was too lazy to write a strategy to get the right number
+            # of random drops based on the input records.
+            keep = np.random.randint(0, 2, len(data)).astype(np.bool_)
+
+            # Keep in mind that we do this on a per chunk basis!
+            self.dropped += len(data) - np.sum(keep)
```
WenzDaniel, line 83:
```suggestion
            self.kept = keep
```

tests/test_context.py
```diff
@@ -0,0 +1,108 @@
...
+            res = data.copy()[keep]
+            return res
+
+    # Init the bookkeeping class
+    dropper = Drop()
+    r, r_changed = _apply_function_to_data(dropper.drop)
+
+    # The number of records should e
+    assert (dropper.dropped == len(r) - len(r_changed))
```
WenzDaniel, line 92:
```suggestion
    assert np.all(r[dropper.kept] == r_changed)

```

tests/test_context.py
WenzDaniel: See comment below, is a bit more thorough.

tests/test_context.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_accumulate():
+    """
+    Test the st.accumulate function. Should add the results and
+        accumulate per chunk. Lets add channels and verify the results
```
WenzDaniel, line 98: Sphinx does not need an indentation here: https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

WenzDaniel (approved):
Thanks a lot. Looks fine, just some smaller optional changes.

tests/test_context.py
jorana: Thanks, that is true

jorana:
@WenzDaniel , thanks!

PR #423, opened 2021-04-10
context testing functions + copy_to_frontend documented

jorana:
Add some tests to functions in the context previously untested, most notably the copy to frontend function.


WenzDaniel (approved):
Looks fine, thanks.

Issue #221, opened 2019-11-22
Containment options for LoopPlugin, documentation


JelleAalbers:
Prompted by a question from @lucasdevries:

Currently LoopPlugin only returns things fully contained in the larger things. For example, if you loop over peaks in an event, you only get the peaks that are completely in the event, not peaks that are partially in it. Or, if you loop over records in a peak, you only get the records that are completely in the peak, not those that contribute to it but have some part outside it.

We should also allow users to loop over all small things that touch a bigger thing. For get_data and get_array this is already implemented using the `time_selection='touching'` option.

Also, LoopPlugin should have documentation :-)

jorana:
Started with the documentation in:
https://github.com/AxFoundation/strax/pull/274 commit 9182f1b 

For resolving this issue I'll prepare another PR.

jorana:
Should be easy by changing this line:
https://github.com/AxFoundation/strax/blob/164a25239ef08b3dad5ddd958d1d404b776a163a/strax/plugin.py#L665

jorana:
Fixed in #424

PR #424, opened 2021-04-10
Loopplugin touching windows + plugin documentation

jorana:
As explained in https://github.com/AxFoundation/strax/issues/221, having a touching windows option might be nice to add in the `LoopPlugins`. This PR adds that.

**Experimental**
Note: as Daniel correctly pointed out (https://github.com/AxFoundation/strax/pull/424#pullrequestreview-633305496), this may lead to ambiguity if for example one peak is touching the window of two events, this would mean that if one is looping over events, that one peak is included twice. Depending on the application, this may or may not be an issue.

For now, we added a warning to alert the user of this (potential) issue.

It's only a time-selection that allows to slightly extend the time range. 

Please see the tests for an example.

**Documentation**
Has been updated, additionally, I wrote out some examples that might be useful and work without any `straxen` installation.

afbeelding


WenzDaniel (changes requested):
Ah here we have to be careful. At least we should discuss it. For touching windows you can become ambiguities and use the same data twice which is not desirable. E.g. if you have a peak which touches two adjacent events it will be associated with both. See as an example: https://github.com/XENONnT/straxen/blob/a49c51ac1fda5be8e35f1ff19b4637d7a89700da/straxen/plugins/veto_events.py#L173-L181 in which I had to deal with such cases. 

While I think such a plugin could be in principle useful, I am not sure if it will be easy to find a general how to solve such an ambiguity.

jorana:
@WenzDaniel , thanks, valid point. While I agree that this a potential issue, it is not per se a bad thing, people should be careful with what they do with that data indeed. If you want to have e.g. a more meta-number like how many peaks are at least partially overlapping with an event, having peaks taken into account twice might not be a huge problem.

I don't have some concrete examples where this would be used, I think there is no a-priori problem if we allow this to be supported.

How about the following:
- I add a comment in the plugin;
- I for now raise a warning in the plugin when this option is ever used (it can be disabled then if needed)

WenzDaniel:
Yes, I agree we should add both a comment and a warning.

jorana:
@WenzDaniel , thanks again done, also added a reference to this PR as well as updated the description of this PR.

WenzDaniel (approved):
Thanks

jorana:
@WenzDaniel thanks for the review and the good comment 😄

PR #427, opened 2021-04-16
Allow Py39 in travis tests

jorana:
Allow python 3.9


WenzDaniel (approved):
We are getting closer to python 4.0, I hope it wont be a transition like for python 2

PR #430, opened 2021-04-26
Refactor concat and get data

WenzDaniel:
In this PR Tianyu and I (mostly Tianyu) refactored the get_hitlet_data function. Tianyus proposal makes the function much more flexible as we search no longer for the hitlet data via the `record_i` field but using time intervals instead. This change allows a much more general usage of the function. For this reason I added a "header" part which checks for different possibilities in calling this function. The corresponding tests have been added too.

Beside get_hitlet_data I refactored a bit concat_overlaping_hits as it was also part of the proposal. However, I did not manage to modify the function such, that it mimics the full behavior of the old method. But since the new method would lead to a performance drop of about 30% I decided to keep the old one. 

A more thorough discussion can be found in this [note](https://github.com/XENONnT/analysiscode/blob/master/nveto/RefactorGetHitletData.ipynb).

Open points:
- [x] Updated plugins, set correct channel offset for mv and nv.
- [x] Updated hitlet splitting
- [x] Add some test for hitlet splitting.
- [x] Test process some data

### DIFF
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -152,20 +152,18 @@ def hitlet_dtype():
     return dtype
 
 
-def hitlet_with_data_dtype(n_samples):
+def hitlet_with_data_dtype(n_samples=2):
     """
     Hitlet dtype with data field. Required within the plugins to compute
     hitlet properties. 
     
     :param n_samples: Buffer length of the data field. Make sure it can
         hold the longest hitlet.
-    :param n_widths: Number of area deciles width.
-
-    Note:
-        The data buffer will be at least 2 samples long
     """
+    if n_samples < 2:
+        raise ValueError('n_samples must be at least 2!')
+
     dtype = hitlet_dtype()
-    n_samples = max(n_samples, 2)
     additional_fields = [(('Hitlet data in PE/sample with ZLE (only the first length samples are filled)', 'data'),
                            np.float32, n_samples),
                          ]
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -168,8 +168,6 @@ def hitlet_with_data_dtype(n_samples):
     n_samples = max(n_samples, 2)
     additional_fields = [(('Hitlet data in PE/sample with ZLE (only the first length samples are filled)', 'data'),
                            np.float32, n_samples),
-                         (('Fragment number in the pulse',
-                           'record_i'), np.int32),
                          ]
 
     return dtype + additional_fields
--- a/strax/processing/hitlets.py
+++ b/strax/processing/hitlets.py
@@ -149,12 +149,19 @@ def get_hitlets_data(hitlets, records, to_pe):
     :param hitlets: Hitlets found in a chunk of records.
     :param records: Records of the chunk.
     :param to_pe: Array with area conversion factors from adc/sample to
-        pe/sample.
-    :param channel_offset: Channel offset of the nveto with respect to
-        zero. Required to query the correct gain values.
+        pe/sample. Please make sure that to_pe has the correct shape.
+        The array index should match the channel number.
     :returns: Hitlets including data stored in the "data" field
         (if it did not exists before it will be added.)
     """
+    # Numba will not raise any exceptions if to_pe is too short, leading
+    # to strange bugs. This check is somewhat not complete, but should
+    # be sufficient in most cases.
+    to_pe_has_wrong_shape = len(to_pe) < hitlets['channel'].max()
+    if to_pe_has_wrong_shape:
+        raise ValueError('"to_pe" has a wrong shape. Array index must'
+                         ' match channel numbers.')
+
     hitelts_is_single_row = isinstance(hitlets, np.void)
     if hitelts_is_single_row:
         # A structured array becomes void type if a single row is called,
--- a/strax/processing/hitlets.py
+++ b/strax/processing/hitlets.py
@@ -136,7 +136,7 @@ def refresh_hit_to_hitlets(hits, hitlets):
 
 
 @export
-def get_hitlets_data(hitlets, records, to_pe):
+def get_hitlets_data(hitlets, records, to_pe, min_hitlet_sample=100):
     """
     Function which searches for every hitlet in a given chunk the 
     corresponding records data. Additionally compute the total area of
@@ -147,6 +147,8 @@ def get_hitlets_data(hitlets, records, to_pe):
     :param to_pe: Array with area conversion factors from adc/sample to
         pe/sample. Please make sure that to_pe has the correct shape.
         The array index should match the channel number.
+    :param min_hitlet_sample: minimal length of the hitlet data field.
+        prevents numba compiling from running into race conditions.
     :returns: Hitlets including data stored in the "data" field
         (if it did not exists before it will be added.)
     """
@@ -176,7 +178,7 @@ def get_hitlets_data(hitlets, records, to_pe):
 
         hitlets_with_data_field = hitlets
     else:
-        n_samples = max(100, hitlets['length'].max())
+        n_samples = max(min_hitlet_sample, hitlets['length'].max())
         hitlets_with_data_field = np.zeros(len(hitlets), strax.hitlet_with_data_dtype(n_samples))
         strax.copy_to_buffer(hitlets,
                              hitlets_with_data_field,
--- a/strax/processing/hitlets.py
+++ b/strax/processing/hitlets.py
@@ -140,7 +140,7 @@ def refresh_hit_to_hitlets(hits, hitlets):
 
 
 @export
-def get_hitlets_data(hitlets, records, to_pe):
+def get_hitlets_data(hitlets, records, to_pe, channel_offset=2000):
     """
     Function which searches for every hitlet in a given chunk the 
     corresponding records data. Additionally compute the total area of
@@ -149,7 +149,9 @@ def get_hitlets_data(hitlets, records, to_pe):
     :param hitlets: Hitlets found in a chunk of records.
     :param records: Records of the chunk.
     :param to_pe: Array with area conversion factors from adc/sample to
-        pe/sample
+        pe/sample.
+    :param channel_offset: Channel offset of the nveto with respect to
+        zero. Required to query the correct gain values.
     :returns: Hitlets including data stored in the "data" field
         (if it did not exists before it will be added.)
     """
@@ -178,12 +180,12 @@ def get_hitlets_data(hitlets, records, to_pe):
                              hitlets_with_data_field,
                              '_copy_hitlets_to_hitlets_width_data')
 
-    _get_hitlets_data(hitlets_with_data_field, records, to_pe)
+    _get_hitlets_data(hitlets_with_data_field, records, to_pe, channel_offset)
     return hitlets_with_data_field
 
 
 @numba.jit(nopython=True, nogil=True, cache=True)
-def _get_hitlets_data(hitlets, records, to_pe):
+def _get_hitlets_data(hitlets, records, to_pe, channel_offset):
     rranges = _touching_windows(records['time'],
                                 strax.endtime(records),
                                 hitlets['time'],
@@ -216,7 +218,7 @@ def _get_hitlets_data(hitlets, records, to_pe):
         h['time'] += int(recorded_samples_offset * h['dt'])
         h['length'] = n_recorded_samples
 
-        h['data'][:] = h['data'][:] * to_pe[h['channel']]
+        h['data'][:] = h['data'][:] * to_pe[h['channel']-channel_offset]
         h['area'] = np.sum(h['data'])
 
 # ----------------------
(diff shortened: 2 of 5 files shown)


strax/processing/hitlets.py
```diff
@@ -133,165 +136,90 @@ def refresh_hit_to_hitlets(hits, hitlets):
...
+        # A structured array becomes void type if a single row is called,
+        # e.g. hitlets[0] which does not work in numba while, hitlets[:1]
+        # does. So we have to convert the row into the correct format first.
+        hitlets = np.array([hitlets])
+
+    data_field_in_hitlets = 'data' in hitlets.dtype.names
+    if data_field_in_hitlets:
+        data_is_not_empty = np.any(hitlets['data'] != 0)
+        if data_is_not_empty:
+            raise ValueError('The data field of hitlets must be empty!')
```
zhut19, line 169: Is this necessary? It is almost impossible someone would accidentally call this twice.

strax/processing/hitlets.py
```diff
@@ -133,165 +136,90 @@ def refresh_hit_to_hitlets(hits, hitlets):
...
+                h['length'])
+
+            is_first_record = ind == 0
+            if is_first_record:
+                recorded_samples_offset = h_start
+            h_start -= recorded_samples_offset
+            h_end -= recorded_samples_offset
+
+            h['data'][h_start: h_end] += r['data'][r_start: r_end] + r['baseline'] % 1
+            n_recorded_samples += r_end - r_start
```
zhut19, line 215: Can you give this another check? This feels wrong to me since the overlapping index is already enforcing the length in hitlet is respected during data fetching.

zhut19: I mean the length field doesn't need to be updated. At least not any case I can think of.

strax/processing/hitlets.py
WenzDaniel: Maybe you are right, I happened me during test. On the other hand this check does not cost much and adds another layer of security so I am more inclined to keep it.

strax/processing/hitlets.py
WenzDaniel: You are right, but let me explain as the intention is different. In a few cases we find hits close to the edge of our recorded pulse such that the left or right hit extension reaches beyond the recorded data. As the data is not defined in those region be definition we (nveto subgroup) decided to chop the hitlets in those cases. This only affects less then 1% of our data.

strax/processing/hitlets.py
zhut19: Okay, it's up to you. I was thinking about tpc data, where after saturation correction, one would like to update the peaklet waveform. But that won't happen here, so it doesn't matter.

strax/processing/hitlets.py
```diff
@@ -620,7 +502,12 @@ def highest_density_region_width(data,
     """
     res = np.zeros((len(fractions_desired), 2), dtype=np.float32)
     data = np.maximum(data, 0)
-    inter, amps = strax.highest_density_region(data, fractions_desired, _buffer_size=_buffer_size)
+    
+    if np.all(data == 0):
+        res[:] = np.nan
+    else:
+        inter, amps = strax.highest_density_region(data,
+                                                   fractions_desired, _buffer_size=_buffer_size)
```
WenzDaniel, line 510: Sorry, I know one should not add more than one feature into a PR, but I am currently optimizing some splitting parameters and I run into this issue. `strax.highest_density_region` behaves as expected and raises an ValueError. However, it is not desired to stop the entire processing because of that. Hence, the function returns now np.nan in case the HDR is not defined.

tests/test_hitlet.py
```diff
@@ -205,6 +254,12 @@ def test_highest_density_region_width():
     truth_dict = {0.5: [[0, 1]], 0.8: [-0.3, 1.3]}
     _test_highest_density_region_width(np.array([1, 0]), truth_dict)
 
+    # Check that negative data does not raise:
+    res = strax.processing.hitlets.highest_density_region_width(np.array([0, -1, -2]),
+                                                          np.array([0.5]),
+                                                          fractionl_edges=True)
+    assert np.all(np.isnan(res)), 'For empty data HDR is not defined, should return np.nan!'
+
```
WenzDaniel, line 262: ... and added the corresponding test.

tests/test_hitlet.py
```diff
@@ -311,22 +366,25 @@ def test_hitlet_properties(hits_n_data):
...
+    odd_hitlets[2]['length'] = 3
+    odd_hitlets[3]['data'][:3] = [-1, -2, 0]
+    odd_hitlets[3]['length'] = 3
+
+    for oh in odd_hitlets:
+        res = strax.get_fwxm(oh)
+        mes = (f'get_fxhm returned {res} for {oh["data"][:oh["length"]]}!'
+               'However, the FWHM is not defined and the return should be nan!'
+               )
+        assert np.all(np.isnan(res)), mes
```
WenzDaniel, line 387: No, Idea how this slipped through the last review. I removed now the indentation level and also added a test for negative waveforms.

strax/processing/hitlets.py
```diff
@@ -133,219 +136,98 @@ def refresh_hit_to_hitlets(hits, hitlets):
...
+                r['time'] // r['dt'],
+                r['length'],
+                h['time'] // h['dt'],
+                h['length'])
+
+            if is_first_record:
+                is_first_record = False
+                recorded_samples_offset = h_start
+            h_start -= recorded_samples_offset
+            h_end -= recorded_samples_offset
```
jorana, line 220: This looks a bit funny, you do this for *each* record, not only the first?

strax/processing/peak_splitting.py
```diff
@@ -26,14 +26,10 @@ def split_peaks(peaks, records, to_pe, algorithm='local_minimum',
...
 
-    if data_type == 'hitlets':
-        # This is only needed once.
-        _, next_ri = strax.record_links(records)
-    elif data_type == 'peaks':
-        next_ri = None
-    else:
+    data_type_is_not_supported = data_type not in ('hitlets', 'peaks')
+    if data_type_is_not_supported:
         raise TypeError(f'Data_type "{data_type}" is not supported.')
```
jorana, line 31: I do appreciate your pilgrimage for clear code, but perhaps this is not really necessary right 😉 
```suggestion
    if data_type not in ('hitlets', 'peaks'):
        raise TypeError(f'Data_type "{data_type}" is not supported.')
```

strax/processing/peak_splitting.py
```diff
@@ -62,7 +58,7 @@ class PeakSplitter:
     find_split_args_defaults: tuple
```
jorana, line 58: Don't forget to update the docstring for `next_ri`

strax/processing/peak_splitting.py
```diff
@@ -202,8 +198,6 @@ def _split_hitlets(split_finder, peaks, orig_dt, is_split, min_area,
                 r = new_hits[offset]
                 r['time'] = h['time'] + prev_split_i * h['dt']
                 r['channel'] = h['channel']
-                # Hitlet specific
-                r['record_i'] = h['record_i']
```
jorana, line 206: Daniel, please correct me if I am wrong, but we might as well delete the `_split_hitlets` function, the two are now exactly the same apart from the `'max_gap'` line in peaks, which is not computed anyway. I think we can very simply clear up all this duplication

jorana (commented):
Thanks Daniel, I think it looks very good indeed. I like these changes a lot.

Perhaps now is a good time to get rid of the `_split_hitlets` as it is the same as the `_split_peaks` function

strax/processing/peak_splitting.py
jorana: Here is a diff:
afbeelding

jorana: Hmm okay, just looked again at https://github.com/AxFoundation/strax/pull/309 and remembered the pain

jorana: Perhaps, partially to make our lives easy, and on the other hand since I foresee a genuine advantage and no clear loss, why not add `max_goodness_of_split` to hitlets?

strax/dtypes.py
```diff
@@ -168,8 +168,6 @@ def hitlet_with_data_dtype(n_samples):
     n_samples = max(n_samples, 2)
     additional_fields = [(('Hitlet data in PE/sample with ZLE (only the first length samples are filled)', 'data'),
                            np.float32, n_samples),
-                         (('Fragment number in the pulse',
```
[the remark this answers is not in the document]
WenzDaniel: No, the data_type with data field is only used within the plugin.

strax/processing/peak_splitting.py
WenzDaniel: Yes, but same comment as above. I did not want to add another thing to make things more complicated :D Adding those fields is not any problem since we are copying in the end data from temp_hitlets_with_data_field to hitlets. So every field we add is only temporarily.

strax/dtypes.py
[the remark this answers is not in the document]
jorana: Thanks, good to know!

strax/processing/hitlets.py
```diff
@@ -133,219 +136,98 @@ def refresh_hit_to_hitlets(hits, hitlets):
...
-                
-            if current_ri == -1:
-                print('Record:\n', r, '\nHit:\n', hit)
-                raise ValueError('Was not able to find record_i')
-
-            if counter > TRIAL_COUNTER_NEIGHBORING_RECORDS:
-                print(ind, last_ri)
-                raise RuntimeError('Tried too often to find correct record_i.')
+    # Numba will not raise any exceptions if to_pe is too short, leading
+    # to strange bugs. This check is somewhat not complete, but should
```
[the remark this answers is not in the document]
jorana: Oh okay, haha, I was already thinking that I missed some big elephant. I would personally call this a very complete check 😉

strax/processing/hitlets.py
```diff
@@ -133,219 +136,98 @@ def refresh_hit_to_hitlets(hits, hitlets):
...
+        if data_is_not_empty:
+            raise ValueError('The data field of hitlets must be empty!')
+
+        data_field_not_long_enough = len(hitlets[0]['data']) < hitlets['length'].max()
+        if data_field_not_long_enough:
+            raise ValueError('The data field must be as large as the longest hitlet in our data.')
+
+        hitlets_with_data_field = hitlets
+    else:
+        n_samples = max(100, hitlets['length'].max())
```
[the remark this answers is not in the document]
jorana: sure, thanks, I remember now indeed. How do you want to proceed?

strax/processing/hitlets.py
```diff
@@ -133,219 +136,98 @@ def refresh_hit_to_hitlets(hits, hitlets):
...
+                             hitlets_with_data_field,
+                             '_copy_hitlets_to_hitlets_width_data')
+
+    _get_hitlets_data(hitlets_with_data_field, records, to_pe)
+    return hitlets_with_data_field
+
+
+@numba.jit(nopython=True, nogil=True, cache=True)
+def _get_hitlets_data(hitlets, records, to_pe):
+    rranges = _touching_windows(records['time'],
```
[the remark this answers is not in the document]
jorana: You mean `if r['channel'] != h['channel']:`, so this solves all problems because hitlets are fully inclusive of records and therefore a touching window can only yield touching one record as the boundaries of a hitlet are always broader/equal to records start/end time with some extra space left and right?

strax/processing/hitlets.py
jorana: okay, yes and `h_start` gets reset every record, whereas `recorded_samples_offset`, got ya

strax/processing/peak_splitting.py
jorana: okay, however you decide. I would at this point not be distracted by making things more simple by replacing a function by an already in place function which is tested.

But I see your point. On the other hand, just go for it 😄

strax/processing/hitlets.py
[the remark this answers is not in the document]
WenzDaniel: I am confused by your question "hitlets are fully inclusive of records" and "hitlet are always broader/equal to records start/end". These two statements are the opposite of another is not it? But either way let me try to clarify: 
A hitlet can in general touch multiple record fragments. A hitlet does need to be fully contained in a record due to the left/right hit extension. As fragments are sorted by time and channel this means that in principle a hitlet can also touch the fragments of another channel. Hence Tianyu required  `if r['channel'] != h['channel']:` to solve this problem.

WenzDaniel:
Okay, I addressed your comments, besides the ones which will require a larger refactoring.

strax/processing/hitlets.py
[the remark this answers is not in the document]
jorana: @WenzDaniel can you make this a kwarg of the function?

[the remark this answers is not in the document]
WenzDaniel: Okay done.

jorana (approved):

PR #435, opened 2021-05-01
fix empty hitlets

jorana:
Due to https://github.com/AxFoundation/strax/pull/430, straxen master is failing:
https://github.com/XENONnT/straxen/runs/2482801332

We did not consider empty hitlets

Return the empty hitlets-dtype if none was given

See the example in the updated test


WenzDaniel:
Already fixed it in another branch. Just checking the peak splitting that it does not introduce this memory leak again, then I will open it.

WenzDaniel:
See #436

PR #436, opened 2021-05-01
Refactor hitlets

WenzDaniel:
With this PR we refactor the hitlet data_type a bit. 

1. get_hitlet_data also supports now empty inputs. A corresponding test was added.
2. The the hitlet dtype with data field was extended by goodness of split and max gap. This allows to share the _split_peaks function of our PeakSplitter class. Hence we can remove some redundant code. 
3. I refactored straxen's hitlet plugin by adding a new function called `create_hitlets_from_hits`. This gives the plugin a bit more structure and generalizes hitlets such that it can also be used more easily for other analyses. Further, this allowed to remove the `refresh_hit_to_hitlets`.

This PR goes together with [#463](https://github.com/XENONnT/straxen/pull/463/files).

### DIFF
--- a/strax/processing/hitlets.py
+++ b/strax/processing/hitlets.py
@@ -165,7 +165,7 @@ def get_hitlets_data(hitlets, records, to_pe, min_hitlet_sample=100):
         (if it did not exists before it will be added.)
     """
     if len(hitlets) == 0:
-        return hitlets
+        return np.zeros(0, dtype=strax.hitlet_with_data_dtype(2))
 
     if len(hitlets) > 0 and len(records) == 0:
         raise ValueError('Cannot get for hitlets if records are empty!')
--- a/strax/processing/hitlets.py
+++ b/strax/processing/hitlets.py
@@ -12,6 +12,45 @@ NO_FWXM = -42  # Value in case FWXM cannot be found.
 # Hitlet building:
 # ----------------------
 @export
+def create_hitlets_from_hits(hits,
+                             save_outside_hits,
+                             channel_range,
+                             chunk_start=0,
+                             chunk_end=np.inf,
+                             min_samples_data_field=100, ):
+    """
+    Function which creates hitlets from a bunch of hits.
+
+    :param hits: Hits found in records.
+    :param save_outside_hits: Tuple with left and right hit extension.
+    :param channel_range: Detectors change range from channelmap.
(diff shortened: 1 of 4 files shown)


strax/processing/hitlets.py
```diff
@@ -7,12 +7,42 @@
...
+                             chunk_end=np.inf,):
+    """
+    Function which creates hitlets from a bunch of hits.
+
+    :param hits: Hits found in records.
+    :param save_outside_hits: Tuple with left and right hit extension.
+    :param channel_range: Detectors change range from channelmap.
+    :param chunk_start: Start of the chunk or run.
+    :param chunk_end: End of the chunk or run.
+    :param min_samples_data_field: Minimal length of the hitlet data field.
```
jorana, line 27: This is not a parameter of the function right

jorana (approved):
In principle this looks very good, I think all was already mentioned in #435 and #430.

Given our ordeal in #309, would you mind double checking that no annoying edge-cases are occurring when processing hitlets after peaks and vise versa since the dtypes might be slightly different?

WenzDaniel:
> Given our ordeal in #309, would you mind double checking that no annoying edge-cases are occurring when processing hitlets after peaks and vise versa since the dtypes might be slightly different?

I checked only for hitlets. But good point I should check for both.

strax/processing/hitlets.py
```diff
@@ -7,12 +7,45 @@
...
+
+    :param hits: Hits found in records.
+    :param save_outside_hits: Tuple with left and right hit extension.
+    :param channel_range: Detectors change range from channel map.
+    :param chunk_start: Start of either a chunk. Ensures that
+        no hitlet is extended beyond chunk boundaries. You can keep the
+        defaults if not used inside of a plugin.
+    :param chunk_end: End of a chunk Ensures that no hitlet is extended
+        beyond chunk boundaries. You can keep the  defaults if not used
+        inside of a plugin.
```
jorana, line 30:
```suggestion
    :param chunk_start: (optional) start time of a chunk. Ensures that
        no hitlet is earlier than this timestamp.
    :param chunk_end: (optional) end time of a chunk. Ensures that
        no hitlet ends later than this timestamp.
```

WenzDaniel:
Okay processing looks good.

Issue #246, opened 2020-03-14
In-rundoc defaults makes availability check slow


JelleAalbers:
The defaults-in-the-rundoc introduced in https://github.com/AxFoundation/strax/pull/223 make the data availability check very slow. For every run we currently send out a separate run metadata query to even be able to check if a piece of data is available. The result should be cached for each run, but it is still very inefficient.

There are (at least) two possible solutions:
  * Add a method to storage frontends to quickly retrieve run defaults for many runs. Databases can easily do this, for json files it won't be any faster than what we currently do.
  * Remove support for in-rundoc defaults. They aren't used yet and maybe there were a bad idea to begin with.

jorana:
I don't think this is ever used and may be obsolete due to CMT

jorana:
Fixed in #440

PR #440, opened 2021-05-08
Speed up run selection by ~100x for fixed defaults

jorana:
Run-selection is unnecessary slow. For each run, each target, the entire lineage is calculated, even when the `use_per_run_defaults` option is false. Also see issue #246.

This makes run-selection particularly slow. This PR solves this.

Using this PR, we calculate the lineage hash only once and use that for all the runs. Of course, one should take care drafting it's context to not by accident set the `use_per_run_defaults` to True even tough it is actually used. Perhaps we should set it to True by default to prevent such issues.

## How about performance.
For xenonnt I ran the following command:

```python
%%timeit
# Setting new context resets the cached st.runs
st = straxen.contexts.xenonnt_online()
st.context_config['check_available'] = ('raw_records', 'peak_basics', 'peaklets', 'event_info')
st.select_runs()
```

### Before
(I actually did a %%time since this took forever)
`Wall time: 15min 33s`

### After 
`9.74 s ± 80.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)`

## Changes on straxen
We should check if contexts actually should have this flag enabled (e.g. the 1T contexts still uses this setting).


JelleAalbers:
Very nice Joran! This is one aspect of strax I feared would become an issue eventually, but wasn't able to fix before I left.

If you rely on static lineages, you'll also want to disallow any options with `default_per_run` values, which are used in the 1T context/plugins. Unless you already changed it earlier, `use_per_run_defaults` only controls whether we check the run metadata for updated configs. It probably makes sense to expand `use_per_run_defaults`'s scope to disable all per-run defaults, e.g. crash when someone tries to register a plugin that has a non-None `default_by_run`. You'll then have to turn `use_per_run_defaults` on for 1T, but I assume that is becoming less and less of an issue :-)

Eventually, you will run into the same speed issue when loading high-level data for many runs (e.g. a whole science run). Superruns would be the best solution, since that would also avoid the slowdown from loading many tiny files over network filesystems. Otherwise, you'll need a deeper solution. Maybe something like this will work:
  - Keep a dict of the fixed lineages as a context attribute.
  - Intercept all config changes so you can update those lineages. For example, you could force/encourage everyone to use `.set_config`, or you could make `.config` a dict subclass with a custom `__setitem__` (and likewise the subdictionaries inside config).
  - Use the fixed lineage instead of `p.lineage` here https://github.com/AxFoundation/strax/blob/master/strax/context.py#L617, so you can replace this call to `get_plugins` https://github.com/AxFoundation/strax/blob/master/strax/context.py#L598 with an empty dict, which you fill the first time you have to actually compute something, e.g. here https://github.com/AxFoundation/strax/blob/master/strax/context.py#L664.

jorana:
@JelleAalbers , thanks for the feedback. I will make some changes to add indeed a crash if someone tries to use that feature when not allowed in the context.

For the latter points you raise, that indeed will save some time, but most importantly, it will save time upon loading data. Loading an entire science run will happen less often than `select_runs` which happens almost every time one starts a notebook. The computation time needed for `select_runs` is really limiting, for loading runs it is a bit less so (also because you take much less runs into account usually).

I will give your broader comment some additional thought, it was not per se what I intended addressing with this PR but if we are disallowing this for run-selection might as well go one level deeper indeed. We would also need to recompute it with temporary registered plugins since that should also reflect such as change in the lineage cache.

tests/test_context.py
```diff
@@ -166,3 +172,93 @@ def test_copy_to_frontend():
...
+        st.register(Records)
+        if not runs_default_allowed:
+            self.assertRaises(strax.InvalidConfiguration, st.register, Peaks)
+        else:
+            st.select_runs()
+
+    def test_register_with_defaults_and_allowed(self):
+        self.test_register_with_defaults(runs_default_allowed=True)
+
+    def test_register_all_no_defaults(self, runs_default_allowed=False):
```
WenzDaniel, line 215: Sorry, but could you explain me the difference to `test_register_with_defaults`  I cannot see any.

WenzDaniel (commented):
Nice PR, I learned a few new things today. I just have a single comment/question which I could not figure out. Can you clarify before approving?

tests/test_context.py
```diff
@@ -166,3 +172,93 @@ def test_copy_to_frontend():
...
+
+class TestPerRunDefaults(unittest.TestCase):
+    """Test the per-run defaults options of a context"""
+    def setUp(self):
+        """
+        Get a folder to write (Similar to tempfile.TemporaryDirectory) to get
+        a temp folder in the temp dir. Cleanup after the tests.
+        """
+        temp_folder = uuid.uuid4().hex
+        self.tempdir = os.path.join(tempfile.gettempdir(), temp_folder)
```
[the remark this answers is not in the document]
jorana: I just wanted a random hash, this was an easy solution

tests/test_context.py
```diff
@@ -166,3 +172,93 @@ def test_copy_to_frontend():
...
+    @staticmethod
+    def _has_per_run_default(plugin) -> bool:
+        """Does the plugin have at least one option that is a per-run default"""
+        has_per_run_defaults = False
+        for option in plugin.takes_config.values():
+            has_per_run_defaults = option.default_by_run != strax.OMITTED
+            if has_per_run_defaults:
+                # Found one option
+                break
+        return has_per_run_defaults
```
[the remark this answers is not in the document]
jorana: Depends, for some of the test cases I want no per run defaults, whereas for others I explicitly want to have per run defaults. E.g. here I want to have only per run defaults for Peaks, not records:
```python
        assert not self._has_per_run_default(Records)
        assert self._has_per_run_default(Peaks)
```

whereas here, I don't really care which of the two has run-defaults:
```
        assert any([self._has_per_run_default(Peaks),
                    self._has_per_run_default(Records)])
```

tests/test_context.py
jorana: yes, I just wanted to to check the `st.register_all(strax.testutils)` function. Such that this is affected treated similarly to the `st.register` function

WenzDaniel: Ah okay, thanks. I am not sure if we gain a lot by testing st.register_all too. Under the hood it calls st.register anyway.

tests/test_context.py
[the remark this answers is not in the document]
WenzDaniel: Okay this explains better the main difference between your two tests above, thanks.

WenzDaniel (approved):
Thanks a lot.

tests/test_context.py
jorana: I agree, I wondered about it, so looked it up and convinced myself that this test was probably not needed. However, it was easy to add and the tests completes within <seconds, so it does not hurt either.

PR #441, opened 2021-05-17
Use zstd for from base-env for testing

jorana:
Testing is failing since ztsd 1.5 is not being installed properly, grep the right version from the base env.


jorana:
Thanks, this seems to work well!

jorana:
fix it hard to also fix the docs building: 
cc783c1b5ffb216df61d2bb72752a81bb7748fd6

PR #442, opened 2021-05-18
Add MB/s pbar

jorana:
Slightly compactify the progress-bar, print in MB/s (NB! MB referring to the _output_ data)

**Why is this useful?**
I think potential new bottle necks can be spotted earlier since one will develop a feeling of what a good rate for a data-type is.

Divide the output size by the time since last chunk

old:
afbeelding

new:
afbeelding


WenzDaniel (approved):
Sounds like a good addition, thanks a lot.

PR #444, opened 2021-05-21
Change non thread safe iteration

petergaemers:
During testing of the mc_chain we ran into an error. When starting up the mailboxes we iterate over a list of mailboxes. However during the iteration this dict changes, leading to a crash. Because you're not allowed to change something you're iterating over. 
This results in the following error:
`  File "/opt/XENONnT/anaconda/envs/XENONnT_development/lib/python3.8/site-packages/strax/processor.py", line 216, in iter
    for m in self.mailboxes.values():
RuntimeError: dictionary changed size during iteration`

The problem is this way of iterating is aparantly bad and not tread safe. After some googling this came up as a solution: https://stackoverflow.com/questions/45276407/dictionary-changed-size-during-iteration-in-multithreading-app (oke it was the very first thing to show up)

So by making a copy of this list we solve our issue

There is a big however though: maybe this actually just sweeps something under the carpet

### DIFF
--- a/strax/processor.py
+++ b/strax/processor.py
@@ -213,7 +213,7 @@ class ThreadedMailboxProcessor:
         final_generator = self.mailboxes[target].subscribe()
 
         self.log.debug("Starting threads")
-        for m in self.mailboxes.values():
+        for m in list(self.mailboxes.values()):
             m.start()
 
         self.log.debug(f"Yielding {target}")


jorana:
Hi Peter, as also mentioned in the link you referenced, this should only happen when we are changing the mailbox dict itself right? Have you been able to trace back where this threads get lost or deleted? I would be slightly worried if we do change the number of threads somehow intentionally.

petergaemers:
I agree, I'm not sure if this is a very bad band aid which actually not really deals with the issue.
I think it'd be useful if someone more knowledgable about multithreading can comment. Maybe @JelleAalbers or @tunnell

tunnell:
Have you run it through the python debugger to go step by step?  Also, maybe @jmosbacher has ideas at some point

tunnell:
PyCharm has a "professional" version that is free for academics, so you should be able to get this @petergaemers.  https://www.jetbrains.com/help/pycharm/debugging-code.html

ramirezdiego:
@Peter Gaemers This is a line that should work out of the box for the MC chain.
```
git clone https://github.com/XENONnT/mc_chain.git
cd mc_chain
module load singularity && singularity shell --bind /scratch,/project2,/dali,/cvmfs /project2/lgrandi/xenonnt/singularity-images/xenonnt-montecarlo-development.simg
# Make sure you don't have your own strax/en installation sourced.
python run_wfsim.py --G4File /dali/lgrandi/ryuichi/public/mc_product/4.0.2/cryostat/nv_Cryostat_0.root \
--RunID "mc_0123"  --CmtRunID "0" --Experiment XENONnT --WfsimConfigFile fax_config_nt_low_field.json \
--WfsimConfigFileNveto fax_config_nt_nveto.json --OutputFolder ./output/ --EntryStart 0 \
--EntryStop 100 --EventRate 100 --ChunkSize 5 --EpixMicroSeparation 0.05 --NeutronVeto
```

Try removing the `--NeutronVeto` option and see what happens. If it works too, then try to fetch a file with only TPC events and run exactly the same line (without `--NeutronVeto` flag, of course).

jmosbacher:
From looking at the code it looks like the issue is with the MailboxDict lazy attribute.
from the ThreadedMailboxProcessor.__init__:  
```python
self.mailboxes = MailboxDict(lazy=lazy)
```
You can see that mailboxes is a subclass of dict that acts similar to a defaultdict when lazy is true.

```python
#strax.processor.py
class MailboxDict(dict):
    def __init__(self, *args, lazy=False, **kwargs):
        super().__init__(*args, **kwargs)
        self.lazy = lazy

    def __missing__(self, key):
        res = self[key] = strax.Mailbox(name=key + '_mailbox',
                                        lazy=self.lazy)
        return res

```
And since temporary plugins are created for multi output plugins, the mailboxes content can change.
From what i can tell this happens in the divide_outputs function in mailbox.py

petergaemers:
@jmosbacher I've tried setting lazy=False and this resulted in a different crash

jmosbacher:
@petergaemers yeah i wouldnt expect that to help because then the thread trying to create these on-the-fly plugins would try to get a mailbox that doesnt exist in processor.mailboxes and would get a KeyError exception...

jmosbacher:
Im not sure its a trivial fix. I would have to investigate a bit more and maybe actually run some code... not sure when i would have time to do this. what is the deadline for this fix?

jmosbacher:
actually looking again, the lazy flag is not connected to this behavior. It behaves as a defaultdict in any case.

jmosbacher:
@petergaemers i think if you add a line like `[self.mailboxes[d] for d in p.provides]` anywhere within the IF block `if p.multi_output:` in the ThreadedMailboxProcessor init method (line 136 in processor.py) I think all the mailboxes that will be used in the thread will be created ahead of time. If you have everything setup to test this would be great. I wont have time to setup a test env that reproduces this at least until the weekend.
```python
if p.multi_output:
    [self.mailboxes[d] for d in p.provides] # <-- Add this line here
    multi_output_seen.append(p)

    # Create temp mailbox that receives multi-output dicts
    # and sends them forth to other mailboxes
    mname = p.__class__.__name__ + '_divide_outputs'
    self.mailboxes[mname].add_sender(
        p.iter(
            iters={dep: self.mailboxes[dep].subscribe()
                    for dep in p.depends_on},
            executor=executor),
        name=f'divide_outputs:{d}')

    
    self.mailboxes[mname].add_reader(
        partial(strax.divide_outputs,
                lazy=lazy,
                mailboxes=self.mailboxes,
                flow_freely=to_flow_freely,
                outputs=p.provides))
```

JelleAalbers:
I think Yossi is right, I don't see how mailbox creation could get so delayed except from `divide_outputs` (which I will immediately admit is rather kludgy, multi-output plugins were an aftertought).

However, we should already be making a mailbox for each produced output. All outputs should be read by somebody, and to register a reader, you call `self.mailboxes[d].add_reader` or `self.mailboxes[d].subscribe`, which create the mailbox through the defaultdict magic. Even an output that is neither saved nor required by another plugin should still be read by the discarder here https://github.com/AxFoundation/strax/blob/master/strax/processor.py#L190. 

You could force-create all the mailboxes explicitly, but mailboxes without a reader will crash when you try to start them (`Attempt to start mailbox ... without subscribers`).

Peter's change would avoid the 'dictionary changed size' error, but if a new mailbox is still made during the iteration, it will never even try to get started (so you won't see this no-subscriber crash). If someone sends it messages anyway, I think those will just pile up until your memory is full (in lazy mode) or hang strax when the mailbox buffer is full (non-lazy mode).

Maybe you can do `self.mailboxes = dict(self.mailboxes)` at the end of `ThreadedMailboxProcessor.__init__` to remove the defaultdict behavior after it should no longer be invoked, so you don't have to think about it anymore. Then you'll get a KeyError at the exact place where someone is trying to send to a non-existing mailbox, and maybe that will tell you what is really going on.

(If we wouldn't have this check https://github.com/AxFoundation/strax/blob/master/strax/plugin.py#L502 I would have guessed some multi-output plugin was producing an output that doesn't match its `provides`. Or maybe that someone was modifying their own `provides`.)

petergaemers:
Thanks for your replies Jelle and Yossi. With some more digging it looks like some of the provides don't have savers:

```DEBUG:ThreadedMailboxProcessor:Processor components are: ProcessorComponents(plugins={'raw_records': RawRecordsFromMcChain}, loaders={}, savers={'raw_records_aqmon': [], 'raw_records_nv': [<strax.storage.files.FileSaver object at 0x7f8fa0049820>], 'raw_records_he': [<strax.storage.files.FileSaver object at 0x7f8fa0049040>], 'truth': [], 'raw_records': [<strax.storage.files.FileSaver object at 0x7f8f9ffe6580>], 'truth_nv': []}, targets=('raw_records',))
```
raw_records_aqmon, truth and truth_nv don't get a saver, so no subscribers, and therefor get killed.
I'm still not sure though why this is happening

JelleAalbers:
Thanks Peter, that helps! So some datatypes get an entry in `savers`, but have no actual saver. Then https://github.com/AxFoundation/strax/blob/master/strax/processor.py#L107 thinks that data is saved, and the discarder logic does not trigger.

So the only question is how the empty lists get into savers. Your debug message is from before ParallelSourcePlugin gets its hands on the components, so the code here https://github.com/AxFoundation/strax/blob/master/strax/plugin.py#L909 is off the hook. That leaves context.py.

I think the offender has to be here: https://github.com/AxFoundation/strax/blob/master/strax/context.py#L765. Since `savers` is a defaultdict at this stage, trying to append an element will auto-create an empty list, even an exception occurs while we actually make the element (in this case, probably it's the frontend refusing to save due to datatype exclusions or directory permissions). To prove this, try:

```python
from collections import defaultdict
q = defaultdict(list)
q[4].append(1/0)
```

Even though this crashes with ZeroDivisionError, q will now have an empty list under 4. Defaultdicts are tricker than I thought...

@petergaemers I'll make a short PR to fix this, maybe you could test that branch on your setup?

petergaemers:
https://github.com/AxFoundation/strax/blob/master/strax/context.py#L765 is indeed the line where the exception happend. Your PR https://github.com/AxFoundation/strax/pull/452 fixes the issue. So we can close this

PR #445, opened 2021-05-21
Don't ignore if pattern also startswith underscore

jorana:
@WenzDaniel just noticed this, let's not remove the underscore from the tag if the pattern also has an underscore.


jingqiangye (approved):
@jorana Thanks for the PR! I just tested it and indeed it nicely solves my issue (https://github.com/AxFoundation/strax/issues/448). Happy to approve ;)

PR #446, opened 2021-05-21
Update run_selection.py

WenzDaniel:
It feels a bit odd that official tags are excluded by default from the run_selection search. In this PR we change the setting


WenzDaniel:
Solved in https://github.com/AxFoundation/strax/pull/445/files

PR #447, opened 2021-05-25
Temporary patch div/zero in hitlets

jorana:
We are seeing some runs (020330) where nveto is failing because of division by zero errors caused here.

We need to figure out where this happens, but that needs some deeper investigation.


WenzDaniel:
Thanks we should also add it in line 367 I cannot use "make a suggestions".

jorana:
@WenzDaniel , no if you return you don't hit that line right?

WenzDaniel:
Yes, but you recompute m in line 367 and hence you can potentially hit the same issue in line 368 again. As I do not know yet why out of the sudden we run into this issue I prefer to do it on both.

jorana:
Right, good point

PR #450, opened 2021-05-27
Match cached buffer chunk start times OverlapWindowPlugin

jorana:
In the overlap window plugin, we are strict for the end-time of the cached chunks. However, we are not so strict about our start times in the cached_input. This PR addresses this.

Since we use `v.split(t=prev_split, allow_early_split=True)`, we can get chunks split on different times (the allow_early_split is True). This causes issues when one from chunk>0 tries to add the cached input into the `super().do_compute`. We don't want to `allow_early_split=False` since that raises an error to deal with. Instead, we can use the split point (i.e. start time) from the other chunk(s) to split on since allow_early_split=True. This will only increase the bit of information that is being send to the compute argument since we allow data from slightly more time to be added to the compute argument of the plugin.

This should all be safeguarded (that we don't store data twice) by `self.sent_until` (otherwise we could have never done `cache_inputs_beyond = int(self.sent_until- 2 * self.get_window_size() - 1)`)

This PR is needed in order to be able to run this straxen PR:https://github.com/XENONnT/straxen/pull/296

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -637,8 +637,8 @@ class OverlapWindowPlugin(Plugin):
                 _, self.cached_input[k] = v.split(t=prev_split,
                                                   allow_early_split=True)
                 prev_split = self.cached_input[k].start
-            n_start_times = len(set([c.start for c in self.cached_input.values()]))
-            if n_start_times == 1:
+            chunk_starts_are_equal = len(set([chunk.start for chunk in self.cached_input.values()])) == 1
+            if chunk_starts_are_equal :
                 self.log.debug(
                     f'Success after {counter}. '
                     f'Extra time = {cache_inputs_beyond-prev_split} ns')
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -591,10 +591,10 @@ class OverlapWindowPlugin(Plugin):
             raise RuntimeError("OverlapWindowPlugin must have a dependency")
 
         # Add cached inputs to compute arguments
-        for k, v in kwargs.items():
+        for data_kind, chunk in kwargs.items():
             if len(self.cached_input):
-                kwargs[k] = strax.Chunk.concatenate(
-                    [self.cached_input[k], v])
+                kwargs[data_kind] = strax.Chunk.concatenate(
+                    [self.cached_input[data_kind], chunk])
 
         # Compute new results
         result = super().do_compute(chunk_i=chunk_i, **kwargs)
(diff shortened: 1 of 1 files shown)


ershockley:
@jorana I would like to review ASAP but I am too ignorant of this stuff to do it in a meaningful way, at least right now. Can you point me to e.g. a relevant straxferno lesson or can we maybe have a chat to explain? Otherwise I defer to @WenzDaniel.

strax/plugin.py
```diff
@@ -627,10 +627,29 @@ def do_compute(self, chunk_i=None, **kwargs):
...
+        # Cache inputs, make sure that the chunks start at the same time to
+        # prevent issues in input buffers later on
+        prev_split = cache_inputs_beyond
+        max_trials = 10
+        for counter in range(max_trials):
+            for k, v in kwargs.items():
+                _, self.cached_input[k] = v.split(t=prev_split,
+                                                  allow_early_split=True)
+                prev_split = self.cached_input[k].start
+            n_start_times = len(set([c.start for c in self.cached_input.values()]))
```
WenzDaniel, line 640:
```suggestion
            chunk_starts_are_equal = len(set([chunk.start for chunk in self.cached_input.values()])) == 1
```

strax/plugin.py
```diff
@@ -627,10 +627,29 @@ def do_compute(self, chunk_i=None, **kwargs):
...
+            if n_start_times == 1:
```
WenzDaniel, line 641:
```suggestion
            if chunk_starts_are_equal :
```

strax/plugin.py
WenzDaniel: One easily misses the set.

WenzDaniel (approved):
Looks very nice, beside the renaming of k and v I just made another naming suggestion based on the things I learned from Uncle Bob :D. I was first puzzled by the == 1 in the if statement because I missed the `set` before. Feel free to change it if you like.

jorana:
Thanks @WenzDaniel for the review, it's much appreciated

ershockley (approved):
The renaming helps a lot! I still don't 100% understand but trust you guys

jorana:
> 
> 
> Nice catch Joran! This looks similar to https://github.com/AxFoundation/strax/blob/master/strax/plugin.py#L354; maybe it's worth (eventually) writing a `split_multiple` function that splits several chunks on the closest split time that is legal / doesn't intersect data in any chunk?

Thanks @JelleAalbers , sounds like a good idea. We are effectively doing that now right (only we now do it here in `plugin.py` (twice) instead of `chunk.py`)?

PR #452, opened 2021-06-02
Prevent late creation of unattended mailboxes

JelleAalbers:

Due to a few subtle interacting bugs / missing checks, a situation can occur where a new mailbox gets created __after__ processing has started. If this happens during https://github.com/AxFoundation/strax/blob/master/strax/processor.py#L216 we get the crash Peter reported in https://github.com/AxFoundation/strax/pull/444. But in principle this could happen at any later time, in which case messages would probably accumulate in the unattended mailbox until we run out of memory. For more details, see https://github.com/AxFoundation/strax/pull/444. 

This will:
  * Replace the `MailboxDict` with a regular dict at the end of `ThreadedMailboxProcessor.__init__`, stopping automatic mailbox creation beyond that point. If (through some bug) someone tries to send to mailboxes['nonexisting'], they will now get the expected `KeyError`.
  * Fix the discarder-detection code at https://github.com/AxFoundation/strax/blob/master/strax/context.py#L765. If (through some bug / oversight) empty-list entries are left in `savers`, we will still add mailboxes and discarders for those outputs when they are produced but not consumed.
  * Rewrite the `savers`-construction code in `context.py` to prevent creation of empty-list entries. (When multiprocessing, `ParallelSourcePlugin` can cause new empty-list entries to appear, but it already cleans them up: https://github.com/AxFoundation/strax/blob/master/strax/plugin.py#L934.)

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -612,7 +612,7 @@ class Context:
         # have to do computation. (their instances will stick around
         # though the .deps attribute of plugins that do)
         loaders = dict()
-        savers = collections.defaultdict(list)
+        savers = dict()
         seen = set()
         to_compute = dict()
 
@@ -736,7 +736,7 @@ class Context:
 
             # Save the target and any other outputs of the plugin.
             for d_to_save in set([d] + list(p.provides)):
-                if d_to_save in savers and len(savers[d_to_save]):
+                if savers.get(d_to_save):
                     # This multi-output plugin was scanned before
                     # let's not create doubled savers
                     assert p.multi_output
@@ -762,13 +762,15 @@ class Context:
                             pass
                     # If we get here, we must try to save
                     try:
-                        savers[d_to_save].append(sf.saver(
+                        saver = sf.saver(
                             key,
                             metadata=p.metadata(
                                 run_id,
                                 d_to_save),
-                            saver_timeout=self.context_config['saver_timeout']
-                        ))
+                            saver_timeout=self.context_config['saver_timeout'])
+                        # Now that we are surely saving, make an entry in savers
+                        savers.setdefault(d_to_save, [])
+                        savers[d_to_save].append(saver)
                     except strax.DataNotAvailable:
                         # This frontend cannot save. Too bad.
                         pass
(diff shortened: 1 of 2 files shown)


petergaemers (approved):

petergaemers:
Ha Jelle, this indeed fixes the problem! Thank you very much

jorana:
Thanks @JelleAalbers !

jorana:
@JelleAalbers @petergaemers @jmosbacher , unfortunately this PR leads to a keyerror on this line:
https://github.com/AxFoundation/strax/blob/master/strax/processor.py#L220

```bash
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/xedaq/miniconda/envs/py38/lib/python3.8/multiprocessing/process.py", line 108, in run
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1350, in run_strax
    st_make()
  File "/home/xedaq/software/straxen/bin/bootstrax", line 1328, in st_make
    st.make(run_id, targets,
  File "/home/xedaq/software/strax/strax/context.py", line 1117, in make
    for _ in self.get_iter(run_ids[0], targets,
  File "/home/xedaq/software/strax/strax/context.py", line 1002, in get_iter
    generator.throw(e)
  File "/home/xedaq/software/strax/strax/context.py", line 975, in get_iter
    for n_chunks, result in enumerate(strax.continuity_check(generator), 1):
  File "/home/xedaq/software/strax/strax/chunk.py", line 266, in continuity_check
    for s in chunk_iter:
  File "/home/xedaq/software/strax/strax/processor.py", line 220, in iter
    final_generator = self.mailboxes[target].subscribe()
KeyError: 'led_calibration'
```
I'm not sure if this bug is in the ParallelProcessPlugin or the ThreadedMailboxProcessor. I guess the latter as the empty mailbox that would otherwise be created from the default dict behavior that is now replaced by by the `'ParallelSourcePlugin_divide_outputs'` mailbox

PR #463, opened 2021-06-07
Fix #452 assert there is a mailbox for the final generator

jorana:
_this PR was edited after Yossi's review and uses his proposal instead_
In PR #452 we have set the behavior of the mailbox dict to be non-defaultdict like. This however causes problems when only ParallelSourcePlugin plugins are subscribed since the results from these plugins are saved under the divide_output mailbox. 

As per @jmosbacher 's suggestion https://github.com/AxFoundation/strax/pull/463#issuecomment-857522815, the correct way of dealing with it is making sure there is a mailbox for each of the outputs of a multi-output plugin.

**The error:**
https://github.com/AxFoundation/strax/pull/452#issuecomment-855790815

### DIFF
--- a/strax/processor.py
+++ b/strax/processor.py
@@ -160,6 +160,7 @@ class ThreadedMailboxProcessor:
                 self.mailboxes[mname].add_reader(
                     partial(strax.divide_outputs,
                             lazy=lazy,
+                            # make sure to subscribe the outputs of the mp_plugins
                             mailboxes={k: self.mailboxes[k] for k in p.provides},
                             flow_freely=to_flow_freely,
                             outputs=p.provides))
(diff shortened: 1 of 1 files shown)


jmosbacher:
@jorana Does this actually work? if there is no yielding from the final generator wouldnt the iter function just go straight to closing the threads? then it would be a crap shot whether the processing was done before the threads were killed no? i feel like i am missing something. maybe im misunderstanding the circumstance when this occurs?
Im wondering if something simpler like [this](https://github.com/jmosbacher/strax/blob/3fdc636ea540d8ab8a03025406eeacd71a5edb39/strax/processor.py#L155) would work. wouldnt that pre-create mailboxes for all the outputs of the ParallelSourcePlugin or any other multi-output plugin that someone may add?
Im just afraid not having a final_generator may cause issues in some cases e.g. if somehow when in lazy mode we end up with nothing pulling at the end of the processing chain and the processor just doesnt do anything.

jorana:
Thanks Yossi, let me checkout what you propose
> @jorana Does this actually work? if there is no yielding from the final generator wouldnt the iter function just go straight to closing the threads?

The `mailbox.cleanup()` just refers to Thread.join() so you would be waiting for all the threads to finish their work right?
https://github.com/jmosbacher/strax/blob/3fdc636ea540d8ab8a03025406eeacd71a5edb39/strax/mailbox.py#L204

I tested it and all the data was successfully written it seemed.

> then it would be a crap shot whether the processing was done before the threads were killed no? i feel like i am missing something. maybe im misunderstanding the circumstance when this occurs?

It did work for at least the run I tried and I believe it's what we have effectively been doing always because we were only yielding from a late created mailbox before.

> Im wondering if something simpler like [this](https://github.com/jmosbacher/strax/blob/3fdc636ea540d8ab8a03025406eeacd71a5edb39/strax/processor.py#L155) would work. wouldnt that pre-create mailboxes for all the outputs of the ParallelSourcePlugin or any other multi-output plugin that someone may add?

Let me try, it indeed looks more elegant.

> Im just afraid not having a final_generator may cause issues in some cases e.g. if somehow when in lazy mode we end up with nothing pulling at the end of the processing chain and the processor just doesnt do anything.
Right, this actually should only happen if we are using the ParallelSourcePlugin which is only allowed in multiprocessing which is only allowed if lazy is not allowed. We could add a check that at least the ParallelSourcePlugin is in the mailboxes.

Let me edit and update this comment while I checkout your sugestion

jorana:
Thanks @jmosbacher , I think your proposed solution looks better and cleaner. Although for my understanding, how are we sure that this specific mailbox we are subscribing to is actually yielding anything useful we can use?

jmosbacher:
@jorana glad to know this actually works (or at least looks like it does).
Do you mean final_generator? are you worried it will be something random that is not needed?

jorana:
All seems to work well on the strax side, thanks Yossi!

JelleAalbers:
Nice fix! Sorry, I should have put the MailboxDict -> dict conversion after the creation of the final_generator -- but creating mailboxes explicitly with producers (rather than the readers) as you do here is probably even clearer.

PR #464, opened 2021-06-07
Fix digitizer saturation check

zhut19:
Fix a bug that's causing pulse saturation check not working. The right condition should be `max_in_record >= int(baseline)`, as `record = int(baseline) -  raw_record` and pulse saturation results into `raw_record = 0`

### DIFF
--- a/strax/processing/peak_building.py
+++ b/strax/processing/peak_building.py
@@ -235,7 +235,7 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
                 p['time'] // dt, n_p)
 
             max_in_record = r['data'][r_start:r_end].max() * multiplier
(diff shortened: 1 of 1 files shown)


jorana:
Looking at the code and the dtypes, I think this makes sense, thanks Tianyu. 

Do you want me too have a look in the data too or have you found/done sufficient evidence and tests such that you are confident?

Other than that, I think we might want to increment the Peaklets version for this right?

jorana (approved):

zhut19:
Thank @jorana, you are welcome to check, so there's less room for a mistake again. A few more things to do for this to work again (I mean before we bump the Peaklet version).
  - Fixing another baseline issue in straxen `_peak_saturation_correction_inner`
  - Allow partial computation of peak width in strax `compute_widths`
  - Add width computation after saturation correction in `Peaklets`

PR #465, opened 2021-06-08
Add select index to compute width

zhut19:

After reconstructing peaklets with saturation correction, we need to update the widths for selected peaklets. For this purpose, I propose to add a `select_peaks_indicies` key to the compute width function.

I couldn't find a way to do this without changing this function in strax because numpy structured array assignment is a bit tricky i.e.
`arr[index]['x'] = new_values` will not update the values
`arr['x'][index] = new_values` will update the values

### DIFF
--- a/strax/processing/peak_properties.py
+++ b/strax/processing/peak_properties.py
@@ -70,7 +70,9 @@ def compute_index_of_fraction(peak, fractions_desired, result):
 @export
 def compute_widths(peaks, select_peaks_indices=None):
     """Compute widths in ns at desired area fractions for peaks
-    returns (n_peaks, n_widths) array
+    :param peaks: single strax peak(let) or other data-bearing dtype
+    :param select_peaks_indices: array of integers informing which peaks to compute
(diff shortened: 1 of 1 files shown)


skazama:
Hi Tianyu, thanks for the PR. I just had a look at the code and looks fine (only changes for peaks with saturated channels). 
@jorana I think this can be ok, but can you please check it too as this PR should be included for the reprocessing campaign.

strax/processing/peak_properties.py
```diff
@@ -68,12 +68,18 @@ def compute_index_of_fraction(peak, fractions_desired, result):
 
 
 @export
-def compute_widths(peaks):
+def compute_widths(peaks, select_peaks_indices=None):
     """Compute widths in ns at desired area fractions for peaks
```
jorana, line 72: could you update the dosctring to include the select_peaks_indices and what None means?

jorana (commented):
Thanks @zhut19 I think it looks good, just one request:

jorana (approved):

PR #466, opened 2021-06-11
CutPlugins save when plugin is target

WenzDaniel:
Simple PR, the name says already all. Cut plugins can take a few minutes to compute. Hence, we should allow users to store the cuts instead of recomputing them every time they start an analysis.


jorana (approved):
Sounds good!

PR #471, opened 2021-06-15
Document fuzzy-for and documentation itself

jorana:
 - Fuzzy for was not documented very well, let's add some.
 - How to maintain the documentation was not mentioned anywhere. Let's also add that

That is what these docs are about

  - Update the documentation 🎉


WenzDaniel (approved):
Sounds good thanks a lot.

PR #472, opened 2021-06-17
Fix define runs and allow storing of superuns

WenzDaniel:
Superruns are composed of many smaller sub-runs. They can be used to group runs for some given logical structure. So far we have not used them since due to some changes in the past the function define_run in run_selection.py broke. With this PR I would like to achieve two things

1. Fix define runs such that we can define superruns again. 
2. Add the possibility not only to create superruns on the run-metadata level, but also to "create" superruns by loading and rechunking the data of the subruns and storing them to disk. 

Regarding point 2. I had to add some changes. I extended our chunks module such that we can concatenate chunks of different run_ids.  For this purpose I added the following changes (+ some other changes for run definition and selection):

Chunks:
- Chunks of superruns have an additional field called subruns with min/max start/end of the all subrun chunks included in the corresponding superrun chunk
- A new function `transform_chunk_to_superrun_chunk` which converts a regular chunk into a superrun chunk
- concatanate chunks was also updated accordingly. 

Context:
- New option if true allow storing of new rechunked superrun chunks.
- Omit lineage check for individual subruns this may collide with https://github.com/AxFoundation/strax/pull/483
- Fixed define_run according to latest changes. I left the code to make superruns from data although it does not fully work yet. If needed I will make the required changes in a later PR.
- scan_runs added superrun support

Storage common:
- Support for subrun chunk superrun chunk transformation

I made a [notebook](https://github.com/XENONnT/analysiscode/blob/master/StraxTests/CreateLoadSuperRuns.ipynb) in which I compare the performance of 400 test subruns compared to a single superrun. I also checked the performance impact when applying cuts via cut-plugins. In average we get a speed boost of about a factor of ~5-20.

Please also see the corresponding straxen PR: https://github.com/XENONnT/straxen/pull/554/files which includes an additional example notebook.


WenzDaniel:
Please do not update with the master branch for the moment.

strax/chunk.py
```diff
@@ -332,3 +360,29 @@ def split_array(data, t, allow_early_split=False):
         t = min(data[splittable_i]['time'], t)
 
     return data[:splittable_i], data[splittable_i:], t
+
+
+@export
+def transform_chunk_to_superrun_chunk(superrun_id, chunk, prev_chunk_end=None):
```
jorana, line 366: Daniel, would you mind explaining why we need this? I'm a bit hesitant to say it's correct as it appears you could have the following situation:

```
run 1: |-----|----------
run 2: -------|----|----
run 3: -------------|--|
r1+3 : |-----|---------|
```
where the start time of the subrun3 chunk is well before the start time of that subrun. Wouldn't this be a bit strange and possibly cause funny things later?

strax/context.py
```diff
@@ -747,10 +760,15 @@ def concat_loader(*args, **kwargs):
                 to_compute[d] = p
                 for dep_d in p.depends_on:
                     check_cache(dep_d)
-
+
```
jorana, line 763:
```suggestion

```

strax/run_selection.py
```diff
@@ -260,27 +263,38 @@ def define_run(self: strax.Context,
                data: ty.Union[np.ndarray, pd.DataFrame, dict],
                from_run: ty.Union[str, None] = None):
     if isinstance(data, (pd.DataFrame, np.ndarray)):
```
jorana, line 265: Thanks for this cleanup, looking at your changes, I remember how buggy this was

strax/run_selection.py
```diff
@@ -289,17 +303,24 @@ def define_run(self: strax.Context,
         raise ValueError(f"Can't define run from {type(data)}")
 
     # Find start and end time of the new run = earliest start time of other runs
-    run_md = dict(start=float('inf'), end=0, livetime=0)
+    run_md = dict(start=datetime.datetime.max.replace(tzinfo=pytz.utc),
```
jorana, line 306: could use a `defaultdict(int)` here

strax/run_selection.py
```diff
@@ -289,17 +303,24 @@ def define_run(self: strax.Context,
...
+        run_md['start'] = min(run_md['start'], run_doc_start)
+        run_md['end'] = max(run_md['end'], run_doc_end)
+        time_delta = run_doc_end - run_doc_start
+        run_md['livetime'] += time_delta.total_seconds()*10**9
 
     # Superrun names must start with an underscore
     if not name.startswith('_'):
         name = '_' + name
-
+
```
jorana, line 323:
```suggestion

```

strax/storage/common.py
```diff
@@ -551,23 +557,46 @@ def save_from(self, source: typing.Generator, rechunk=True, executor=None):
...
+                            next_chunk = next(source)
+
+                            if _is_super_run:
+                                # If we are creating a superrun, we load data from subruns
+                                # and the loaded subrun chunk becomes a superun chunk:
+                                next_chunk = strax.transform_chunk_to_superrun_chunk(run_id, 
+                                                                                     next_chunk, 
+                                                                                     last_chunk_end)
+                                last_chunk_end = next_chunk.end
+
```
jorana, line 581: I might now see why you made this transform_chunk_to_superrun_chunk function (see above). But would the issue you are trying to solve not much easier be addressed in `continuity_check`?

strax/storage/common.py
```diff
@@ -551,23 +557,46 @@ def save_from(self, source: typing.Generator, rechunk=True, executor=None):
...
+                                                                            last_chunk_end)
+                            last_chunk_end = chunk.end
+
                 except StopIteration:
                     exhausted = True
 
                 if chunk is None:
                     break
-
+
```
jorana, line 599:
```suggestion

```

strax/testutils.py
```diff
@@ -191,6 +194,26 @@ def compute(self, records):
         p['area'] = self.config['base_area'] + self.config['bonus_area']
         return p
 
+# Same as Peaks but without any default_by_run options.
+@strax.takes_config(
+    strax.Option('base_area', default=0),
+    strax.Option('give_wrong_dtype', default=False),
+    strax.Option('bonus_area', default=42))
+class PeaksWoByRunDefault(strax.Plugin):
```
jorana, line 202: probably you tried, but sub-classing Peaks did not work?

strax/run_selection.py
```diff
@@ -260,27 +263,38 @@ def define_run(self: strax.Context,
                data: ty.Union[np.ndarray, pd.DataFrame, dict],
```
jorana, line 263: Would you mind adding a docstring to this function? Perhaps for data it might be worth adding what data it is or rename the variable as it's very generic

tests/test_superruns.py
```diff
@@ -0,0 +1,247 @@
...
+                                     register=[Records, RecordsExtension],
+                                     use_per_run_defaults=False)
+        self.context.set_context_config({'write_superruns': True})
+        
+        logger = self.context.log
+        logger.addFilter(lambda s: not re.match(".*Could not estimate run start and end time.*",
+                                                s.getMessage()))
+
+        self._create_subruns()
+        self.context.define_run(self.superrun_name, data=self.subrun_ids)  # Define superrun
```
jorana, line 34: `data` should be a DataFrame or an array right, now it appears to be a list instead. After seeing this I added this other comment about the docs of `define_run`

tests/test_superruns.py
```diff
@@ -0,0 +1,247 @@
...
+                                             'records',
+                                             progress_bar=False)
+        superrun_data = self.context.get_array(self.superrun_name, 'records')
+        assert np.all(subrun_data['time'] == superrun_data['time'])
+
+        # Deactivate the lineage check for subruns:
+        superrun_data = self.context.get_array(self.superrun_name,
+                                               'records',
+                                               _check_lineage_per_run_id=False
+                                               )
```
jorana, line 67: This appears to be a function predating #483 right? At least I think `_check_lineage_per_run_id` shouldn't do anything right?

jorana: There are a few more `_check_lineage_per_run_id` still in this test

tests/test_superruns.py
```diff
@@ -0,0 +1,247 @@
...
+from bson import json_util
+import re
+
+
+class TestSuperRuns(unittest.TestCase):
+
+    def setUp(self, superrun_name='_superrun_test'):
+        self.offset_between_subruns = 10
+        self.superrun_name = superrun_name
+        self.tempdir = tempfile.mkdtemp()
```
jorana, line 20: I assume you use this tempfile since you want to re-use the results and not start making records from scratch. Perhaps this is worth commenting on

tests/test_superruns.py
```diff
@@ -0,0 +1,247 @@
...
+        logger.addFilter(lambda s: not re.match(".*Could not estimate run start and end time.*",
+                                                s.getMessage()))
+
+        self._create_subruns()
+        self.context.define_run(self.superrun_name, data=self.subrun_ids)  # Define superrun
+
+    def test_run_meta_data(self):
+        """
+        Check if superrun has the correct run start/end and livetime.
+        """
```
jorana, line 39: want to add an assert statement that the subruns are stored whereas the superrun is not prior to running the test. It kind of depends on the order of how the other tests are run so if you rename them or change the setup, you might run into issues.

tests/test_superruns.py
```diff
@@ -0,0 +1,247 @@
...
+        
+    def test_rechnunking_and_loading(self):
+        """
+        Tests rechunking and loading of superruns with multiple chunks.
+        
+        The test is required since it was possible to run into race conditions with 
+        chunk.continuity_check in context.get_iter and transform_chunk_to_superrun_chunk 
+        in storage.common.Saver.save_from.
+        """
+        self.context.set_config({'recs_per_chunk': 100000}) # Make chunks > 200 MB
```
jorana, line 158: This can make the rest unnecessarily slow or heavy on I/O, perhaps set the max size in the `st._plugin_class_registry['records'].chunk_target_size_mb =1`

tests/test_superruns.py
```diff
@@ -0,0 +1,247 @@
...
+        
+        rr = self.context.get_array(self.subrun_ids, 'records')
+        endtime = np.max(strax.endtime(rr))
+
+        last_subrun_id = int(self.subrun_ids[-1])+1
```
jorana, line 163: the last subrun_id should be `self.subrun_ids[-1]` right? Especially with the +2 one line below I'm not sure I follow

tests/test_superruns.py
```diff
@@ -0,0 +1,247 @@
...
+        rr_superrun = self.context.get_array('_superrun_test_rechunking', 'records')    
+        rr_subruns = self.context.get_array(self.subrun_ids, 'records')    
+        
+        chunks = [chunk for chunk in self.context.get_iter('_superrun_test_rechunking', 'records')]
+        assert len(chunks) > 1, 'Number of chunks should be larger 1. Has the default chunksize changed?'
+        assert np.all(rr_superrun['time'] == rr_subruns['time'])
+
+    def test_superrun_triggers_subrun_processing(self):
+        """
+        Tests if superrun processing can trigger subrun processing.
```
jorana, line 185:
```suggestion
        Tests if superrun processing can trigger subrun processing. Which it should.
```

docs/source/advanced/superrun.rst
```diff
@@ -3,51 +3,100 @@ Superruns
...
+            '124': [(start, stop), (start, stop), ...],})
 
-From a dictionary of time range tuples. The times must be 64-bit integer UTC timestamps since the unix epoch::
+.. From a dataframe (or record array) with strax data::
 
-    st.define_run('_awesome_superrun', {
-        '123': [(start, stop), (start, stop), ...],
-        '124': [(start, stop), (start, stop), ...],})
+..    st.define_run('_awesome_superrun', events_df)
+    st.define_run('_awesome_superrun', events_df, from_run='123')
```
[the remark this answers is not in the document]
WenzDaniel: Its commented out for the moment as we cannot make superruns from dataframes and/or arrays yet. I am also not sure if you really need it.

tests/test_superruns.py
WenzDaniel: I do not think this is true. setUp and TearDown are called before every single test case so you are not running into these kind of things.

tests/test_superruns.py
WenzDaniel: Ah sorry, yes should have been more explicit here. I create two additional runs here to test the rechunking.

strax/context.py
```diff
@@ -685,21 +688,31 @@ def check_cache(d):
...
                     run_id, 'sub_run_spec')['sub_run_spec']
+
+
+                # Make subruns if they do not exist, since we do not 
+                # want to store data twice in case we store the superrun
+                # we have to deactivate the storage converter mode.
+                stc_mode = self.context_config['storage_converter']
+                self.context_config['storage_converter'] = False
                 self.make(list(sub_run_spec.keys()), d)
+                self.context_config['storage_converter'] = stc_mode
```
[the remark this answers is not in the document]
WenzDaniel: Depends, these lines are actually coming from a previous version in which I had not added yet `write_superruns` as context config. I thought it made sense to keep it in case we ether want to copy only the superrun from one storage frontend to another but not its subruns.

tests/test_superruns.py
jorana: Right, I see. You are right! I was confused for a second with setUpClass and tearDownClass for a second. Not that it hurts to check of course but okay.

This is the behavior I thought you envisaged, so my comment was assuming so 😂.

Hereby a purely academic example of how things would have worked with setUpClass and tearDownClass. This could decrease testing time if e.g. the setup itself takes a lot of time and can be reused.
```python
import tempfile
import os
import unittest
import shutil
import uuid


class TestTheTestingSuite(unittest.TestCase):
    """Test the per-run defaults options of a context"""
    @classmethod
    def setUpClass(cls) -> None:
        temp_folder = uuid.uuid4().hex
        cls.tempdir = os.path.join(tempfile.gettempdir(), temp_folder)
        assert not os.path.exists(cls.tempdir)

    @classmethod
    def tearDownClass(cls):
        shutil.rmtree(cls.tempdir)

    def test_one(self, make='1'):
        os.makedirs(os.path.join(self.tempdir, make))
        print(os.listdir(self.tempdir))

    def test_two(self):
        self.test_one('2')
```

```
tests/test_test.py::TestTheTestingSuite::test_one ['1']
PASSED
tests/test_test.py::TestTheTestingSuite::test_two ['1', '2']
PASSED
```

strax/chunk.py
WenzDaniel: In case if both chunks from run 1 and run 3 are 200 MBs you are right. But I disagree with you. I think it would be better to adjust the chunk boundaries for a superrun chunk to be consistent with the data handling  for regular runs. 

> Wouldn't this be a bit strange and possibly cause funny things later?

What kind of issues do you suspect, maybe I forgot something? Actually, I thought it is saver in this way I implemented it, as I am not sure what happens if you query time_ranges outside of any chunk boundary. 

From a physics point of view it should not matter if there are any gaps in your data. You just have to compute your livetime correctly and therefore I added the subrun field to our chunks as well as the livetime field to the superrun run-metadata.

strax/context.py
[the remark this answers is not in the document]
WenzDaniel: But I also do not mind to change it. I am not even sure if this will be ever a use case.

jorana:
@WenzDaniel maybe one thing I did not think of before, but would superruns allow for post-combining nv + tpc runs? I think /  hope so, it would be extremely useful!

tests/test_superruns.py
WenzDaniel: Ah interesting I was not aware of this, but as you pointed out this requires some additional thinking :D Compared to some of our other tests I the superrun test are fairly fast. Hence, I will address this comment maybe later.

jorana: For sure, it's only 8 seconds on actions. This example is purely academic on this point

strax/run_selection.py
WenzDaniel: I never worked with defaultdicts, but I think it does not really help here.  I would need different values for start and end because of,
```
run_md['start'] = min(run_md['start'], run_doc_start)
run_md['end'] = max(run_md['end'], run_doc_end)
```
would not I? Hence I do not see anything to gain.

WenzDaniel:
Uff hard question. I have to admit I do know. I never planed on creating superruns based on in time overlapping runs. But there are some additional challenges beside the technical once. E.g. the time alignment between the different detectors which is different for each subrun.

strax/storage/common.py
WenzDaniel: Lets keep discussing in the thread above.

strax/chunk.py
WenzDaniel: > I might now see why you made this transform_chunk_to_superrun_chunk function (see above). But would the issue you are trying to solve not much easier be addressed in continuity_check?

I also thought about that, but see my comment above.

strax/chunk.py
```diff
@@ -233,11 +236,35 @@ def concatenate(cls, chunks):
         data_type = data_types[0]
 
         run_ids = [c.run_id for c in chunks]
+        is_superrun = np.all([r.startswith('_') and not r.startswith('_temp_') for r in run_ids])
```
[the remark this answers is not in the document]
WenzDaniel: > r in run_id is not going to be a list always so looping over it might not always work.

Hmm maybe it was late yesterday or I do not understand what you mean :D. It should always be a list see line above. 

> Also, can't you do simply is_superrun = subruns is not None

Yes you are right I will change it. Subruns just came later into the PR.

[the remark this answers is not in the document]
jorana: Ah you are also right, so easy to mix up `run_id` and `run_ids`.

strax/chunk.py
jorana: Chunking issues can be nasty. E.g. if you were to combine super-runs into super-runs by adding `run2` to `r1+3`, you would have overlapping chunks.

>I think it would be better to adjust the chunk boundaries for a superrun chunk to be consistent with the data handling for regular runs.

But this feels very patchy right? I think just adding a line in the continuity check to see if the subrun is different from one chunk to another would be more robust. Also you wouldn't then have to worry about how you compute the live-time, you would get it correctly either from looking at the chunks or the docs. Also I think it's very strange to have chunks of a subrun extending well beyond the boundaries of that run and can for sure lead to confusion/errors later.

Is there any compelling argument why a superrun should have continuous chunks even tough we know that for a fact, there is no reason to assume the chunks are continuous from run to run. If it is only the continuity check which motivates this?

>In case if both chunks from run 1 and run 3 are 200 MBs you are right.

Sorry, I don't follow this point?

strax/run_selection.py
jorana: Right it would only be for the livetime where you'd actually be using it's behavior. There is nothing wrong with the current implementation 😉

strax/chunk.py
```diff
@@ -233,11 +236,35 @@ def concatenate(cls, chunks):
...
             raise ValueError(
                 f"Cannot concatenate {data_type} chunks with "
                 f"different run ids: {run_ids}")
+
         run_id = run_ids[0]
+        
+        subruns=None
+        if is_superrun:
+            start_times = [c.start for c in chunks]
+            sorted_ind = np.argsort(start_times)
```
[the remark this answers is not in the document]
WenzDaniel: I removed the sorting. A check is not needed as we have already one at the end of concatenate which gets raised in case of unsorted data.

WenzDaniel:
I addressed all outstanding comments. I will have a last look tomorrow morning with a fresh pair of eyes. After that we can merge.

WenzDaniel:
Okay I am happy to merge if there are no other comments.

jorana:
Nice! Thanks Daniel, for bookkeeping, can you mention what happens if we query in between subruns?

strax/context.py
```diff
@@ -685,21 +688,30 @@ def check_cache(d):
...
                 sub_run_spec = self.run_metadata(
                     run_id, 'sub_run_spec')['sub_run_spec']
+
+                # Make subruns if they do not exist, since we do not 
+                # want to store data twice in case we store the superrun
+                # we have to deactivate the storage converter mode.
+                stc_mode = self.context_config['storage_converter']
+                self.context_config['storage_converter'] = False
                 self.make(list(sub_run_spec.keys()), d)
+                self.context_config['storage_converter'] = stc_mode
```
jorana, line 706: Did you not wanted to remove this?

strax/context.py
```diff
@@ -747,10 +759,15 @@ def concat_loader(*args, **kwargs):
                 to_compute[d] = p
                 for dep_d in p.depends_on:
                     check_cache(dep_d)
-
+
```
jorana, line 762:
```suggestion

```

strax/processor.py
```diff
@@ -51,7 +51,8 @@ def __init__(self,
                  allow_lazy=True,
                  max_workers=None,
                  max_messages=4,
-                 timeout=60):
+                 timeout=60,
+                 is_superrun=False,):
```
jorana, line 55:
```suggestion
                 is_superrun=False,
                 ):
```

jorana (approved):
Thanks Daniel for the changes

jorana:
Feel free to merge 👍

PR #473, opened 2021-06-18
Re ordered time field in cut plugins

WenzDaniel:
Trivial change, I just re-aranged the time field position in our CutPlugin.  In an array of merged plugins the result will look like
```
False, 1621861432156179212, 1621861432159096710,  True, False,  True,  True, False
```
which drove me crazy.


jorana:
sure 😉

PR #476, opened 2021-06-29
Change default progress_bar=False for get_iter

ershockley:
Proposal for very simple change that will be less verbose. I think most users want to load multiple runs at a time anyway and so we automatically get a tqdm progress bar there, not to mention the fact that the main use-case is just loading data, not creating it.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -894,7 +894,7 @@ class Context:
                  selection_str=None,
                  keep_columns=None,
                  allow_multiple=False,
-                 progress_bar=True,
+                 progress_bar=False,
                  _chunk_number=None,
                  **kwargs) -> ty.Iterator[strax.Chunk]:
         """Compute target for run_id and iterate over results.


kdund (approved):

jorana:
@ershockley , sure, indeed with the many runs simultaneously it doesn't make much sense. However, for single runs, it should be fine right?

As such, how about instead adding the following line here:

```python
kwargs.setdefault('progress_bar', False)
```
https://github.com/AxFoundation/strax/blob/ershockley-patch-1/strax/context.py#L1133

ershockley:
In those cases then you could just pass `progress_bar=True`, right? I feel like one should opt for more verbosity. 

Anyway what you suggest is still an improvement.

WenzDaniel:
I prefer Jorans idea. The progress bar should be seen as a service the user does not have to take care of. If we set it to false by default we can also simplify the code and just remove the lines since (almost) no analysts will ever set it to true.

jorana:
closing in favour of #480

PR #480, opened 2021-07-07
change default pbar behavior (for multiple runs)

jorana:
As discussed in https://github.com/AxFoundation/strax/pull/476, change the default behavior for loading multiple runs since having two competing pbars is messy.

Also make the runs-pbar configurable.

Rather than disabling the pbar altogether (as in  #476) only do it for multiple runs in the multi-run code.


jorana:
🤷‍♂️ codefactor, I did not even touch those lines?

WenzDaniel (approved):
Looks good.

PR #482, opened 2021-07-08
Seperate waveforms

petergaemers:
This pr add the needed stuff to have seperate top waveforms stored in peaks
sum_waveform (and some others) are modified slightly to not only sum the full waveform but also include the waveform from the top array

  - Update the docstring(s)
  - Update the documentation


WenzDaniel:
@petergaemers thanks for your PR. There are seem to be some tests which are failing due to missing arguments in `split_peaks() missing 2 required positional arguments: 'n_top_pmts' and 'store_top_waveform'`. Since the changes are in your private repo could you updated the test then I review your changes.

strax/dtypes.py
```diff
@@ -194,6 +194,8 @@ def peak_dtype(n_channels=100, n_sum_wv_samples=200, n_widths=11):
           'n_hits'), np.int32),
         (('Waveform data in PE/sample (not PE/ns!)',
           'data'), np.float32, n_sum_wv_samples),
+        (('Waveform data in PE/sample of the top array(not PE/ns!)',
+          'data_top'), np.float32, n_sum_wv_samples),
```
jorana, line 198: @petergaemers could you look into reducing this precision to reduce the size as per Jelles suggestions as well as making this an optional field? Otherwise making it optional in the processing has little effect

jorana (changes requested):
Thanks Peter for reopening, apart from Daniels comment, could you also check this:

strax/dtypes.py
terliuk: I thought that we agreed to start it from the full precision , as we will start with small number of runs and creation of top/bottom waveforms will be disabled by default. Other issue with reduced precision is the fact that we have undershoots to negative values etc, what can significantly affect peaks.

jorana: > as well as making this an optional field

This at least should be easy to implement

>I thought that we agreed to start it from the full precision 

I don't remember converging to this convention. Also with this full precision you can have negative areas? But okay if you looked into it and it's not possible get this working, please go ahead

WenzDaniel:
FYI: Given the changes in https://github.com/AxFoundation/strax/pull/486 we cannot merge this PR anymore. I will close it an open a new one with the desired changes, but based on https://github.com/AxFoundation/strax/pull/486.

jorana:
I propose to keep this PR open as a reminder that this should be added (very) soon.

JoranAngevaare:
> 
> 
> FYI: Given the changes in #486 we cannot merge this PR anymore. I will close it an open a new one with the desired changes, but based on #486. I apologize for any inconveniences.

@WenzDaniel , did you intent to include this in the upcoming release?

JoranAngevaare:
replaced by #565

PR #483, opened 2021-07-09
Fix lineage if per run default is not allowed

jorana:

- Massive speedup for loading many runs
- When `use_per_run_defaults == False` we don't allow for changing lineages in the context. If so, there is no need to for every run recompute the lineage of each plugin
- This is a similar PR to https://github.com/AxFoundation/strax/pull/440.

**This only works when we have fixed lineages**, but when they are fixed, we don't need a full initialization of the plugin when we are going to load a new run. We can just re-use a plugin and only change run run_id attribute. To this end:
 - Initialize the plugin once
 - Cache the plugin in the _fixed_plugin_cache. To prevent a changed config not being reflected in the plugin (e.g. you load data and later change an option), we make a sha1 of the `st.config` any changes there will result in a new hash, and as such a new key to cache the plugins under.
 - Every next time a datakey/processor is requested, change return the requested plugin from the _fixed_plugin_cache and only change the run_id

Before 
afbeelding

After
afbeelding

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -974,7 +974,7 @@ class Context:
             if len(set(plugins[d].data_kind_for(d) for d in targets)) == 1:
                 temp_name = ('_temp_'
                              + ''.join(
-                            random.choices(string.ascii_lowercase, k=10)))
+                                random.choices(string.ascii_lowercase, k=10)))
                 p = type(temp_name,
                          (strax.MergeOnlyPlugin,),
                          dict(depends_on=tuple(targets)))
@@ -1004,14 +1004,14 @@ class Context:
 
         seen_a_chunk = False
         generator = strax.ThreadedMailboxProcessor(
-            components,
-            max_workers=max_workers,
-            allow_shm=self.context_config['allow_shm'],
-            allow_multiprocess=self.context_config['allow_multiprocess'],
-            allow_rechunk=self.context_config['allow_rechunk'],
-            allow_lazy=self.context_config['allow_lazy'],
-            max_messages=self.context_config['max_messages'],
-            timeout=self.context_config['timeout']).iter()
+                components,
+                max_workers=max_workers,
+                allow_shm=self.context_config['allow_shm'],
+                allow_multiprocess=self.context_config['allow_multiprocess'],
+                allow_rechunk=self.context_config['allow_rechunk'],
+                allow_lazy=self.context_config['allow_lazy'],
+                max_messages=self.context_config['max_messages'],
+                timeout=self.context_config['timeout']).iter()
 
         try:
             _p, t_start, t_end = self._make_progress_bar(run_id,
@@ -1101,10 +1101,10 @@ class Context:
             pbar.n = 0
         # Let's add the postfix which is the info behind the tqdm marker
         seconds_per_chunk = time.time() - pbar.last_print_t
-        pbar.mbs.append((nbytes / 1e6) / seconds_per_chunk)
+        pbar.mbs.append((nbytes/1e6)/seconds_per_chunk)
         mbs = np.mean(pbar.mbs)
         if mbs < 1:
-            rate = f'{mbs * 1000:.1f} kB/s'
+            rate = f'{mbs*1000:.1f} kB/s'
         else:
             rate = f'{mbs:.1f} MB/s'
         postfix = f'#{n_chunks} ({seconds_per_chunk:.2f} s). {rate}'
@@ -1295,8 +1295,8 @@ class Context:
             else:
                 # Function returned a scalar or flat array
                 result['result'] = (
-                        np.sum(data, axis=0)
-                        + result.get('result', 0))
+                    np.sum(data, axis=0)
+                    + result.get('result', 0))
 
             n_chunks += 1
 
@@ -1512,7 +1512,7 @@ class Context:
 
         if target_compressor is not None:
             self.log.info(f'Changing compressor from {md["compressor"]} '
-                          f'to {target_compressor}.')
+                           f'to {target_compressor}.')
             md.update({'compressor': target_compressor})
 
         for t_sf in target_sf:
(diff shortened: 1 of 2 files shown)


strax/context.py
```diff
@@ -415,12 +416,56 @@ def _set_plugin_config(self, p, run_id, tolerant=True):
...
                     p.config[parent_name] = option_value
 
+    def _config_hash(self):
+        """
+        Dump the current config to a hash as a sanity check for building
+        the _fixed_plugin_cache. If any item changes in the config, so
+        does this hash.
+        """
+        return hashlib.sha1(str(self.config).encode('ascii')).hexdigest()
+        # return strax.deterministic_hash(self.config)
```
jorana, line 432: I tried using the deterministic hash but it does not like immutibledict. Perhaps there is an other solution but my guess is that this serves the purpose quite well.

jmosbacher: Unfortunately this is not deterministic, you need to sort any dictionaries and sets to get a deterministic string. Also the string representation may depend on the version of the package of non-bultin values. Why not just fix the hashablize function [here](https://github.com/AxFoundation/strax/blob/ba3eb8e9e3e18f7a9b4a2a855c107478c05f5e2b/strax/utils.py#L254)
```python
from collections.abc import Mapping

254        if isinstance(obj, Mapping): # instead of if isinstance(obj, dict)
...
```

jorana: Thanks Yossi, sure, let me try that. 

Actually, it's not much of an issue if it's non-deterministic, the contexts will always just build the `_config_hash` on the fly, so any changes in order of the config or the version of a package may lead to a different context_hash but as long as it does not change for every run, it shouldn't matter. But you are right, better do make it deterministic because even if there is no issue now because some day it might if one assumes it is deterministic (which shouldn't be a bad assumption).

strax/context.py
```diff
@@ -415,12 +416,56 @@ def _set_plugin_config(self, p, run_id, tolerant=True):
...
+                return True
+
+        return False
+
+    def _plugins_to_cache(self, plugins: dict) -> None:
+        if self.context_config['use_per_run_defaults']:
+            # There is no point in caching if plugins (lineage) can change per run
+            return
+        if self._fixed_plugin_cache is None or self._config_hash() not in self._fixed_plugin_cache:
+            self._fixed_plugin_cache = {self._config_hash(): dict()}
```
WenzDaniel, line 454: I assume Yossi's comment is also the reason why you are always making a new dictionary when the hash cannot be found?

WenzDaniel (approved):
I have not tested it, but from the code it looks fine. I am not sure if the deterministic hashing is really needed. I assume that changing options is only done by very few analysts. But if it can be implemented easily it would be a nice to have.

strax/context.py
jorana: Very keenly spotted Daniel, I should have put a comment here.

So my suspicion is that the cache is only once created, at most twice if you change some option or register some plugin. I did think the likelihood that one was eating away memory for keeping the cache was greater than the change that someone was flipping between options often (although neither is very likely).

WenzDaniel: > So my suspicion is that the cache is only once created, at most twice if you change some option or register some plugin. 

Yes, this would be my guess, too. As I said I am fine like it is but if you can implement Yossi's suggestion easily it would be a nice to have,

jorana: I fully agree

jmosbacher:
In general I think this is a good idea. My biggest concern is with multi-threading. Do we really have a guarantee that two runs are not being processed concurrently by two different threads?

WenzDaniel:
I think so. The list of plugins needed to process some data is created before we send the different jobs. 
.

strax/context.py
jorana: @jmosbacher , I tried your suggestion but it doesn't work. I initially also tried something like this (by doing the the same for `immutabledict` instead of `Mapping`), however, it turns out that since immutabledict has a hash method, this whole logic is never accessed. 

I'll propose something similar to get it working.

jorana:
@jmosbacher @WenzDaniel , thanks for the reviews, I've added three things:

- 93e2993, convert Mapping to dict in order to allow using `immutabledict` for the deterministic hash
- cae95cf , rely on the deterministic hash instead of a `sha1`. This does make it ~5 times slower without an immediate benefit for the purpose pursued here. Let me slightly move it around here to maximize performance.
- cae95cf , we *must* also check the versions of the plugins, I had initially forgotten this.

jorana:
> 
> 
> In general I think this is a good idea. My biggest concern is with multi-threading. Do we really have a guarantee that two runs are not being processed concurrently by two different threads?

For completeness, this is all setup much later after this function https://github.com/AxFoundation/strax/blob/a7d7c11e2b6e6b7213c7ecc05b71fefb75c5baa2/strax/context.py#L647. In fact during multi-run loading/processing you loop over the components here:
https://github.com/AxFoundation/strax/blob/a7d7c11e2b6e6b7213c7ecc05b71fefb75c5baa2/strax/utils.py#L435

strax/context.py
jmosbacher: Oh, sorry about that. I assumed the issue was with the hash function but I guess it was in the json encoder all along? In that case maybe add the Mapping logic to the NumpyJSONEncoder class?

jmosbacher:
@jorana Sorry, maybe im misunderstanding whats happening in your code. Arent you caching the plugin instances themselves so different runs will be processed with the same plugin instance? If I try to process multiple runs by default those will be processed concurrently with a thread pool right? but if the plugins are cached all threads would share the same plugin instance and the run_id on that instance would just be the last one requested no? what am I missing here?

jorana:
Ah right, I see your point better now. What you are saying is correct thanks for being persistent. Let's write a test for this as indeed the implications can be substantial.

I went ahead with merging this PR as I wanted to get things going for Daniels PR on super-runs but would have been better to check this carefully first.

jmosbacher:
I think a possible solution to this would be to return a copy of the cached plugin instead of the original instance. Also would need a recursive copy of the plugin.deps variable I guess.

PR #485, opened 2021-07-13
Copy plugin instance for processing

jorana:
So although #483 seemed nice, @jmosbacher  correctly pointed out that for multi-threading multiple runs, this can cause issues. In fact with the `.deps` tracking in the plugins, we were ever more obscurely shipping configs from older runs into the processing.

I should have written a test for it to catch these issues. So that is added in this PR.

## Design choice: reliability over performance
For the copy function several things could be copied that are setup during the init of a plugin:
 - `Plugin.deps` (the dependencies of the plugin that point to lower-level plugins)
 - `Plugin.infer_dtype()` results
 - `Plugin.setup()` results
 - `Plugin.config` (set in `Context._set_plugin_config`
 - `Plugin.lineage`

In the current implementation we went for reliability over performance. This means we only copy the `Plugin.config` and `plugin.lineage`. Below we discuss why the others are not copied:
 - `Plugin.setup()`. Since `Plugin.setup()` sets up things like [`Plugin.to_pe` ](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/peaklet_processing.py#L110), those results should definitely **not** be copied. We could rely on the user to make sure to write the setup function in such a way that running it a second time just overwrites attributes like `Plugin.to_pe` but I can come up with implementations where you start re-using such attributes so I would rather not have to rely on the user not making an unintentional mistake. 
 - `Plugin.deps` : since this is where we keep a copy of the instances that are responsible for processing lower level data, these copies have the same issue as the lineage from above. Therefore, also these need be rebuild for every new run if we copy a plugin from the cache.
 - `Plugin.infer_dtype()` results depend often on `Plugin.deps`. There are a few plugins (e.g. [this one](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/double_scatter.py#L72)) where class attributes are set during the `infer_dtype` which are then used in the compute. Maybe this is sloppy and we should move this to `setup` of those plugins, but I would rather just run `Plugin.infer_dtype()` again. 
 
#### Note on performance
If we would copy the `Plugin.infer_dtype()`-results as well and not use `Context._fix_dependency`, we could make `Context._get_plugins` ~15 times faster since it again does a recursive init. However, I don't mind if we spend a little more time here (still faster then without caching) to be safe rather than sorry for the reasons explained above. Especially since `Context._get_plugins` is now not per-se called for `Context.key_for` which is the driving function for loading data. Creating data takes time so spending ~O(30 ms) to start up a plugin or ~O(2 ms) is not going to make a very noticeable difference. The speed increase of `Context.key_for` (~750x) will show up during loading of data.

So now we instead make copies such that each tread in multiprocessing can be handled by a separate instance.

Also, for `Context.key_for` we *can* get the lineage from the cache without having to recompute anything (again if no run-defaults are allowed). So now the key-generation will be much faster, this will also make checking if something is stored much faster,

## Can you give a minimal working example (or illustrate with a figure)?
The performance increase in getting a single plugin is nice still:
afbeelding

The biggest performance increase is for the key-generation which we use on data-loading. This will greatly affect things like e.g. cuts:
afbeelding

### DIFF
--- a/tests/test_fixed_plugin_cache.py
+++ b/tests/test_fixed_plugin_cache.py
@@ -1,0 +1,113 @@
+from strax.testutils import Records, Peaks
+import strax
+import unittest
+import numpy as np
+
+
+class ChannelIsRunidRecords(Records):
+    """Set the channel field equal to the run_id"""
+    def compute(self, chunk_i):
+        res = super().compute(chunk_i)
+        res.data['channel'][:] = int(self.run_id)
+        return res
+
+
+class MaxChannelPeaks(Peaks):
+    def infer_dtype(self):
+        # We are going to check later that the infer_dtype is always called.
+        dtype = strax.peak_dtype() + [
+            (('PMT with median most records',
+              'max_pmt'), np.int16)
+        ]
+        self.dtype_is_set = True
+        return dtype
+
+    def compute(self, records):
+        assert np.all(records['channel'] == int(self.run_id))
+        res = super().compute(records)
+        res['max_pmt'] = records['channel'].mean()
+        return res
+
+
+class TestContextFixedPluginCache(unittest.TestCase):
+    """Test the _fixed_plugin_cache of a context"""
+    def test_load_runs(self, n_runs=3, config_update=None, **kwargs):
+        """Try loading data for n_runs to make sure that we are """
+        run_ids = [str(r) for r in range(n_runs)]
+        st = self.get_context(use_per_run_defaults=False)
+        if config_update is not None:
+            st.set_context_config(config_update)
+        data = st.get_array(run_ids, 'records', **kwargs)
+        run_id_channel_diff = data['run_id'].astype(np.int64) - data['channel']
+        assert np.all(run_id_channel_diff == 0)
+
+        # To be sure also double check Peaks as self.deps of the Plugin
+        # class should be correctly taken care of by the context.
+        peaks_data = st.get_array(run_ids, 'peaks')
+        run_id_max_pmt_diff = peaks_data['max_pmt'] - peaks_data['run_id'].astype(np.int64)
+        assert np.all(run_id_max_pmt_diff == 0)
+
+    def test_get_plugin(self, n_runs=3):
+        run_ids = [str(r) for r in range(n_runs)]
+        st = self.get_context(use_per_run_defaults=False)
+        plugins_seen = []
+        for run in run_ids:
+            p = st.get_single_plugin(run, 'records')
+            plugins_seen.append(p)
+            assert p.run_id == run
+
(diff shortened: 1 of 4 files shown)


jmosbacher:
Im a bit confused about the implementation of the copy function. Doesnt it defeat the purpose if it only copies the variables set by the context? i thought the purpose of the caching was to capture the side effects of the user defined setup method that can be slow due to loading configs from the DB etc.?

jorana:
> 
> 
> Im a bit confused about the implementation of the copy function. Doesnt it defeat the purpose if it only copies the variables set by the context? i thought the purpose of the caching was to capture the side effects of the user defined setup method that can be slow due to loading configs from the DB etc.?

Hi Yossi, thanks, that was not the purpose, I updated the description of the PR to explain the choices I made here (and we had a chat :wink: ) I hope this makes sense now.

jmosbacher (approved):
Looks good to me. Nice work.

tests/test_fixed_plugin_cache.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def test_load_runs_multicore(self):
+        """Load the runs. If the references are mixed up the results are inconsistent"""
+        multicore_config = dict(allow_lazy=False,
+                                timeout=60,
+                                allow_multiprocess=True,
+                                )
+        self.test_load_runs(n_runs=10, config_update=multicore_config, max_workers=10)
+
+    def test_cache_changes(self):
```
WenzDaniel, line 72: Do you think it would be needed to have another assert which tests if the changes are propagated to the data itself? So far you checked only for the run_id is not it?

WenzDaniel (approved):
Thank you two for this change and careful check. May I ask for a small addition. Given that this change has become a bit more complicated than the initial idea, I think it would be worth to have a context config which allows to explicitly disable this short cut. This could be handy for debugging if we ever encounter some.

tests/test_fixed_plugin_cache.py
jorana: @WenzDaniel , yes it's worth checking. In fact it is done here:
https://github.com/AxFoundation/strax/blob/e1917108f51c9517e28d3040161ad8ce2b7ea675/tests/test_fixed_plugin_cache.py#L41

To this end, I set the plugin.run_id to the 'channel' field in records:
https://github.com/AxFoundation/strax/blob/e1917108f51c9517e28d3040161ad8ce2b7ea675/tests/test_fixed_plugin_cache.py#L11

jorana:
>Thank you two for this change and careful check. May I ask for a small addition. 

I think for debugging purposes we can just use 'user_per_run_defaults' right? That disables all this (as in my example in the description)

PR #486, opened 2021-07-14
Fixing peaklet baseline bias

WenzDaniel:
Our current treatment of the floating point part of our baseline leads to an area (and summed waveform) bias of several 10ish % in peaklets. The main reason for the positive bias is coming form the fact, that we are also adding the floating point part of the baseline in zero-padded regions. A more detailed note can be found [here](https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:wenz:comissioning:tpc:sr0:s1_bias).

The solution is rather simple: We make our sum_waveform hit dependent and only build peaklets over the regions in which we found hits before. 

In terms of performance is the new algorithm ~ x1.5 slower compared to the old one.


JelleAalbers:
Hi Daniel, Great work spotting this! I'm amazed such a large area bias could have slipped our validation tests on 1T data. CES peak positions below saturation territory agreed between pax and strax (e.g. xenon:xenonnt:analysis:strax_clustering_classification#s1_s2_events), but maybe that's driven mostly by the S2s? The 1PE peak also seemed about right in https://github.com/AxFoundation/strax/pull/261, at least if the gain calibration includes underamplified / sub-threshold hits.

With the 'zero-padded part' of the waveform, you mean the parts zeroed out by [cut_outside_hits](https://github.com/AxFoundation/strax/blob/master/strax/processing/data_reduction.py#L50), right? There is also a part of the record data field beyond `record['length']` (for the final `record_i` in a pulse), but that shouldn't do anything; no routine should even look at it.

Still, sorry.. I should have realized we can't zero an integer-valued waveform if the true zero level is a floating value. Ultimately I guess this is payback for not copying the record data into the hit datatype, as we did in pax, or as in your hitlets..

If the hit-dependent sum waveform somehow doesn't work out, an alternative might be to let `cut_outside_hits` assign not zero, but a special value way outside the dynamic range of the digitizer (e.g. define `TRUE_ZERO = -32768`), and have  `sum_waveform` ignore samples with that value. To avoid reprocessing records, you could run the modified cut_outside_hits again after hitfinding in peaklets (it should be idempotent). 

But that's just a random idea. I think your solution is to essentially  merge `cut_outside_hits` with `sum_waveform`; that should certainly work too. It would make `cut_outside_hits` in `PulseProcessing` really only a disk space saver.

WenzDaniel:
Hi Jelle, good to hear from you. I hope you are doing fine in your new position. Thanks for your comment. 

> With the 'zero-padded part' of the waveform, you mean the parts zeroed out by cut_outside_hits, right?

Yes, you are right. 

> I'm amazed such a large area bias could have slipped our validation tests on 1T data. 

We were also quite surprised 

> but maybe that's driven mostly by the S2s?

Small few electrons S2s are affected most indent as the number of padded samples are largest. But we also found a bias of about 10 % for S1 and 20 % for S2s with our waveform simulator. 

> If the hit-dependent sum waveform somehow doesn't work out, an alternative might be to let cut_outside_hits assign not zero, but a special value way outside the dynamic range of the digitizer (e.g. define TRUE_ZERO = -32768), and have sum_waveform ignore samples with that value. To avoid reprocessing records, you could run the modified cut_outside_hits again after hitfinding in peaklets (it should be idempotent).

Not sure how this affects the data size after compression, but would not this defeat the entire purpose of records?  The most simplistic solution would be to just skip the records step, but I pretty sure Evan would not approve :-D.

> I think your solution is to essentially merge cut_outside_hits with sum_waveform; that should certainly work too. 

Yes, I hope so.  I will run some additional test before finally merging.

strax/processing/peak_building.py
```diff
@@ -197,55 +203,71 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
...
+                r = records[prev_record_i[record_i]]
+                is_saturated |= _build_hit_waveform(h, r, hit_waveform)
 
-            swv_buffer[p_start:p_end] += pe_waveform
+            if h['right_integration'] > n_samples_record and next_record_i[record_i] != -1:
+                r = records[next_record_i[record_i]]
+                is_saturated |= _build_hit_waveform(h, r, hit_waveform)
 
-            area_pe = pe_waveform.sum()
+            p['saturated_channel'][ch] = is_saturated
```
zhut19, line 264: Here should be `|=` as well

JelleAalbers:
Thanks Daniel, I'm doing well; just checking in here occasionally out of curiosity ;-) 

If the compressor is smart it should recognize TRUE_ZERO is common and assign it a short encoding, as it does for 0 now. But again, never mind if your solution is already working well.

strax/processing/peak_building.py
```diff
@@ -154,10 +154,15 @@ def store_downsampled_waveform(p, wv_buffer):
 
 @export
 @numba.jit(nopython=True, nogil=True, cache=True)
-def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
+def sum_waveform(peaks, hits, records, record_links, adc_to_pe, select_peaks_indices=None):
     """Compute sum waveforms for all peaks in peaks
     Will downsample sum waveforms if they do not fit in per-peak buffer
 
+    :param peaks: Peaks for which the summed waveform should be build.
+    :param hits: Hits which are inside peaks. Must be sorted according
```
zhut19, line 162: I guess this should be updated

strax/processing/peak_building.py
```diff
@@ -197,55 +203,69 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
...
-            (r_start, r_end), (p_start, p_end) = strax.overlap_indices(
-                r['time'] // dt, n_r,
-                p['time'] // dt, n_p)
+            # Get overlapping samples between hit and peak:
+            (h_start, h_end), (p_start, p_end) = strax.overlap_indices(
+                h['time'] // dt, n_samples_hit,
+                p['time'] // dt, n_samples_peak)
+
+            # Create hit waveform
+            hit_waveform = np.zeros(n_samples_hit)
```
zhut19, line 245: Probably wouldn't matter, but can you use a buffer instead of creating it for each hit?

strax/processing/peak_building.py
```diff
@@ -255,6 +275,38 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
...
+    Updates hit_waveform inplace. Result is still in ADC counts.
+
+    :returns: Boolean if record saturated within the hit.
+    """
+    (h_start_record, h_end_record), (r_start, r_end) = strax.overlap_indices(
+        hit['time'] // hit['dt'], hit['length'],
+        record['time'] // record['dt'], record['length'])
+    
+    if not (r_end - r_start):
+        print(r_start, r_end)
```
zhut19, line 291: We don't need the prints anymore, right?

WenzDaniel:
Okay I added Tianyu's comments. I run also all tests in straxen with https://github.com/XENONnT/straxen/pull/601 and all tests pass. We should not merge this PR without merging https://github.com/XENONnT/straxen/pull/601. Waiting for Jorans review now.

zhut19 (approved):

strax/processing/peak_building.py
```diff
@@ -154,10 +154,17 @@ def store_downsampled_waveform(p, wv_buffer):
...
+    waveform other regions in which hits were found. This is required
+    to avoid any bias due to zero-padding and baselining.
     Will downsample sum waveforms if they do not fit in per-peak buffer
 
+    :param peaks: Peaks for which the summed waveform should be build.
+    :param hits: Hits which are inside peaks. Must be sorted according
+        to record_i.
+    :param records: Records to be used to build peaks.
+    :param record_links: Tuple of previous and next records.
     :arg select_peaks_indices: Indices of the peaks for partial
```
jorana, line 168: maybe replace :arg with :param while now at it

strax/processing/peak_building.py
```diff
@@ -197,55 +207,68 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
...
+            (h_start, h_end), (p_start, p_end) = strax.overlap_indices(
+                h['time'] // dt, n_samples_hit,
+                p['time'] // dt, n_samples_peak)
+
+            hit_waveform[:] = 0
+
+            # Get record which belongs to main part of hit (wo integration bounds):
+            r = records[record_i]
+
+            is_saturated = _build_hit_waveform(h, r, hit_waveform)
```
jorana, line 253: for clean code I would separate `_build_hit_waveform` and `_is_saturated` as one changes the state and the other returns something. It's not very important but just not entirely intuitive if _build_hit_waveform returns a bool for saturation 😉

strax/processing/peak_building.py
```diff
@@ -255,6 +278,31 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
...
+    """
+    (h_start_record, h_end_record), (r_start, r_end) = strax.overlap_indices(
+        hit['time'] // hit['dt'], hit['length'],
+        record['time'] // record['dt'], record['length'])
+
+    # Get record properties:
+    record_data = record['data'][r_start:r_end]
+    multiplier = 2**record['amplitude_bit_shift']
+    bl_fpart = record['baseline'] % 1
+    # TODO: check numba does casting correctly here!
```
jorana, line 297: Did you check? If so better not push TODOs that no one will do later

strax/processing/peak_building.py
```diff
@@ -255,6 +278,31 @@ def sum_waveform(peaks, records, adc_to_pe, select_peaks_indices=None):
...
+    :returns: Boolean if record saturated within the hit.
+    """
+    (h_start_record, h_end_record), (r_start, r_end) = strax.overlap_indices(
+        hit['time'] // hit['dt'], hit['length'],
+        record['time'] // record['dt'], record['length'])
+
+    # Get record properties:
+    record_data = record['data'][r_start:r_end]
+    multiplier = 2**record['amplitude_bit_shift']
+    bl_fpart = record['baseline'] % 1
```
jorana, line 296: I was wondering if we could re-use https://github.com/AxFoundation/strax/blob/e1b77aa7dd9b951dc14f0066f3e4e900ff4833ab/strax/processing/pulse_processing.py#L110. But it seems not. However, it did make me realize that according to this definition, the 'area' of a record is not really going to make much sense as we are also adding the floating part of the BL there for the entire length of the record.

Not very much a comment to your PR, just a realization.

strax/processing/peak_building.py
```diff
@@ -292,13 +340,26 @@ def find_peak_groups(peaks, gap_threshold,
...
-
+@export
 @numba.njit(nogil=True, cache=True)
-def _find_hit_integration_bounds(
-        lone_hits, peaks, records, save_outside_hits, n_channels):
-    """"Update lone hits to include integration bounds
-
-    save_outside_hits: in ns!!
+def find_hit_integration_bounds(
+        lone_hits, peaks, records, save_outside_hits, n_channels,
```
jorana, line 346: perhaps refactor `lone_hits` and `peaks` to `hits` and `excluded_intervals`

jorana (approved):
Thanks Daniel, the changes look good!

Just a few minor comments from my side

strax/processing/peak_building.py
WenzDaniel: Yes, I agree it is not nice at all. But I am lacking an idea how it can be separated nicely without doing things twice. I will think a little bit more about it maybe I find a solution.

strax/processing/peak_building.py
```diff
@@ -330,12 +391,18 @@ def _find_hit_integration_bounds(
         # Ensure we do not integrate parts of peaks
         # or (at least for now) beyond the record in which the hit was found
         r = records[h['record_i']]
```
jorana, line 393: maybe also remove the comment above this line

strax/processing/peak_building.py
WenzDaniel: Okay no,  I think splitting things up would make it only more complicated and result in more duplicated code.

PR #488, opened 2021-07-15
Fix #472

jorana:
Fix indentation after #472

Little nuance, see:
afbeelding

### DIFF
--- a/strax/run_selection.py
+++ b/strax/run_selection.py
@@ -107,7 +107,8 @@ def scan_runs(self: strax.Context,
                 if not _is_superrun:
                     # If there is no name, make one from the number
                     doc['number'] = int(doc['name'])
-            doc.setdefault('name', f"{doc['number']:06d}")
+            if not _is_superrun:
+                doc.setdefault('name', f"{doc['number']:06d}")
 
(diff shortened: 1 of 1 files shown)


strax/run_selection.py
```diff
@@ -99,15 +99,16 @@ def scan_runs(self: strax.Context,
...
                 if 'name' not in doc:
                     raise ValueError(f"Invalid run doc {doc}, contains "
                                      f"neither name nor number.")
-                _is_superrun = doc['name'].startswith('_')
                 if not _is_superrun:
-                    doc['number'] = int(doc['name'])
                     # If there is no name, make one from the number
-                    doc.setdefault('name', f"{doc['number']:06d}")
+                    doc['number'] = int(doc['name'])
+            if not _is_superrun:
```
jorana, line 110: This should not even be strictly necessary right because a super-run always has a name

WenzDaniel: > doc.setdefault('name', f"{doc['number']:06d}")

Was not touched by me this was there before.

WenzDaniel (approved):
Thanks for fixing.

PR #491, opened 2021-07-18
Fix nveto processing

WenzDaniel:
During my investigation for [straxen #596](https://github.com/XENONnT/straxen/issues/596#issuecomment-881949739) I noticed that in rare cases hitlet splitting produces zero length rows if new peaks are not sorted by time. Given the nature of records and hitlets compared to peaks this can happen from time to time. In this PR we add an explicit sorting and raise condition if zero-length peaks are supposed to be returned after splitting as this is not a desired behavior. 

Unfortunately, I do not belief that this change will solve #596.


jorana:
Thanks Daniel,
Would it for explicitness not be better to do this step in the plugin computation in straxen? I think that might be clearer than making this additional step in the peak splitting routine?

WenzDaniel:
This would not work as we split here recursively. Hence one has to sort before  L103-106.

jorana:
Sorry, perhaps I am a bit slow after my vacation but why would, after splitting, the time-ordering not be preserved?

WenzDaniel:
For peaks it is yes but not for hitlets as you have them for each channel. 
Lets assume we split a hitlet in channel 1 into two: 
```
ch1.: |-----|-----|
ch2::       |----|
```
If you do not sort again the order would be ch_1_1 ch_1_2 ch_2 while the correct order is ch_1_1, ch_2, ch_1_2.

jorana (approved):
Thanks Daniel

PR #492, opened 2021-07-22
Added pytest.ini

WenzDaniel:
Tiny PR which adds a ini for pytest which sets `numba.NumbaExperimentalFeatureWarning` to ignore. Reduce the amount of lengthy warning messages during testing.


jorana (approved):
Thanks Daniel

Issue #503, opened 2021-08-01
Stop supporting/testing with python 3.6


jorana:
Many packages are not supporting python 3.6 anymore. See e.g. numpy, pandas, tensorflow. This leads to different requirements in the tests which will continue diverging. See the [current requirements file for testing purposes](https://github.com/AxFoundation/strax/blob/master/extra_requirements/requirements-tests.txt).

This has been going on for a while, see the release schedule of numpy (https://numpy.org/neps/nep-0029-deprecation_policy.html)

> On Jun 23, 2020 drop support for Python 3.6 (initially released on Dec 23, 2016)

Since we rely heavily on numpy as well do many other requirements, we might as well follow suit. I don't see compelling reasons to keep supporting/testing on py3.6.

@ershockley @ahiguera-mx @jmosbacher @WenzDaniel , would you see any issue with this?

WenzDaniel:
Yes I agree, I do not remember were I read it recently, (either pyhep or scipy?) but most of the scientific stacks are ending their python 3.6 support. I wanted to discuss this point with you anyhow, but did not want to disturb during your well deserved holidays.

jmosbacher:
I am very much in favor of this. python 3.6 sucks

ahiguera-mx:
No objection from my side.

Aaron Higuera

PR #505, opened 2021-08-03
disable testing py3.6

jorana:
Disable testing with python3.6, fix issue #503


jmosbacher (approved):
Great!

PR #506, opened 2021-08-03
Fix peaks merging

WenzDaniel:
When comparing the reconstructed S2 area with raw data covered by S2s we have a bias of few percent. The bias comes from the fact that we are not considering lone hits in between peaklets during the merging step. This affects especially few electron S2s.  

In this PR we are adding a tiny function which adds the are of lone hits to merged S2s if they are fully contained in a S2. Further, we updated the peak-data fields with a delta pulse as approximated representative for the corresponding lone_hits. 
This approximation allows us to get a better estimate on the peak width and rise time. This overcomes the issue of loading records again which would make playing around with peak building more cumbersome.  

See also: https://github.com/XENONnT/straxen/pull/623


jmosbacher (approved):
looks fine to me.

jorana:
Would it have been worth writing a small test for this PR?

PR #507, opened 2021-08-05
A few improvements on handling ONLINE vs OFFLINE corrections

ahiguera-mx:
In the current configuration changes in the past are not allowed for either ONLINE or OFFLINE versions.
This PR allows user to change values in the past for OFFLINE versions only

This PR allows user to change values in the past for OFFLINE versions only

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/corrections.py
+++ b/strax/corrections.py
@@ -160,18 +160,32 @@ class CorrectionsInterface:
         for req in required_columns:
             if req not in df.columns:
                 raise ValueError(f'Must specify {req} in dataframe')
-        # Compare against
+        # Compare against existing data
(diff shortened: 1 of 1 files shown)


jmosbacher (changes requested):
Two things: 
1) Can you please add a check that any replaced values in the past are nans?
2) The logic here seems a bit unintuitive. why are you checking each old row in the new dataframe and requiring it to be the same instead of just selecting all rows with a date after now from the new dataframe and concating it with all the rows from the old dataframe that have a date prior to now? That way the user doesn't have to have all the old rows in his dataframe, he can just have the new ones.

PR #508, opened 2021-08-05
Added export

WenzDaniel:


jorana (approved):
👍

jorana:
[Would it have been worth writing a small test for this #506?](https://github.com/AxFoundation/strax/pull/506#issuecomment-893274065)

WenzDaniel:
Yes, I agree, I only thought that we would like to merge this week such that we can reprocess. Since you need more time for the other PR I can add it in the meantime.

tests/test_peak_merging.py
```diff
@@ -48,3 +48,29 @@ def test_replace_merged(intervals, merge_instructions):
...
+
+@hypothesis.given(fake_hits,
+                  hypothesis.strategies.integers(min_value=0, max_value=100))
+@hypothesis.settings(deadline=None)
+def test_add_lone_hits(hits, peak_length):
+    peak = np.zeros(1, dtype=strax.peak_dtype())
+    peak['length'] = peak_length
+    peak['dt'] = 1
+
+    to_pe = np.ones(10000)
```
jorana, line 61:
```suggestion
    to_pe = np.ones(np.max(hits['channel']))
```

jorana (approved):
Thanks Daniel!

tests/test_peak_merging.py
WenzDaniel: I had this before :D That is why the previous version was failing. In case hits are empty np.max has no identity. Hence I put 10k to avoid some more complicated if else thingy.

jorana: haha okay, I see

jorana: are you sure by the way? I see that there was a typo in `hits` that was the issue. But okay, it's really not much of an issue

WenzDaniel: Yes, I corrected the typo first and hit then the other issue.

PR #509, opened 2021-08-06
Allow creation and storing of superruns if SaveWhen > 0

WenzDaniel:
Changed get_components such that superruns can be created from every plugin type except SaveWhen.NEVER plugins. This addresses: https://github.com/XENONnT/straxen/issues/605

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -723,7 +723,7 @@ class Context:
                 # we have to deactivate the storage converter mode.
                 stc_mode = self.context_config['storage_converter']
                 self.context_config['storage_converter'] = False
-                self.make(list(sub_run_spec.keys()), d)
+                self.make(list(sub_run_spec.keys()), d, save=(d,))
                 self.context_config['storage_converter'] = stc_mode
 
                 ldrs = []
@@ -744,7 +744,8 @@ class Context:
                     if not ldr:
                         raise RuntimeError(
                             f"Could not load {d} for subrun {subrun} "
-                            f"even though we made it??")
+                             "even though we made it? Is the plugin "
+                             "you are requesting a SaveWhen.NEVER-plguin?")
                     ldrs.append(ldr)
 
                 def concat_loader(*args, **kwargs):
@@ -777,6 +778,7 @@ class Context:
                     raise strax.DataNotAvailable(
                         f"{d} for {run_id} not found in any storage, and "
                         "your context specifies it cannot be created.")
+                    
                 to_compute[d] = p
                 for dep_d in p.depends_on:
                     check_cache(dep_d)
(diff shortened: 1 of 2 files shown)


jorana:
Quick question, although this seems to works, would it not be more intuitive to always make the data when making a super-run, also when using a SaveWhen.NEVER plugin or do you foresee a reason why not to do this? Otherwise, for SaveWhen.NEVER we will keep the same behavior right?

WenzDaniel:
Yes, I thought about it too. But in the end I decided against it to distinguish more explicitly between SaveWhen.NEVER and SaveWhen.EXPLICIT. I think the latter was added later, was not it? At least I had the impression that SaveWhen.EXPLICIT makes SaveWhen.NEVER a bit obsolete. I think any plugin should be saveable if asked for, unless there might be some technical reasons for which we would forbid saving and make it a SaveWhen.NEVER plugin.

jorana:
Right, I never used savewhen explicit, I had to write a test to understand it:
https://github.com/AxFoundation/strax/pull/510

But agree, SaveWhen.Never is a bit of a strange choice generally when SaveWhen.Explicit is available

strax/context.py
```diff
@@ -798,7 +800,9 @@ def concat_loader(*args, **kwargs):
                 if d not in targets:
                     return
             elif p.save_when == strax.SaveWhen.EXPLICIT:
-                if d not in save:
+                # If we arrive here in case of a superrun the user want to save
+                # as self.context_config['write_superruns'] is true.
+                if d not in save and not _is_superrun:
```
jorana, line 805: Do I get it correctly that this means now we will always save it? Because if `_is_supperrun` and `p.save_when == strax.SaveWhen.EXPLICIT` we will always save it irrespective of the save argument?

WenzDaniel: This one we can debate. So here is the example I was think of: 

The user would like to make a superrun for a SaveWhen.EXPLICIT plugin and sets `st.set_context_config({'write_superruns': True})`. If we do not put `and not _is_superrun` here and the user calls `st.make(superrun_id, target)` without `save=(target,)` the data for the subruns will be created (if the data does not exists), but the superrun will not. I thought since the user specified already `st.set_context_config({'write_superruns' he/she wishes to store the superrun and we save analysts the desperate experience in creating data for hours before they realize in the end they forgot to put the `save`-argument.

jorana: okay, of course it shouldn't take an hour but I see. I would have just raised an error when someone would request making a superrun for an explicit plugin without the save argument but this approach also makes sense. Perhaps add a small comment for or check for the 'write_supperruns' ?

How does this work for the components in the processing chain that are not part of the request? E.g. I ask for `cut_kr` depending on `event_info_double` (savewhen.explicit), do we then also start writing `event_info_double`, I think we do right?

WenzDaniel: > How does this work for the components in the processing chain that are not part of the request? E.g. I ask for cut_kr depending on event_info_double (savewhen.explicit), do we then also start writing event_info_double, I think we do right?

No we should not. In line 726 we process all subruns which do not exists yet by calling `make(sub_run_id, target, save_when=(target, ))` E.g. in your example we would do something like `make('0123456', cut_kr, save_when=(cut_kr, ))` . Hence `event_info` double should not be stored.

jorana: okay thanks

jorana (approved):

Issue #511, opened 2021-08-13
Double dependencies freez mailbox system


WenzDaniel:
**Problem:**
In some studies we noticed a peak reconstruction bias due to our peak merging procedure. Currently, if we merge two peaklets with a lone hit in between the information of the lone hit is not added to the peak. This leads to an area bias of a few percent in the reconstructed peaks. In order to solve this issue I made https://github.com/AxFoundation/strax/pull/506 and https://github.com/XENONnT/straxen/pull/623. These changes require a double dependency of merged_s2s on peaklets (see also attached graph). In addition peaklets and lone_hits are chunked differently. 

This double dependency is currently not properly supported and let our mailbox system end up in a dead lock when processing up to merged_s2s without pre-processing peaklets. If peaklets are made before calling `st.make(.... `merged_s2s`)` things work as expected. 
 
![example](https://user-images.githubusercontent.com/43881800/129310692-419ca8a1-7f58-404c-a83d-cda42b5a258c.png)

WenzDaniel:
Let me also add the debugging logging info:

```
DEBUG:strax:Final plugins: ('merged_s2s',)
DEBUG:ThreadedMailboxProcessor:Processor components are: ProcessorComponents(plugins={'merged_s2s': MergedS2s, 'peaklets': Peaklets, 'peaklet_classification': PeakletClassification, 'lone_hits': Peaklets}, loaders={'records': functools.partial(<bound method StorageFrontend.loader of strax.storage.files.DataDirectory, path: /dali/lgrandi/wenz/strax_data/strax_tests>, 025258-records-wowj65ge6b, time_range=None, chunk_number=None, fuzzy_for=(), fuzzy_for_options=(), allow_incomplete=False)}, savers={'peaklets': [<strax.storage.files.FileSaver object at 0x7fc4242eefa0>], 'lone_hits': [<strax.storage.files.FileSaver object at 0x7fc4242feeb0>], 'peaklet_classification': [<strax.storage.files.FileSaver object at 0x7fc45c6be340>], 'merged_s2s': [<strax.storage.files.FileSaver object at 0x7fc4242eebb0>]}, targets=('merged_s2s',))
[...]

PR #512, opened 2021-08-17
Fix double dependency

WenzDaniel:
In this PR we are fixing two different issues. 

1. The double dependency issue as described in https://github.com/AxFoundation/strax/issues/511. This issue is triggered in lazy and multiprocessing mode. 
2. An issue in inline_plugins of our ParallelSource-plugin. The issue is triggered if the lowest data type involved in the plugin aliment is already created. In that case a [loader instead of the plugin](https://github.com/AxFoundation/strax/blob/48411c797bfe435aaf7b3f2993cc3ee222020af4/strax/context.py#L757-L760) itself is added to the processor components which will trigger the following error:

```
>   p.deps = {d: plugins[d] for d in p.depends_on}
E   KeyError: 'records'
```

The issue can be for example triggered in the following way: 

```
straxer some_run --target records --context xenonnt_online --workers 8 --notlazy --multiprocess --profile_ram --build_lowlevel 
straxer some_run --target merged_s2s --context xenonnt_online --workers 8 --notlazy --multiprocess --profile_ram --build_lowlevel 
```

As those issues are not related two distinct solutions were made. But before explaining the solutions let me first start with the other stuff I changed in this PR.

Additional tests:

Before working on the solutions, I extended our multi_output test in order to reliably trigger the issues mentioned above. The following tests were added:

1. test_double_dependency
2. test_double_dependency_notlazy
3. test_double_dependency_multiprocess
4. test_double_dependency_inline_plugins

The first 3 tests trigger the first issue while the last one triggers the second issue. I also had to extend the already existing Plugin landscape in strax for that. I added `ZipRecords` to have a minimal testing example for issue 1. plus `EvenRecsClassified` and `ZipRecordsAdditionalDependency` to mimic the plugin structure in https://github.com/AxFoundation/strax/issues/511 which can cause both issue. 

Additional logs:

Especially, the first issue was difficult to debug due to the lack of logging information. Hence I added a small class which wraps our thread locking and notifying conditions to tackle this issue. Also for the second issue I kept some of the most general debug logging for future issues (Hopefully may they never be used....). 

Issue 1.: 

In order to lift the dead lock issue we have to make sure that at least one mailbox of the multi-output plugin is skipped during the initial fetching. This works similarly as the flow_freely parameter which is also used in divided-outputs. However, we cannot use this parameter directly as it would cause other failures during processing, like out of order chunks. 

Issue 2.: 

I added to our processing components in addition to the plugins components the `loader_plugins` dictionary. This allows us to load the plugin dependency for the parallel source plugin as it is not a part of the plugin dictionary.

Just rerun the straxer commands above.

### DIFF
--- a/strax/mailbox.py
+++ b/strax/mailbox.py
@@ -504,8 +504,8 @@ def divide_outputs(source,
                     # Do not block on account of these guys
                     not_block_reason = 'flow_freely' if flow_freely else 'is first data_type'
                     m.log.debug(f'Not locking {d} because it {not_block_reason}')
+                    is_first = False
                     continue
-                is_first = False
                 m = mailboxes[d]
                 if lazy:
                     with m._lock:
--- a/strax/mailbox.py
+++ b/strax/mailbox.py
@@ -493,9 +493,6 @@ def divide_outputs(source,
     # TODO: this code duplicates exception handling and cleanup
     # from Mailbox.send_from! Can we avoid that somehow?
     i = 0
-    # Flag which prevents the first data type of a divide_outputs plugin
-    # to lock the threads. Otherwise this may lead to a dead-lock in
-    # case of a double dependency plugins (e.g. merged_s2s).
     try:
         while True:
             for d in outputs:
--- a/strax/mailbox.py
+++ b/strax/mailbox.py
@@ -493,14 +493,14 @@ def divide_outputs(source,
     # TODO: this code duplicates exception handling and cleanup
     # from Mailbox.send_from! Can we avoid that somehow?
     i = 0
-    counter = 0
+    is_first = True
     try:
         while True:
             for d in outputs:
-                if d in flow_freely or counter < 1:
+                if d in flow_freely or is_first:
                     # Do not block on account of these guys
                     continue
-                counter += 1
+                is_first = False
                 m = mailboxes[d]
                 if lazy:
                     with m._lock:
--- a/strax/mailbox.py
+++ b/strax/mailbox.py
@@ -493,12 +493,17 @@ def divide_outputs(source,
     # TODO: this code duplicates exception handling and cleanup
     # from Mailbox.send_from! Can we avoid that somehow?
     i = 0
+    # Flag which prevents the first data type of a divide_outputs plugin
+    # to lock the threads. Otherwise this may lead to a dead-lock in
+    # case of a double dependency plugins (e.g. merged_s2s).
     is_first = True
     try:
         while True:
             for d in outputs:
                 if d in flow_freely or is_first:
                     # Do not block on account of these guys
+                    not_block_reason = 'flow_freely' if flow_freely else 'is first data_type'
+                    m.log.debug(f'Not locking {d} because it {not_block_reason}')
                     continue
                 is_first = False
                 m = mailboxes[d]
--- a/strax/mailbox.py
+++ b/strax/mailbox.py
@@ -496,16 +496,13 @@ def divide_outputs(source,
     # Flag which prevents the first data type of a divide_outputs plugin
     # to lock the threads. Otherwise this may lead to a dead-lock in
     # case of a double dependency plugins (e.g. merged_s2s).
-    is_first = True
     try:
         while True:
             for d in outputs:
                 m = mailboxes[d]
-                if d in flow_freely or is_first:
+                if d in flow_freely:
                     # Do not block on account of these guys
-                    not_block_reason = 'flow_freely' if flow_freely else 'is first data_type'
-                    m.log.debug(f'Not locking {d} because it {not_block_reason}')
-                    is_first = False
+                    m.log.debug(f'Not locking {d}')
                     continue
                 if lazy:
                     with m._lock:
--- a/strax/mailbox.py
+++ b/strax/mailbox.py
@@ -500,13 +500,13 @@ def divide_outputs(source,
     try:
         while True:
             for d in outputs:
+                m = mailboxes[d]
                 if d in flow_freely or is_first:
                     # Do not block on account of these guys
                     not_block_reason = 'flow_freely' if flow_freely else 'is first data_type'
                     m.log.debug(f'Not locking {d} because it {not_block_reason}')
                     is_first = False
                     continue
-                m = mailboxes[d]
                 if lazy:
                     with m._lock:
                         if not m._can_fetch():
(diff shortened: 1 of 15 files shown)


strax/plugin.py
```diff
@@ -902,7 +902,9 @@ class ParallelSourcePlugin(Plugin):
     @classmethod
     def inline_plugins(cls, components, start_from, log):
         plugins = components.plugins.copy()
-
+        loader_plugins = components.loader_plugins.copy()
```
jorana, line 905: You mention this in the PR description but for my understanding, why could we not add this to the `plugins`-dict  and have to make a separate attribute?

strax/mailbox.py
```diff
@@ -467,13 +493,14 @@ def divide_outputs(source,
     # TODO: this code duplicates exception handling and cleanup
     # from Mailbox.send_from! Can we avoid that somehow?
     i = 0
+    counter = 0
```
jorana, line 496: Perhaps make it a bit more explicit (I would have to figure out the difference between `counter` and `i` whereas ` counter` is mostly a flag right)?
```suggestion
    at_least_one_flowing = False
```
If I understand, this is the key fix to the deadlock? If so perhaps add a few comments / ref to this PR

strax/mailbox.py
```diff
@@ -467,13 +493,14 @@ def divide_outputs(source,
...
     try:
         while True:
             for d in outputs:
-                if d in flow_freely:
+                if d in flow_freely or counter < 1:
```
jorana, line 500: Are by the way sure about this being right? The ` flow_freely` mailboxes are the one that are not required by any other, if we continue here, does it mean we will not be able to read it elsewhere (I might have to read up a bit on the mailbox system)?

https://github.com/AxFoundation/strax/blob/master/strax/processor.py#L102-L116

jorana (commented):
Thanks Daniel, also for adding tests and better logging!

I'm always a bit slow on the Mailbox system so I have a couple of questions.

strax/plugin.py
WenzDaniel: Because then it would be recomputed.

strax/mailbox.py
WenzDaniel: No this is not the case. Although I just notice I did not include it in the tests as I do not check if data is stored afterwards. I will add it. 
In my first attempt I added it to the flow_freely parameters but this does not work due to your mentioned dependencies. However, if you just skip it during the first fetching you prevent the mailboxes from locking each other. I would lie if I say I understand it by 100% but this is what I learned when writing down the "locking-scheme" for a working and non-working case. In addition I processed two background runs and it worked like a charm. All data has been created and is loadable.

jorana: okay, I must say I don't fully follow this either. If we are are not 100% sure we understand, please don't hold back on adding comments like `preventing is_first from locking to prevent dependencies deadlocks` and adding a few debug statements of which `d` is actually not locked because of `is_first`

jmosbacher (commented):
Very nice PR, the explanation is really clear. Tests are very nice. 
I guess the tests not passing is due to #510 ?

WenzDaniel:
> I guess the tests not passing is due to #510 ?

No, I realized that I did a mistake in the solution for issue 1. The issue is solved by the proposed changes, but only due to the fact that I did a mistake. I thought the issue would be solved by simply not calling m._lock for all data_type dependencies. However, it turned out in the proposed changes I do not call m._fetch for any data_type dependency. 

Hence, I am currently working on a correct solution. I think I understand my mistake but I am currently testing. The solution for issue 2 should be fine as it is.

WenzDaniel:
@jmosbacher and @jorana, I apologize once more for the hastily review request I did. Now the problem is solved and understood. Let me explain why case 1 leads to a dead-lock.

In case of a multi-output plugin we are calling [divided_outputs](https://github.com/AxFoundation/strax/blob/4adacdcf212b4d39662b44bfaad57b9ffa453fa1/strax/processor.py#L146-L166) which will create a single reader plugin depending only on one of the provided outputs (e.g. `peaklets` for `peaklets` and `lone_hits`). Hence for this purpose the [flow-freely parameter](https://github.com/AxFoundation/strax/blob/4adacdcf212b4d39662b44bfaad57b9ffa453fa1/strax/mailbox.py#L473-L475) was introduced which is computed based on the [required and provided](https://github.com/AxFoundation/strax/blob/4adacdcf212b4d39662b44bfaad57b9ffa453fa1/strax/processor.py#L106-L117) plugins. Data_types which are tagged as flow freely cannot lock the threading in lazy mode as they do not have any reader associated with (because it is not needed they are read together with the other output from the multi-output plugin). Hence, if we [lock](https://github.com/AxFoundation/strax/blob/4adacdcf212b4d39662b44bfaad57b9ffa453fa1/strax/mailbox.py#L479-L480) our threads in those cases they never can be woken up and we end-up in a dead-lock.

This is exactly what happens for plugins with a double dependency on a multi-ouput plugin. In that case both data_types are registered as `provided` and `required` and the flow_freely `set` is empty. Hence I added these [lines](https://github.com/AxFoundation/strax/pull/512/commits/e4dfd5298469f26f26d8927ba261f35489df6bab) in porcessor.py to update flow_freely accordingly for those cases.

strax/mailbox.py
```diff
@@ -467,14 +493,17 @@ def divide_outputs(source,
     # TODO: this code duplicates exception handling and cleanup
     # from Mailbox.send_from! Can we avoid that somehow?
     i = 0
+    # Flag which prevents the first data type of a divide_outputs plugin
+    # to lock the threads. Otherwise this may lead to a dead-lock in
+    # case of a double dependency plugins (e.g. merged_s2s).
```
WenzDaniel, line 498:
```suggestion
```

strax/processor.py
```diff
@@ -156,12 +157,21 @@ def __init__(self,
                         executor=executor),
                     name=f'divide_outputs:{d}')
 
+                # If we have a plugin with double dependency both outputs
+                # of a multioutput-plugin are required. Hence flow-freely
+                # is empty an needs to be updated here:
+                provided_data_types = set(p.provides)
+                reader_data_types = set(strax.to_str_tuple(d))
+                _flow_freely = (to_flow_freely | ((provided_data_types - reader_data_types) -
+                                                  to_flow_freely))
```
jorana, line 166: Please make sure if this is correct, but it looks like this could be a bit simpler, right?
```suggestion
                double_dependency = (provided_data_types - reader_data_types)
                to_flow_freely |= double_dependency 
```

jorana (approved):

strax/processor.py
jorana: you'd also have to change the _flow_freely a few lines down

jmosbacher (approved):
Looks good to me. Maybe there is some edge case where this fails but I cant think of any.

PR #514, opened 2021-08-24
Update splitter for merged s2s data kind

zhut19:
The gap-size-based merging algorithm we are using now is not able to / optimized for splitting merged multiple scatter S2s, and thus motivates potentially adding the natural break after the merging. This pr is going to update the splitter so that it can also process merged s2s, particularly to circumvent calling sum waveform.

The first change is to the peak building part that for merged s2s, instead of resetting `dt` and `data` with the records, we simply copy the part of `data` from the pre-split peak and keep the same `dt`. (One thing to note is that those two fields are only kept within the splitting process, so the split iteration can work. Once outside, those split peaks will be rebuilt from peaklets, so that they have the correct area_per_channel, saturation info, and so on)

The second change is to make it possible to return the peaks in two arrays, one without any splitting and the new peaks from the splitting. This return option is needed, as mentioned above, we need to rebuild the peaks from peaklets but just for those split peaks.

### DIFF
--- a/strax/processing/peak_merging.py
+++ b/strax/processing/peak_merging.py
@@ -92,7 +92,7 @@ def replace_merged(orig, merge):
         return orig
 
     skip_windows = strax.touching_windows(orig, merge)
-    skip_n = len(np.unique(np.concatenate([np.arange(l, r) for l, r in skip_windows])))
+    skip_n = np.diff(skip_windows, axis=1).sum()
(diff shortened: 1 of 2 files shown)


JoranAngevaare:
Hi Tianyu, should we keep this PR or can I close it and re-open if it comes up again?

PR #529, opened 2021-09-01
Add read by index plus some extra checks

ahiguera-mx:
Add function to read by index (collections include index in the db)
Without indexes, a MongoDB database has to scan every document in a collection to look for the relevant documents. However, using indexes greatly reduces the number of documents MongoDB needs to scan
`db.collection.find({'document':'value'}).explain()['executionStats'])` should return `IXSCAN` which is faster than `COLLSCAN` 
  - Update the docstring(s)
  - Update the documentation


strax/corrections.py
```diff
@@ -75,6 +75,28 @@ def list_corrections(self):
         database = self.client[self.database_name]
         return [x['name'] for x in database.list_collections()]
 
+    def read_by_index(self, correction, when):
```
jmosbacher, line 78: the naming is not descriptive of what the functions is doing. The function will return the closest item in time before and after the given date. That is true whether the time field is indexed on mongodb or not. I would rename to "read_at" or something that would help a maintainer (or you in a year from now) quickly understand what the function does without reading the code.

strax/corrections.py
```diff
@@ -160,25 +182,58 @@ def write(self, correction, df, required_columns=('ONLINE', 'v1')):
...
             if req not in df.columns:
                 raise ValueError(f'Must specify {req} in dataframe')
-        # Compare against
+        # Compare against existing data
+        # We can change OFFLINE values in the past
+        # We cannot change ONLINE values in the past
+        # We can add a new date(row) in the past(ONLINE) as long as it is the same value as in the past
         logging.info('Reading old values for comparison')
-        df2 = self.read(correction)
+        df_old = self.read(correction)
```
jmosbacher, line 190: why not just read the database from the earliest time being set? this still pulls the entire collection right?

strax/corrections.py
```diff
@@ -160,25 +182,58 @@ def write(self, correction, df, required_columns=('ONLINE', 'v1')):
...
+                                new_value = df.loc[df.index == new_date[i].to_pydatetime(), column].values[0]
+                                old_value = df_old.loc[df_old.index < new_date[i].to_pydatetime(), column][-1]
+                                if new_value != old_value:
+                                    raise ValueError(f'{column} changed in past, not allowed')
+            else:
+                for column in df_old.columns:
+                    if 'ONLINE' in column:
+                        if not (df.loc[df.index < now, column] == df_old.loc[df_old.index < now, column]).all():
+                            raise ValueError(f'{column} changed in past, not allowed')
+
```
jmosbacher, line 212: the logic here is very verbose and unnecessarily complex. Why arent you using pandas builtin API for this?

strax/corrections.py
```diff
@@ -160,25 +182,58 @@ def write(self, correction, df, required_columns=('ONLINE', 'v1')):
...
+    @staticmethod
+    def after_date_query(date):
+        return [{"$match": {"time": {"$gte": pd.to_datetime(date),}
+                           }
+                },
+                {"$sort": {"time": 1}},
+                {"$limit": 1}
+                ]
+
```
jmosbacher, line 236: you should add a variable that sets how many items you want on each side `after_date_query(date, limit=1)

strax/corrections.py
```diff
@@ -160,25 +182,58 @@ def write(self, correction, df, required_columns=('ONLINE', 'v1')):
...
-                logging.debug(f'Checking {column}')
-                if not (df2.loc[df2.index < now, column] ==
-                        df.loc[df.index < now, column]).all():
-                    raise ValueError(f'{column} changed in past, not allowed')
-
+            if len(df_old) < len(df):
+                new_date = df.index.difference(df_old.index)
+                logging.info(f'A new date(row) was added {new_date}, lets check past values')
+                for column in df_old.columns:
+                    if 'ONLINE' in column:
```
jmosbacher, line 200: what if the column is not online? shouldn't you check that all the replaced values are nans?

jmosbacher (changes requested):
still some work to do in updating values

strax/corrections.py
ahiguera-mx: good suggestion

ahiguera-mx:
Hi Yossi,
Thanks for the very useful comments, I believe the current logic covers all usage cases

jmosbacher (approved):
I think this is safe to merge but I think it would really help to have a few test cases added at some point to ensure the updating logic remains consistent when things change

ershockley (approved):

WenzDaniel:
Hej guys short question. Is this PR relevant for reprocessing? If not I would prefer to add those tests first, mentioned by Yossi. I know it is annoying, but based on experience I also know if it is not done right away we tend to forget.

ahiguera-mx:
> Hej guys short question. Is this PR relevant for reprocessing? If not I would prefer to add those tests first, mentioned by Yossi. I know it is annoying, but based on experience I also know if it is not done right away we tend to forget.

I think is needed for reprocessing, but @ershockley should confirm. We already have some automated test but those are in straxen, here https://github.com/XENONnT/straxen/blob/master/tests/test_cmt.py

ershockley:
Yes we need this for reprocessing.

WenzDaniel:
> We already have some automated test but those are in straxen, 

Yes but this is strax and not straxen. Strax is a stand a lone thing. Hence we should test it also here. As I said I know it is an annoying work, but it pays off later.

strax/corrections.py
```diff
@@ -75,6 +75,29 @@ def list_corrections(self):
         database = self.client[self.database_name]
         return [x['name'] for x in database.list_collections()]
 
+    def read_at(self, correction, when, limit=1):
+        """Smart logic to read corrections at given time (index), i.e by datetime index
+        :param correction: pandas.DataFrame object name in the DB (str type).
+        :return: DataFrame as read from the corrections database with time
```
WenzDaniel, line 81: Could you also add when and limit to the doc-string? I do not know what this is suppose to be.

strax/corrections.py
```diff
@@ -75,6 +75,29 @@ def list_corrections(self):
...
+        index or None if an empty DataFrame is read from the database
+        """
+        before_df = pdm.read_mongo(correction, self.before_date_query(when, limit), self.client[self.database_name])
+        after_df = pdm.read_mongo(correction, self.after_date_query(when, limit), self.client[self.database_name])
+
+        df = pd.concat([before_df, after_df])
```
WenzDaniel, line 87: Sorry, but I do not understand this logic here. Why is this different from querying the entire time range at once? Unfortunately, I do not know much about MongoDB queries. Or does this limit mean you pull the last/first entry before and after the specified time?

strax/corrections.py
```diff
@@ -160,25 +183,60 @@ def write(self, correction, df, required_columns=('ONLINE', 'v1')):
         for req in required_columns:
             if req not in df.columns:
                 raise ValueError(f'Must specify {req} in dataframe')
-        # Compare against
+        # Compare against existing data
+        # We can change OFFLINE values in the past only if they are NaN
+        # We cannot change ONLINE values in the past
+        # We can add a new date(row) in the past(ONLINE) as long as it is the same value
```
WenzDaniel, line 189: Can we  also get some developer documentation about this (not needed in this PR but in general)? CMT has become a very integral part and should be much better documented. Especially, about the jargon about ONLINE and OFFLINE corrections. How corrections should be implemented etc.

strax/corrections.py
```diff
@@ -75,6 +75,29 @@ def list_corrections(self):
...
+        # No data found
+        if df.size == 0:
+            return None
+        # Delete internal Mongo identifier
+        del df['_id']
+
+        df['time'] = pd.to_datetime(df['time'], utc=True)
+        df = df.set_index('time')
+        df = df.sort_index()
+        return df
```
WenzDaniel, line 98: Duplicate code with read method.

strax/corrections.py
```diff
@@ -160,25 +183,60 @@ def write(self, correction, df, required_columns=('ONLINE', 'v1')):
...
+                for column in df_old.columns:
+                    if 'ONLINE' in column:
+                        if not (df.loc[df.index < now, column] == df_old.loc[df_old.index < now, column]).all():
+                            raise ValueError(f'{column} changed in past, not allowed')
+                    else:
+                        for i, item in enumerate(new_dates):
+                            old_value = df.loc[df.index < new_dates[i].to_pydatetime(), column][-1]
+                            if not np.isnan(old_value).all():
+                                raise ValueError(f'{column} only NaN values can be updated')
+
```
WenzDaniel, line 213: I am sorry but I do not understand this logic here. To me it seems to be way to complicated. Further, I do not understand how you can change values in the past. Should not the ONLINE-part not always raise an ValueError in this case? Or is "correction" a string like "ONLINE_gain" or just "gain" and you get all columns "ONLINE", "v1", "v2" etc.?  Could you maybe illustrate with a sketch and explain in words what you would like to achieve?

WenzDaniel (commented):
In your write method: `:param required_columns: DataFrame must include an online and v1 columns` should not this be a property of the class. Can you have the same class instance but with different requirements on the minimal corrections shape? Or is this an option load specific columns. E.g. I can also specify to load ONLINE till version v4? In this case I think the doc-string is a bit misleading. 

I am sorry, but for me it is not possible to tell whether the code will do the desired change. Could you give some additional information and draw a sketch? 
In addition I think we need desperately much more documentation here. I am sorry, but to be honest normally I would not approve this PR.

strax/corrections.py
```diff
@@ -160,25 +183,60 @@ def write(self, correction, df, required_columns=('ONLINE', 'v1')):
...
-                        df.loc[df.index < now, column]).all():
-                    raise ValueError(f'{column} changed in past, not allowed')
+            new_dates = df.index.difference(df_old.index)
+            if not new_dates.empty:
+                for i, item in enumerate(new_dates):
+                    if item < now:
+                        for column in df_old.columns:
+                            new_value = df.loc[df.index == new_dates[i].to_pydatetime(), column].values[0]
+                            old_value = df_old.loc[df_old.index < new_dates[i].to_pydatetime(), column][-1]
+                            if new_value != old_value:
```
a1exndr, line 201: I think here you should also add a check whether new_value and old_value are both NaN, as `np.nan == np.nan` will return False and hence raise a ValueError, which we do not want in that case. Or do I misunderstand the purpose of this check or the possible values of new_value and old_value (at least in the previous version of this check with `(df2.loc[df2.index < now, column] == df.loc[df.index < now, column]).all()` this lead to unwanted errors)?

strax/corrections.py
ahiguera-mx: This logic uses the advantages of querying by index, without indexes MongoDB database has to scan every document in a collection to look for the relevant documents. However, using indexes greatly reduces the number of documents MongoDB needs to scan so it is faster

strax/corrections.py
ahiguera-mx: not sure I understand what you are suggesting, I've tested and it works even in the cases of NaNs (as you already know from our slack discussion)

strax/corrections.py
ahiguera-mx: good catch, done

strax/corrections.py
ahiguera-mx: thanks, fixed

strax/corrections.py
ahiguera-mx: done

strax/corrections.py
ahiguera-mx: This logic is meant to allow changes to a pandas.dataframe, I can mentioned a few scenarios: the user wants to update the datetime (index) according to the actual time of the event e.g. (gain calibration time) this implies adding a new row. Another scenario, filling out non-physical values  (NaNs). 
In both cases this logic checks that 1) changes in the past are not allowed for ONLINE and only non-physical values can be modify in the past

ahiguera-mx:
Hi @WenzDaniel 
Thanks a lot for the useful comments,  I've tried to resolve all the issues raised to the best of my ability, please let me know if I missed something

ahiguera-mx:
Hi @WenzDaniel, 
In the latest commit I've added a test for the corrections part, so with this I think I addressed all comments. Thanks a lot again for the very useful review

requirements.txt
```diff
@@ -13,3 +13,4 @@ pdmongo
 lz4
 immutabledict
 packaging
+mongomock
```
jmosbacher, line 16: I dont think mongomock should be a requirement in the regular install, rather only for development.

jmosbacher:
I would like to see some more test cases. e.g. changes to existing dates with/without nans both in the past and future and for online and offline case. Also what about inserting new dates in-between existing ones?

ahiguera-mx:
> I would like to see some more test cases. e.g. changes to existing dates with/without nans both in the past and future and for online and offline case. Also what about inserting new dates in-between existing ones?

Hi @jmosbacher 
I've added more test, some of the test you suggest would throw an error which will break the automated test like changed non-nan values in the past

requirements.txt
ahiguera-mx: I have to add `mongomock.MongoClient` to `corrections.py` [here](https://github.com/ahiguera-mx/strax/blob/eef44093654055eb19970d98b0854f0f2667999e/strax/corrections.py#L52), if I remove mongomock, this would break, right?

jmosbacher:
> > I would like to see some more test cases. e.g. changes to existing dates with/without nans both in the past and future and for online and offline case. Also what about inserting new dates in-between existing ones?
> 
> Hi @jmosbacher
> I've added more test, some of the test you suggest would throw an error which will break the automated test like changed non-nan values in the past

@ahiguera-mx tests should be separated into different functions, the cmt setup should be a fixture and pytest.raises(Exception) should be used for testing correct behavior of errors.

requirements.txt
jmosbacher: you can use mongomock.patch() in the test instead of having mongomock.MongoClient in the codebase itself.

WenzDaniel:
> and pytest.raises(Exception) should be used for testing correct behavior of errors.

Yes there are several different methods which you can use to test whether a function or method returns the correct exception. unittest also offers such options.

WenzDaniel:
Hej Aaron I have a general comment about your tests. Maybe I am missing something, but I think you are not really testing if the outcome is correct, is not it? E.g. take for example `test_modify_nan`. You modify df2 and write the result, but you are not testing if the result is written correctly is not it? 
I would expect another read and check,

ahiguera-mx:
> Hej Aaron I have a general comment about your tests. Maybe I am missing something, but I think you are not really testing if the outcome is correct, is not it? E.g. take for example `test_modify_nan`. You modify df2 and write the result, but you are not testing if the result is written correctly is not it?
> I would expect another read and check,

Hey @WenzDaniel 
If there were any issues the `write()` will throw an error, as in the example `test_change_past`
For read and write a general test is done at `test_db()`

jmosbacher:
> > Hej Aaron I have a general comment about your tests. Maybe I am missing something, but I think you are not really testing if the outcome is correct, is not it? E.g. take for example `test_modify_nan`. You modify df2 and write the result, but you are not testing if the result is written correctly is not it?
> > I would expect another read and check,
> 
> Hey @WenzDaniel
> If there were any issues the `write()` will throw an error, as in the example `test_change_past`
> For read and write a general test is done at `test_db()`

@ahiguera-mx this is also the general comment I gave during the meeting yesterday, a test should not only check that the function doesn't throw an error when it runs to but also check that it has performed the correct task ie the resulting dataframe is consistent with our rules of what can and cannot be changed.

ahiguera-mx:
> > > Hej Aaron I have a general comment about your tests. Maybe I am missing something, but I think you are not really testing if the outcome is correct, is not it? E.g. take for example `test_modify_nan`. You modify df2 and write the result, but you are not testing if the result is written correctly is not it?
> > > I would expect another read and check,
> > 
> > 
> > Hey @WenzDaniel
> > If there were any issues the `write()` will throw an error, as in the example `test_change_past`
> > For read and write a general test is done at `test_db()`
> 
> @ahiguera-mx this is also the general comment I gave during the meeting yesterday, a test should not only check that the function doesn't throw an error when it runs to but also check that it has performed the correct task ie the resulting dataframe is consistent with our rules of what can and cannot be changed.
@jmosbacher @WenzDaniel 
What I meant by  "function throwing an error" is that the result is consistent with our rules of what can and cannot changed, I can add another read() and assert that dataframe is identical to what we modify, but it seems redundant to me. However if you guys think is necessary I'm happy to add it

WenzDaniel:
Hej Aaron, thanks for the clarification. I overlooked your `test_db`. I think things are already looking quite nice. I have one last suggestion and a request. 

Your tests have quite some duplicated code. You may want to look into this [test](https://github.com/AxFoundation/strax/blob/752a73c8b072bc8bdc9c353cb84ea12176bb1a65/tests/test_superruns.py#L15-L39) for example. You can see this set-up method. It is called always before you run any other test of this sub-class. In this way you could for example initialize your db and your dataframes. In this way we reduce the code duplication and you do not have to change all tests again in case we modify some CMT function. 

Regarding your `write`-method. Could we quickly meet Monday after the analysis meeting such that we can check if we can simplify the logic. I still do not understand your code. It is quite nested, but maybe I also have a wrong shape of the corrections in my mind.

PR #530, opened 2021-09-03
Add drop column option

WenzDaniel:
Trivial change, adds the option to drop columns from array/dataframe when reading data. I think this option is more useful than the already existing `keep_columns` option. In most use cases a user will only drop/keep a subfractions of columns if the data does not fit in memory. Hence, it is more effective to drop only unwanted large sized columns like `*per_pmt`-fields which can be easier done by just specifying what to drop instead of what to keep. This is especially true for peak like data. I also kept the old `keep_columns` option. The code will raise an error if the user specified both.
I extended the corresponding test too and updated all doc-strings.


strax/context.py
```diff
@@ -1652,6 +1654,7 @@ def add_method(cls, f):
 :param selection_str: Query string or sequence of strings to apply.
 :param keep_columns: Array field/dataframe column names to keep. 
     Useful to reduce amount of data in memory.
+:param drop_columns: Same but for dropping columns.
```
jorana, line 1657: _extemely minor_: perhaps be a bit more explicit (I usually don't read all the prams from top to bottom, only the ones that I suspect might be usefull

strax/utils.py
```diff
@@ -529,13 +529,16 @@ def iter_chunk_meta(md):
...
+                    drop_columns=None,
                     time_range=None,
                     time_selection='fully_contained'):
     """Return x after applying selections
 
     :param x: Numpy structured array
     :param selection_str: Query string or sequence of strings to apply.
     :param time_range: (start, stop) range to load, in ns since the epoch
+    :param keep_columns: Field names of the columns to keep.
+    :param drop_columns: Same but for dropping certain columns.
```
jorana, line 541: _extemely minor_: same comment as above

jorana (approved):
Nice feature, well documented and tested, a great example of a nice PR 👍

Issue #536, opened 2021-10-01
corrections write function uses numpy without importing it


jmosbacher:
**Describe the bug**
The corrections interface uses np.isnan [here](https://github.com/AxFoundation/strax/blob/5466cb4c15663f8b71cef1891cffd07438491ef8/strax/corrections.py#L193) but never imports numpy

**Expected behavior**
numpy should be imported in head of file

JoranAngevaare:
fixed in #537

PR #537, opened 2021-10-01
Fix issue 536

ahiguera-mx:
Fix issue https://github.com/AxFoundation/strax/issues/536
add numpy import 

  - Update the docstring(s)
  - Update the documentation


jmosbacher:
Maybe we can add a test that would explore this branch in the logic?

WenzDaniel:
I agree a simple test which would have run the code would be enough.

jmosbacher:
@ahiguera-mx will you be adding the test in this PR or a separate one?

ahiguera-mx:
@jmosbacher 
The test already exists here https://github.com/AxFoundation/strax/blob/master/tests/test_corrections.py
And in here the numpy was imported, I guess that is did not complain about the missing import

jmosbacher:
> @jmosbacher The test already exists here https://github.com/AxFoundation/strax/blob/master/tests/test_corrections.py And in here the numpy was imported, I guess that is why it did not complain about the missing import

Can you point me to the test that reaches this branch?

ahiguera-mx:
> > @jmosbacher The test already exists here https://github.com/AxFoundation/strax/blob/master/tests/test_corrections.py And in here the numpy was imported, I guess that is why it did not complain about the missing import
> 
> Can you point me to the test that reaches this branch?

This should be https://github.com/AxFoundation/strax/blob/master/tests/test_corrections.py#L53

jmosbacher:
> This should be https://github.com/AxFoundation/strax/blob/master/tests/test_corrections.py#L53

that doesn't seam right, I guess you mean  https://github.com/AxFoundation/strax/blob/master/tests/test_corrections.py#L63?

jmosbacher (approved):

WenzDaniel:
> nd in here the numpy was imported, I guess that is why it did not complain about the missing import

No, python does not work like this. You can test it by yourself. Just make two files with two simple functions e.g.:

```
# test.py

def test(a):
     return np.mean(a)
```

```
# test2.py
from test import test
import numpy as np
def test2(a):
     return test(a)
```

You will see that calling test2 from test2.py will fail. Hence the mentioned line of code does not seem to be reached by your tests. This is why it did not show up.

JoranAngevaare:
Hi all, it's sometimes not so clear which parts of the code are reached by the tests. Therefore we have codefactor:
https://coveralls.io/builds/43241044/source?filename=strax%2Fcorrections.py

Here you can see that the lines are (currently) not covered on the master branch:
afbeelding

Every time you make a commit to a PR, this can be improved and you can navigate to the relevant lines via the checks that are automatically running on github:
afbeelding


I hope this is helpful.

WenzDaniel:
Your tests have not tested the corresponding line of code. So it would be good to have one. Other then that it is fine.

PR #538, opened 2021-10-02
Two pbar patches

JoranAngevaare:
Two small patches to the pbar

Fix:

- afbeelding

and make this a notebook pbar:

- afbeelding


WenzDaniel (approved):
Thanks a lot. I think there was another case I found recently which was not covered. I think it was something like multi-run + usage of selection strings or something. If it still pops up I will revisit it in another PR.

PR #540, opened 2021-10-04
Add get_zarr method to context

jmosbacher:
This PR adds a get_zarr method to the context, to create persistent arrays which are useful for loading large datasets that dont fit in memory. 

For each requested target the method iterates over all run_ids and chunks and adds them to the given storage location (overwrite or append is optional). The zarr group is then returned to the user.

Please add comments if you think behavior should be different (e.g. auto merge targets with same datatypes or using hash in array names etc.) will add tests once the behavior is finalized.


JoranAngevaare:
Thanks Yossi, looking at the documentation of https://zarr.readthedocs.io/en/stable/tutorial.html, this looks like something very useful.

Few questions before diving into the code:
 - how does the zarr `'./strax_data'` get re-used? Now it looks like you are using the `target` rather than the `datakey` as the lookup. I think this can cause errors if one changes the context, right, it will still return the same data? If we can make this reproducible, it might be quite a benefit?
 - You chose to put this zarr outside of the storage-frontend system. Is the greatest advantage that you can create objects easily that transcend run_ids, I see that having zarr as another frontend also poses issues as you'll have two systems doing similar things? There are some cons too, like not being able to load data in it the "normal" way, as far as the context is concerned, the data only exists in this one function, and no other function is aware of it. Probably you thought of more pros/cons?
 - This functionality looks rather similar to the multi-run functionality, I'm not sure where which of the approaches has the benefit in which use case. Perhaps this is something worth adding to the docs?

WenzDaniel:
> this would allow users to mix and match runs from different contexts as they please 

I slightly disagree with this statement. Sounds a bit messy to me and to easy to screw-up. I think if you want to compare different "data-settings-contexts" you also should use different "strax-contexts". In this way we always have a nice and clean separation.

WenzDaniel:
Btw I am wondering if you cannot use the group feature for this level of organization: 
https://zarr.readthedocs.io/en/stable/tutorial.html#groups

jmosbacher:
If the behavior seems reasonable to all I will add an explanation to the docs and some tests.

WenzDaniel:
> indeed, but looking at their source code they use an ipytree widget for the notebook display of the tree so you need to install it explicitly or pip install zarr[jupyter], I guess we can add it to the dependencies if you think it would be useful for people

Yes this is what I did. I will try again later. 

> zgrp['peak_basics'].attrs['RUNS']

Ahh thanks I tried `zgrp.RUNS`.

jmosbacher:
I think this PR is ready to merge unless anyone wants some more changes. Maybe I should add an EXPERIMENTAL warning when this method is called?

WenzDaniel:
> Maybe I should add an EXPERIMENTAL warning when this method is called?

Sure if you like. I will advertise the PR later, but I am not so sure if many people will use it.

Issue #539, opened 2021-10-04
tests/test_hitlet.py test too stringent? 


JoranAngevaare:
Master failed on https://github.com/AxFoundation/strax/runs/3787298221?check_suite_focus=true for `test_conditional_entropy(data, size_template_and_ind_max_template)` with:
```python
data = array([-3.8847911, -0.5546908, -0.5546908, -8.079916 ,  7.364694 ,
        7.378992 , -0.5546908, -0.5546908, -0.5546908], dtype=float32)
size_template_and_ind_max_template = [1, 0]
```

Perhaps `rtol` is too small?

WenzDaniel:
Hmm sure, if this starts to become an issue we can loosen the tolerance a bit. I have a different question though Since currently our tests are passing, I was always in the belief hypothesis would retry a certain test if it failed in the past. Is this only true for tests run locally?

WenzDaniel:
Are you fine with observing the situation a bit before loosing the test? If so feel free to close.

JoranAngevaare:
@WenzDaniel can you have a look:
https://github.com/AxFoundation/strax/runs/4758294267?check_suite_focus=true

WenzDaniel:
Sure let me loosen the test requirement a bit.

PR #544, opened 2021-10-07
Failing example for test_conditional_entropy

JoranAngevaare:
Replicate #539 , let's also loosen the test requirement to prevent another fail


WenzDaniel:
Shall we keep the examples? I think they can be removed.

PR #556, opened 2021-10-15
Fixed several logic bugs in the write function of strax/corrections.py

a1exndr:
In `strax/corrections.py`, the `write` function checks whether the changes made in the new data that is to be written are allowed. At this, the main rule is to prohibit changes to correction values for times that may have already been processed. 

Several logical bugs were found in the current implementation of this function: 
- The code following the last `else` in the function will never be executed as there we try to loop over an empty list (`new_dates.empty` is `True` in this part).
- The current logic only checks the values for new timestamps (if there are any) or (if no new timestamps are added) changes of existing values. Hence, the case of a simultaneous change of old and addition of new values is not fully covered, as in that case only the newly added values (i.e. for new timestamps) will be checked.
- The current function allows adding a new value in the past if it is the same as its respective preceding existing value. This is reasonable for ONLINE correction values, as these are constantly extrapolated to the next correction value. However, for offline corrections that may be linearly interpolated between the given correction values, this can alter processed data. As the type of interpolation for the respective offline corrections is only specified at a later point, we do not know a priori whether a constant extrapolation (for which the current rule would not change the deduced correction values) or linear interpolation (for which the interpolated value would change) will be used. The current rule has therefore to be tightened (see below).
- We can block the processing of certain periods by setting offline correction values for selected times to NaN. All times that are directly preceded or succeeded by a NaN correction value will not return a usable correction value but will raise an error. We should hence be allowed (and this is the idea behind the concept of putting NaNs) to change NaN values and add new values before / after a NaN. This is not possible with the current code.

The proposed updated version of the `write` function in `strax/corrections.py` now follows the rules that we agreed upon, implemented as follows: 
- Changes / additions for future times are allowed.
- We cannot change ONLINE values in the past.
- We can add a new date (row) in the past for ONLINE as long as it has the same correction value as for the preceding existing time stamp. As ONLINE uses always the correction value of the latest preceding timestamp, this does not change the resulting correction values for any past times.
- For OFFLINE, only NaN values may be updated in the past. Data corrected by this values could not be processed at that point.
- We can add a new date (row) in the past for OFFLINE only if it does not affect already potentially processed times, i.e. if the existing correction value before or after the timestamp that is to be added is NaN. We also allow to add a new value (for new dates) if it is the same value as the entire column - this is relevant e.g. for indices or constant corrections, where for none of the possible interpolation methods the resulting values will be changed. Adding new NaN values (for new dates) is allowed as well. This is especially indispensable to fill the newly created entries in  deprecated versions if a new row is added in order to update e.g. the latest offline version in the past (mind the table-structure of the underlying pandas dataframes). This will then automatically invalidate the respective period in the deprecated version, as intended.


ahiguera-mx:
Hi @a1exndr
Thanks for putting this together and thanks @WenzDaniel for the quick review, one thing would be very useful is for every check we  add a comment to the specific condition we are checking with a link to the corresponding test in `test_corrections.py` then everything would be easier to follow

WenzDaniel:
I think it is worth to discuss this on Thursday with a broader audience. I do not like the loophole, but I see that this is currently the only way how we can fix things.

jmosbacher:
I suggest to merge this PR. PR #550 is compatible with this client so we can merge this to fix some urgent bugs and also allow a very smooth transition to a new cmt client when its ready.

WenzDaniel (approved):
I am more concerned about the loop holes we add with this, but it is a Team A call I think.

jmosbacher:
Indeed, this was a major concern for me too but as Alex pointed out the current logic has a much larger loop hole (I think "loop gate" was the exact terminology). @a1exndr do you think it would be viable to implement the solution we discussed with the fill value before this merge or would you prefer to merge and add that fix later?

a1exndr:
I totally agree that this fix is far from perfect. I'll try to implement the alternative fix discussed during the workshop, but I won't have the time to do so properly in the next days. So we can either merge this PR and patch most of the bugs in the current implementation with it or wait a bit longer until the alternative fix is in place, but this leaves the current "loop gate" open for some time (and might again leave to forbidden changes of online values like we recently had) - your call.

jmosbacher:
Its unfortunate that the thursday meeting was canceled as I would prefer to give people a chance to object before merging but I think if @ahiguera-mx @JoranAngevaare and @ershockley are all ok with it I would just merge and then add the patch when you have time.

PR #565, opened 2021-10-30
Add capability for building summed waveform over channel subset

DCichon:
This PR aims to introduce the capability of building a separate summed waveform over a subset of channels only (see also the version by @petergaemers in #482 which has been made before changes were introduced in #486 that broke compatibility). At the moment, both top and bottom array waveforms are built for testing purposes. Only one of these will remain as soon as real data testing concludes in order to reduce data usage. Also, parameter and variable names are open to changes to make them more generic.

### Later edits by Joran
The current implementation from Dominic doesn't support not having the `data_top` field in the numba functions he changed here. So, in order for the defaults to work, you need this field.

Now, imagine that you DON'T want to store top/bottom wf because you run out of storage / it's too intense on your memory / whatever / there is only a top array (in the case of our HE channels). You can do that in two ways:

1.  Either change the numba code to also compile on a dataformat that does not have this field. I considered first doing this but I figured that it would result in so much code duplication that I decided against it. Another option was to start returning the result for the top array separately and leave it to the user to add it to the field in peaklets. This was an option but mixing functions that write half of the results in place and the other by returning was in my oppinion less clear.
2.   Just add the field data_top in the peaklets that enter this part of the numba code. If you desire not to store it, just remove it from the dtype prior to saving it to disk. If this was the default - it's not particularly elegant as you are dragging an extra field along that you don't have to and you also require a second copy step. However, this is not the default and only used for our _he-data.

Let's have a chat otherwise if that is useful. I wrote it now down explicitly for bookkeeping


strax/processing/peak_merging.py
```diff
@@ -159,6 +173,10 @@ def add_lone_hits(peaks, lone_hits, to_pe):
...
         p['area'] += lh_area
 
-        # Add lone hit as delta pulse to waveform:
+        # Add lone hit as delta pulse to waveforms:
         index = (p['time'] - lh_i['time'])//p['dt']
         p['data'][index] += lh_area
+
+        if n_top_channels > 0:
+            if lh_i['channel'] < n_top_channels:
+                p['data_top'][index] += lh_area
```
WenzDaniel, line 182: @terliuk this is the line which adds the lone hit information to the top waveform.

strax/processing/peak_merging.py
```diff
@@ -160,6 +174,10 @@ def add_lone_hits(peaks, lone_hits, to_pe):
         p['area'] += lh_area
         p['area_per_channel'][lh_i['channel']] += lh_area
 
-        # Add lone hit as delta pulse to waveform:
+        # Add lone hit as delta pulse to waveforms:
         index = (p['time'] - lh_i['time'])//p['dt']
```
FaroutYLq, line 178: This line is wrong. Effectively it will lead to a mirrored relative timestamp inside a merged S2 peak. Let's do one thing each time and fix this in another PR. It should be 
```index = ( lh_i['time']-p['time'])//p['dt']```

**Review by FaroutYLq — APPROVED** [review]
Besides the bug in `strax/processing/peak_merging.py`. This worths another PR to fix. Everything else has been tested and I am happy with them.

**Review by FaroutYLq — APPROVED** [review]

File: strax/dtypes.py [hunk 3]
```diff
@@ -175,14 +175,14 @@ def hitlet_with_data_dtype(n_samples=2):
     return dtype + additional_fields
 
 
-def peak_dtype(n_channels=100, n_sum_wv_samples=200, n_widths=11):
+def peak_dtype(n_channels=100, n_sum_wv_samples=200, n_widths=11, digitize_top=True):
```
jmosbacher, line 178: Shouldnt `digitize_top` default to False so that we preserve previous behavior by default and maintain backward compatibility? or is back compatibility impossible?

terliuk: I was wondering the same and I think that our lineages will not let us do that easily (do I got it right, @WenzDaniel ?).

JoranAngevaare: Backward compatibility is not assured as you need the `data_top` field in the array. Otherwise numba doesn't compile.

We could get around this but it would require doubling the numba code or use numba generated jit to get it working. 

Hence, in https://github.com/XENONnT/straxen/pull/1099 I actually worked around this by just dumping the `data_top` field just prior to returning the data.

jmosbacher: So should we just remove the option if it doesnt work when its set to false?

JoranAngevaare: No, because in https://github.com/XENONnT/straxen/pull/1099, I do this option to remove the `data_top`-field to save storage and RAM?

JoranAngevaare: okay, let me break it down.

So the current implementation from Dominic doesn't support not having this field in the numba functions he changed here. So, in order for the defaults to work, you need this field. 

Now, imagine that you DON'T want to store top/bottom wf because you run out of storage / it's too intense on your memory / whatever / there is only a top array (in the case of our HE channels). You can do that in two ways: 
1. either change the numba code to also compile on a dataformat that does not have this field. I considered first doing this but I figured that it would result in so much code duplication that I decided against it. Another option was to start returning the result for the top array separately and leave it to the user to add it to the field in peaklets. This was an option but mixing functions that write half of the results in place and the other by returning was in my oppinion less clear.
2. Just add the field `data_top` in the peaklets that enter this part of the numba code. If you desire not to store it, just remove it from the dtype prior to saving it to disk. If this was the default - it's not particularly elegant as you are dragging an extra field along that you don't have to and you also require a second copy step. However, this is not the default and only used for our `_he`-data.

Let's have a chat otherwise if that is useful. I wrote it now down explicitly for bookkeeping

jmosbacher: Ok, if its easier this way then i am fine with that.

JoranAngevaare:
I also tested that the data produced yields the same for:
1. current master
2. the version that Lanqing tested
3. the updates that I made to that later

The notebook for the test is here:
https://github.com/XENONnT/analysiscode/pull/98

PR #567, opened 2021-10-31
fail for - run_ids

JoranAngevaare:
For WFSim, sometimes users use a run-naming scheme which relies on `'-'`, for example `'run-0'`. This sounds quite intuative and there is also nothing clearly wrong with choosing such a name. However, this leads to very strange results with some of the Strorage frontends, which do not understand folders with too many `'-'`s. 

So let's fail when a user tries using a run_id with `-`. I think we can easily live without this flexibility.


WenzDaniel (approved):
Looks good. Thanks

PR #569, opened 2021-10-31
Infer type from default option

### LINKED ISSUE #568: Option type is never inferred from default value
The Option class init overrides the type parameter before it checks whether its OMITTED, resulting in the code that inferrs the type from the default value never to run
[the offending line](https://github.com/AxFoundation/strax/blob/1f4d22edb65c3d70b3d173d7fee24d403c2b5e0a/strax/config.py#L113)

```python
        type = builtins.type # this overrides the type parameter
        if sum([self.default is not OMITTED,
                self.default_factory is not OMITTED,
                self.default_by_run is not OMITTED]) > 1:
            raise RuntimeError(f"Tried to specify more than one default "
                               f"for option {self.name}.")

        if type is OMITTED and default is not OMITTED: # will never evaluate to True since type is now set to builtins.type
            self.type = type(default)
```

**Expected behavior**
type is expected to be inferred from the default value when its not OMITTED

jmosbacher:
Should fix #568. Overriding of type parameter by builtins.type is removed and some logic is added to handle numpy types correctly.  Explicit type of the `apply_data_function` context option is set since it can handle iterables and functions but the default value is a tuple.

This should me merged with care since it may result in breaking existing code if users were setting options to values with a different type as the default value. We may want to switch the validation error to a warning in cases where the type was inferred from a default value.


JoranAngevaare (approved):
Thanks Yossi, from my side it looks good!

_codefactor and forks are such a mess_

PR #587, opened 2021-11-08
infer_dtype rename and docstring

jmosbacher:
This is an addendum to #586, created to add a docstring but while doping that I realized the name was a bit misleading. I think infer_type is more accurate.


JoranAngevaare:
Thanks looks good!

JoranAngevaare (approved):

PR #588, opened 2021-11-08
Plugin log property

jmosbacher:
OverlapWindowPlugin has a log attribute set in its init function. Not sure why there is a need for logs to be accessed via self but I think this is a better way of implementing that in a way that does not impede serialization.


JoranAngevaare:
@jmosbacher , nice idea, I think this is good, any reason why it's still a draft?

JoranAngevaare:
@jmosbacher I'll just merge this, I don't see a reason why not

JoranAngevaare (approved):

PR #590, opened 2021-11-18
Add funcs for getting stored source

JoranAngevaare:
Sometimes you want to process `event_info` but don't know if you have all the required datatypes that are needed for making that data. Rather than doing a try for except, you might from the start want to know if that data can be computed, and where you need to start processing from.

In this PR we add just that.

Look at the target you want to compute, if it's not stored, check if the dependencies are stored. If finally you reach a plugin that is either not stored or not depending on anything (`raw_records`), return `None`, indicating that you cannot make this data.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1655,13 +1655,19 @@ class Context:
                     f'Trying to write {data_key} to {t_sf} which already exists, '
                     'do you have two storage frontends writing to the same place?')
 
-    def get_source(self, run_id: str, target: str) -> ty.Union[set, None]:
+    def get_source(self,
+                   run_id: str,
+                   target: str,
+                   check_forbidden: bool = True,
+                   ) -> ty.Union[set, None]:
         """
         For a given run_id and target get the stored bases where we can
         start processing from, if no base is available, return None.
 
         :param run_id: run_id
         :param target:  target
+        :param check_forbidden: Check that we are not requesting to make
+            a plugin that is forbidden by the context to be created.
         :return: set of plugin names that are needed to start processing
             from and are needed in order to build this target.
         """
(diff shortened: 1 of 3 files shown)


strax/context.py
```diff
@@ -1656,6 +1655,52 @@ def copy_to_frontend(self,
...
+        if self.is_stored(run_id, target):
+            return {target}
+
+        deps = strax.to_str_tuple(self._plugin_class_registry[target].depends_on)
+        if not deps:
+            return None
+
+        forbidden = strax.to_str_tuple(self.context_config['forbid_creation_of'])
+        if check_forbidden and target in forbidden:
+            return None
```
WenzDaniel, line 1683: Maybe be more explicit in this case and raise an error or make a warning. Like you have further down.

WenzDaniel (approved):
Nice and helpful feature thanks a lot. Just one minor comment.

strax/context.py
```diff
@@ -1656,6 +1655,52 @@ def copy_to_frontend(self,
...
+
+        stored_sources = set()
+        for dep in deps:
+            if self.is_stored(run_id, dep):
```
jmosbacher, line 1687: isnt this redundent? shouldnt the recursive call to self.get_source check if its stored or not?

jmosbacher (approved):
Very nice idea, this will probably be used a lot. Just left one small comment (may be possible to make the function a bit smaller==simpler) but even that is not very important.

strax/context.py
JoranAngevaare: Ah great point!

strax/context.py
JoranAngevaare: good point!

PR #591, opened 2021-11-18
Fix buffer issue in highest density regions, adds tests

WenzDaniel:
Solves https://github.com/XENONnT/straxen/issues/765 and adds additional test. The issue occurs if a single signals is divided into many small intervals e.g. bipolar noise. In this PR we change the default behavior from raising an error into returning np.nans in that case.


JoranAngevaare:
@WenzDaniel should we bump-version the associated nv/mv processing?

WenzDaniel:
I was already wondering the same, in principle all runs which need this change were not processed anyhow. Other then that nothing was changed,

JoranAngevaare (approved):
okay, sure. I suspect so from the changes but haven't checked in detail.

This is an "approve light" as I did not go through the code very minutiously, please merge if you are fully confident 😉

PR #593, opened 2021-11-19
Add mode and tags to superrun.

WenzDaniel:
In this pull request we add some additional features to superruns. When defining a new superrun all tags and run_modes of the corresponding subruns are appended to the rund_docs. Some additional changes:

* If more than one storage frontend contains run_metadata select_runs returns a dataframe with duplicated indices. 
* Normal rundocs are time zone unaware, hence we are now stripping the timezone information from the superruns to have a consistent datatype
* Normal rundocs compute their livetime as timedelta between end and start. However, timedelta is not json serializable, hence superruns store the livetime as seconds float. We convert this now into a timedelta in select_runs.  
* Livetime was changed from ns to s to be more consistent with normal runs.
* Added source field to superruns

Last but not least I added tests for the new features.

The livetime change means, that previously superruns will raise the following error: 

```
~/scratch-midway2/programs/anaconda3/envs/strax_main_py38/lib/python3.8/site-packages/pandas/core/arrays/timedeltas.py in objects_to_td64ns(data, unit, errors)
   1065     values = np.array(data, dtype=np.object_, copy=False)
   1066 
-> 1067     result = array_to_timedelta64(values, unit=unit, errors=errors)
   1068     return result.view("timedelta64[ns]")
   1069 

pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.array_to_timedelta64()

pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.convert_to_timedelta64()

pandas/_libs/tslibs/timedeltas.pyx in pandas._libs.tslibs.timedeltas.delta_to_nanoseconds()

OverflowError: Python int too large to convert to C long
```

**Update:**

I piggybaged a small change in multi-runs which affects superruns as well as normal runs. `multi-runs` usually adds a `run_id` field as a string to the resulting array. While this can be quite handy it can also let your memory explode in case of large data-sets. I made the appending of the run_id field optional and switched it off for superruns.


tests/test_context.py
```diff
@@ -48,6 +48,19 @@ def nothing(data, r, t):
...
+def test_byte_strings_as_run_id():
+    with tempfile.TemporaryDirectory() as temp_dir:
+        st = strax.Context(storage=strax.DataDirectory(temp_dir,
+                                                       deep_scan=True),
+                           register=[Records])
+
+
+        records_bytes = st.get_array(b'0', 'records')
+        records = st.get_array('0', 'records')
+        assert np.all(records_bytes == records)
```
JoranAngevaare, line 60: 👍 thanks for checking. The only place I'm a bit concerned is for making masks based on run_id since:
```python
 b'1' == '1' -> False
```
Hence I opened issue #598 as I felt I might have to think a bit more about it

WenzDaniel:
Yes, this is also what the failing test is complaining about.

JoranAngevaare:
> Yes, this is also what the failing test is complaining about.

Maybe, to prevent things from breaking, make it an option and warn if one is loading more than (pick your favorite number) 10_000 entries with a warning like:
```python
if len(ids) > 10_000:
  UserWarning(f'You are using {ids.nbytes/10**6} MB for the run_id field. Reduce by 400% by passing "run_id_raw=True" see github.com/AxFoundation/strax/issues/598')
```

WenzDaniel:
Sounds like a good compromise.

WenzDaniel:
@JoranAngevaare can we go ahead and merge? The only thing failing is the complex method warning.

strax/utils.py
```diff
@@ -474,6 +473,24 @@ def multi_run(exec_function, run_ids, *args,
...
+             )
+        run_id_numpy = run_id_numpy.astype('S')  # Use byte string to reduce memory usage.
+
+    # Get from kwargs whether output should contain a run_id field.
+    # In case we have a multi-runs with superruns we should skip adding
+    # run_ids and sorting according run_id does not make sense.
+    if 'add_run_id_field' in kwargs:
+        add_run_id_field = kwargs.pop('add_run_id_field')
+    else:
+        add_run_id_field = not _is_superrun
```
JoranAngevaare, line 492:
```suggestion
    # Get from kwargs whether output should contain a run_id field.
    # In case we have a multi-runs with superruns we should skip adding
    # run_ids and sorting according run_id does not make sense.
    if 'add_run_id_field' in kwargs:
        add_run_id_field = kwargs.pop('add_run_id_field')
    else:
        add_run_id_field = not _is_superrun
    if add_run_id_field and not run_id_as_bytes and len(run_id_numpy) > 70:
        warn('You are asking for more than 70 runs at a time with add_run_id_field=True. '
             'Changing run_id data_type from string to bytes would reduce memory consumption. '
             'Do so with passing "run_id_as_bytes=True" . When you do, '
             'please note that "run_id" != b"run_id"! You can convert a byte string back to '
             'a normal string via b"byte_string".decode("utf-8"). '
             )
    elif run_id_as_bytes or not add_run_id_field:
        run_id_numpy = run_id_numpy.astype('S')  # Use byte string to reduce memory usage.
```


Ah sorry I wasn't clear. I think this is a nice feature to have, but I would propose to make it optional with it's default turned off (just to prevent scripts from breaking). 

If we do this and add an option `run_id_as_bytes` like we did for `add_run_id_field`, we can have people use it if they want without breaking their scripts.

JoranAngevaare (approved):
Thanks Daniel, we discussed this before and looking over the changes it seems to make a lot of sense, thanks!

I did not really test it though but the tests look convincing.

I will approve, but I hope you agree on the one remaining comment I would have:

strax/utils.py
```diff
@@ -529,7 +551,12 @@ def multi_run(exec_function, run_ids, *args,
             pbar.close()
             return None
 
-        final_result = [final_result[ind] for ind in np.argsort(run_id_output)]
+        if add_run_id_field:
```
dachengx, line 554: Hey @WenzDaniel , why this change is needed?

dachengx: It is resulting in an error: https://github.com/AxFoundation/strax/issues/783.

PR #596, opened 2021-11-24
Fix online monitor bug for only md stored

JoranAngevaare:
Sometimes, we get errors that data cannot be loaded from the MongoFrontend when only the metadata is written. This PR adds a test that was failing (before making the changes in the MongoFrontend).

Check that a `chunk` of data is written rather than only the `metadata` document

Please see the tests


strax/storage/mongo.py
```diff
@@ -176,16 +176,24 @@ def __init__(self, uri, database, col_name=None, *args, **kwargs):
...
-        query = backend_key_to_query(str(key))
-        if self.db[self.col_name].count_documents(query):
-            self.log.debug(f"{key} is in cache.")
+        # Should have at least one non-metadata chunk, otherwise there
+        # is no data to load
+        query = {**backend_key_to_query(str(key)),
+                 'provides_meta': False
+                 }
+        if self.collection.count_documents(query):
+            self.log.debug(f"{key} is in database.")
```
JoranAngevaare, line 194: This is the fix

tests/test_mongo_frontend.py
```diff
@@ -135,16 +128,124 @@ def test_interrupt_iterator(self):
...
+
+        # Now check that both frontends understand there is no data (even when
+        # allow_incomplete is set)
+        stored_in_st = self.is_stored_in_mongo
+        stored_in_incomplete_st = self.stored_in_context(st_incomplete_allowed)
+        if stored_in_st or stored_in_incomplete_st:
+            raise RuntimeError(f'Only metadata written and should NOT stored in st '
+                               f'({stored_in_st}) and st_incomplete ('
+                               f'{stored_in_incomplete_st}) '
+                               f'\n{self.chunk_summary}')
```
JoranAngevaare, line 217: This is the check that failed before and is now fixed

strax/storage/mongo.py
```diff
@@ -85,10 +85,10 @@ def _read_chunk(self, backend_key, chunk_info, dtype, compressor):
     def _saver(self, key, metadata, **kwargs):
         """See strax.Backend"""
         # Use the key to make a collection otherwise, use the backend-key
-        col = self.db[self.col_name if self.col_name is not None else str(key)]
```
JoranAngevaare, line 88: This option did not work anyway and is just confusing

strax/storage/mongo.py
```diff
@@ -160,7 +160,7 @@ def _clean_first_key_from_registry(self):
 class MongoFrontend(StorageFrontend):
     """MongoDB storage frontend"""
 
-    def __init__(self, uri, database, col_name=None, *args, **kwargs):
+    def __init__(self, uri, database, col_name, *args, **kwargs):
```
darrylmasson, line 163: No type hints?

strax/storage/mongo.py
darrylmasson: Alternately the same query would work but we check that `count_documents()` returns at least 2

darrylmasson (approved):

strax/storage/mongo.py
jmosbacher: So if this didnt work im assuming that means there is no existing code this can break?

strax/storage/mongo.py
```diff
@@ -176,16 +176,24 @@ def __init__(self, uri, database, col_name=None, *args, **kwargs):
...
+        # Should have at least one non-metadata chunk, otherwise there
+        # is no data to load
+        query = {**backend_key_to_query(str(key)),
+                 'provides_meta': False
+                 }
+        if self.collection.count_documents(query):
+            self.log.debug(f"{key} is in database.")
             return self.backends[0].__class__.__name__, str(key)
-        self.log.debug(f"{key} is NOT in cache.")
+        self.log.debug(f"{key} is NOT in database.")
```
jmosbacher, line 196: I really hate having the log as an instance attribute because it prevents pickling but OK maybe that's just personal taste.

jmosbacher (approved):
Looks good.

strax/storage/mongo.py
JoranAngevaare: Right, this is true and actually what I did first. Due to another bug it did not seem to work.

strax/storage/mongo.py
JoranAngevaare: No indeed, I actually tried to write a test for it but this old logic. But apart from this one line has abandoned this many collection option. See e.g. this line that is assuming there is only one collection:
https://github.com/AxFoundation/strax/blob/22452f2c3f93ae309115f30ce24e50d04c8c8b88/strax/storage/mongo.py#L193

So yeah, I guess it's impossible to break a formerly unused and broken option.

strax/storage/mongo.py
JoranAngevaare: Sure, let me add some

strax/storage/mongo.py
JoranAngevaare: It's indeed something that can be annoying because e.g. multiprocessing needs pickling, but this is probably not inside the scope of this PR

WenzDaniel:
Since approved I merge to get ready for release.

PR #599, opened 2021-11-29
speedup `get_source` with lookupdict

JoranAngevaare:
Speedup https://github.com/AxFoundation/strax/pull/590 with ~40x. We did a lot of checks twice since we went all the way to the lowest dependency for each of the dependencies. This means that e.g. if you have a plugin depending on `records_a` and `records_b` which both depend on `raw_records` you would check `raw_records` twice.

Rather than using a set, we can also use a lookup dict to keep track of what we have seen. This speeds up the function by ~40x.

## Additional benefits
 - Allow multiple targets `st.get_source('029000', ('cuts_basic', 'event_info'))`
 - Fix getting dependencies from CutMerger plugins (`depends_on` as class property)
 - Informative function `stored_dependencies`

### Example 1
What sources do we need to start for making `'cuts_basic' and 'event_info'`:
```python
st = cutax.xenonnt_v5()
st.get_source('029000', ('cuts_basic', 'event_info'))
# gives
{'cut_daq_veto', 'event_basics', 'event_pattern_fit', 'event_positions'}
``` 
This means `'cut_daq_veto', 'event_basics', 'event_pattern_fit', 'event_positions'` are stored and we can compute `'cuts_basic'` and `'event_info'` for run '029000'.

### Example 2
or get a list of plugins that are needed and stored:
```python
st.stored_dependencies('029000', 'cuts_basic')
# gives
{'cuts_basic': False,
 'cut_daq_veto': True,
 'event_info': False,
 'corrected_areas': False,
 ... }
```
Meaning we still need to make `'event_info'` and `'corrected_areas'` to compute `cuts_basic` whereas we already have `cut_daq_veto` computed.


ershockley (approved):
Works very well, thanks @JoranAngevaare!

strax/context.py
```diff
@@ -1666,37 +1666,91 @@ def get_source(self,
...
             return None
 
+    def stored_dependencies(self,
+                            run_id: str,
+                            target: ty.Union[str, list, tuple],
+                            check_forbidden: bool = True,
+                            _targets_stored: ty.Optional[dict] = None,
+                            ) -> ty.Optional[dict]:
+        """
+        For a given run_id and target(s) get a dictionary of all the datatypes taht
```
WenzDaniel, line 1694:
```suggestion
        For a given run_id and target(s) get a dictionary of all the datatypes that:
```

WenzDaniel (approved):
Just a typo I will commit and merge.

PR #600, opened 2021-11-30
Remove config warning

jmosbacher:
As @JoranAngevaare has pointed out, this warning has been annoying for many and may not be worth the trouble.
This PR removes the warning and just sets infer_type=False to keep the behavior people are used to even if it was technically a bug, at has become a feature :)


JoranAngevaare (approved):
Thanks @jmosbacher , yeah, I think this is for the better.

PR #613, opened 2021-12-14
Make truly HDR

WenzDaniel:
Small PR which fixes the definition of the highest density region.  The difference is illustrated in the image below:

example

In the previous version we computed the area wrongly in case one is truly  interested in the HDR.  So far this function was only used to get a width estimate for the hitlet data_type. No need in straxen needed. The behavior of the already existing function was changed such that they behave consistently. 

I also added a test which shows that the HDR of the 68.27% a normal distribution corresponds to the expected +/- sigma.

### DIFF
--- a/strax/processing/statistics.py
+++ b/strax/processing/statistics.py
@@ -4,13 +4,13 @@ import numba
 import strax
 export, __all__ = strax.exporter()
 
+
 @export
 @numba.njit(cache=True)
-def highest_density_region(data, fractions_desired, _buffer_size=10):
+def highest_density_region(data, fractions_desired, only_upper_part=False, _buffer_size=10):
     """
     Computes for a given sampled distribution the highest density region
     of the desired fractions.
-
     Does not assume anything on the normalisation of the data.
 
     :param data: Sampled distribution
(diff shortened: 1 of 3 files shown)


strax/processing/statistics.py
```diff
@@ -4,25 +4,25 @@
...
     Does not assume anything on the normalisation of the data.
 
     :param data: Sampled distribution
     :param fractions_desired: numpy.array Area/probability for which
         the hdr should be computed.
     :param _buffer_size: Size of the result buffer. The size is
         equivalent to the maximal number of allowed intervals.
-
+    :param only_upper_part: Boolean, if true only computes
+        area/probability between maximum and current hight.
```
JoranAngevaare, line 22:
```suggestion
        area/probability between maximum and current height.
```

strax/processing/statistics.py
```diff
@@ -50,6 +50,7 @@ def highest_density_region(data, fractions_desired, _buffer_size=10):
             continue
 
         lowest_sample_seen = data[max_to_min[j]]
+        lowest_sample_seen *= only_upper_part
```
JoranAngevaare, line 53: maybe an explicit int conversion?

JoranAngevaare (approved):
good catch

JoranAngevaare:
will we have a version bump in straxen for this?

PR #614, opened 2021-12-16
Remove tight coincidence channel from data_type

WenzDaniel:
Removes tight coincidence channel from data_type as analysts find it confusing.


JoranAngevaare (approved):
Fully agree, I never so much liked it to be honest

PR #616, opened 2021-12-22
Don't cache `hitlet_properties`

JoranAngevaare:
I today spend again a few hours trying to chase a `ZeroDivisionError` in this code only to have it magically disappear again. I am afraid that with the caching within caching of functions, numba might not be always as reliable as we might hope. I also can't spot in the code where it might be coming from so maybe it's just Daniels nemesis showing up again to wish us a nice christmas 🧙 ?

I propose to use caching a bit less. Hopefully we don't spend again chasing these errors.

While trying to find the cause of the issue, I played around with the caching of this function and the error magically disappeared 🪄. Now I wanted to see what the effect of caching this function was, and it turned out to be none. So I propose to disable tich caching. Please see my test below so you can repeat it if you want.

Should have included this:
```python
Working on DESKTOP-F4PI41P.localdomain with the following versions and installation paths:
python	v3.9.5	(default, Jun  4 2021, 12:28:51) [GCC 7.5.0]
strax	v1.1.4	/mnt/c/google_drive/PhD-master/ubuntu-storage/ubuntu-windows/software/strax/strax	git branch:master | 415cc25
straxen	v1.2.0	/mnt/c/google_drive/PhD-master/ubuntu-storage/ubuntu-windows/software/straxen/straxen	git branch:master | 2927ac8
cutax	v1.1.1	/mnt/c/google_drive/PhD-master/ubuntu-storage/ubuntu-windows/software/cutax/cutax	git branch:master | 1eae37d
```
See https://gist.github.com/JoranAngevaare/44f98c973cad64b10e3911f1abbb3c7d


WenzDaniel (approved):
Okay if caching does not have any impact on the speed we can drop it.

Issue #566, opened 2021-10-30
Add remoteness attribute to SF


JoranAngevaare:
It would be nice if SFs had a remoteness attribute, and that data availability is checked based on the ordering of the remoteness level

JoranAngevaare:
Fixed in #617

PR #617, opened 2021-12-23
StorageFrontend remoteness attribute and test

JoranAngevaare:
The issue is described here:
https://github.com/AxFoundation/strax/issues/566

Rather than looping over the frontends in the order they were specified by the user, loop over the frontends in order of `StorageType`

See some examples in the test


strax/context.py
```diff
@@ -679,17 +679,29 @@ def _find_options(self):
...
+    @property
+    def _sorted_storage(self) -> ty.List[strax.StorageFrontend]:
+        """
+        Simple ordering of the storage frontends on the fly when e.g.
+        looking for data. This allows us to use the simple self.storage
+        as a simple list without asking users to keep any particular
+        order in mind. Return the fastest first and try loading from it
+        """
+        storage_sfs = self.storage
+        storage_sfs.sort(key=lambda x: x.storage_type)
```
jmosbacher, line 691:
```suggestion
        storage_sfs = sorted(self.storage, key=lambda x: x.storage_type)
```
I think you want to keep original list since order is used in `copy_to_frontend` right? I think the way it is would sort in place the original list no?

strax/context.py
```diff
@@ -679,17 +679,29 @@ def _find_options(self):
...
         :param time_range: 2-length arraylike of (start, exclusive end) of row
-        numbers to get. Default is None, which means get the entire run.
+            numbers to get. Default is None, which means get the entire run.
         :param chunk_number: number of the chunk for data specified by
-        strax.DataKey. This chunck is loaded exclusively.
+            strax.DataKey. This chunck is loaded exclusively.
         :return: partial object
         """
-        for sb_i, sf in enumerate(self.storage):
+        for _, sf in enumerate(self._sorted_storage):
```
jmosbacher, line 704:
```suggestion
        for sf in self._sorted_storage:
```
double jeopardy!

jmosbacher (changes requested):
Very nice! I think this is a great idea. just two small comments.

strax/context.py
JoranAngevaare: you are absolutely right!

strax/context.py
JoranAngevaare: Nice!

JoranAngevaare:
Thanks a lot Yossi for the super quick and good review! I've implemented them both 👍

jmosbacher (approved):

PR #630, opened 2022-01-10
py3.10

JoranAngevaare:
Add py3.10 into testing


jmosbacher (approved):
Nice, cant wait to start using pattern matching ;)

Issue #504, opened 2021-08-01
Make save_when in plugin optional per datatype.


JoranAngevaare:
Currently, we setting the save_when property of an entire plugin to a given `strax.SaveWhen`. However, for storage purposes, it might be nice if we can specify which items we want to store and which we don't want to store, similar to the optional rechunking is configurable per datatype provided by a plugin.

JoranAngevaare:
Problem with this is that also loading should recognize this as we usually raise errors if somethings is already computed

PR #631, opened 2022-01-11
Extend save when

WenzDaniel:
Extends SaveWhen mechanic of plugins such that we can specify the SaveWhen type per provided data_type. See [also ](https://github.com/AxFoundation/strax/issues/504). This is especially handy for plugins like `PulseProcessing` for which we would like to store pulse_counts for monitoring purposes but not `records` given its size.

This PR also solves an issue for superruns. Making the superrun of a multi-output plugin not finishing if one of the outputs is not stored. (Same bug also leads to re-simulation of raw data if superrun is created for raw-records?)

In addition I also fixed the long outstanding bug that multi-runs would not properly raise errors which happen in one of the threads. 

In this PR we change `save_when` such that it can be either specified as a dictionary or as an integer as it was done before. In the latter case we convert `save_when` into a dictionary during initialization of the class.  In this way we keep a backwards compatibility to the plugins in straxen. Depending on the specified saving types `get_components` will add the corresponding savers to the processor components.

<s>Since two checks in our plugin class require some knowledge about the saving type of the plugin, `get_components` also specifies a new class attribute called `_save_when` depending on the targets `save_when`-setting.  </s>

**Update:** After a quick chat with Joran we decided to use max(save_when_per_provides) instead of _save_when. 

Here is a MWE with `PulseProcessing `for a standard context:

```
p = st._plugin_class_registry['records']
p.save_when
# returns: <SaveWhen.TARGET: 2>
```
When initializing the plugin we set SaveWhen for each provides: 
```
p = st.get_single_plugin('009104', 'records')
p.save_when
# returns: {'records': <SaveWhen.TARGET: 2>,
 'veto_regions': <SaveWhen.TARGET: 2>,
 'pulse_counts': <SaveWhen.TARGET: 2>}
```
Now we can get the saver components for example for:
```
components = st.get_components('009104', 'pulse_counts')
components.savers
# returns: {'pulse_counts': [<strax.storage.files.FileSaver at 0x7f6347cf2a30>]}
components.plugins['veto_regions']._save_when
# returns: <SaveWhen.TARGET: 2>
```

Now let us repeat but adjust the SaveWhen settings of PulseProcessing:

```
p = st._plugin_class_registry['records']
p.save_when = {'records': strax.SaveWhen.TARGET,
               'veto_regions': strax.SaveWhen.NEVER,
               'pulse_counts': strax.SaveWhen.ALWAYS,}

p = st.get_single_plugin('009104', 'records')
p.save_when
#returns: {'records': <SaveWhen.TARGET: 2>,
 'veto_regions': <SaveWhen.NEVER: 0>,
 'pulse_counts': <SaveWhen.ALWAYS: 3>}
```

and get components gives:

```
components = st.get_components('009104', ('records',))
components.savers
# returns: {'pulse_counts': [<strax.storage.files.FileSaver at 0x7f6347cf2df0>],
 'records': [<strax.storage.files.FileSaver at 0x7f634f04aa60>]}
```

**Open points:**

- [x] Add tests
- [x] Check MergeOnlyPlugin plugin e.g. with peaks
- [x] Will try to save already existing data if one of the provides is SaveWhen always -> Drop saver if loader exists.


WenzDaniel:
For tracking https://github.com/AxFoundation/strax/issues/633

JoranAngevaare:
@WenzDaniel you'll ask for a review when this is ready right?

WenzDaniel:
Ah yes sorry. It is ready

WenzDaniel:
Min also confirmed that it solves the issue with wfsim.

strax/context.py
```diff
@@ -832,31 +828,29 @@ def concat_loader(*args, **kwargs):
                 to_compute[target_i] = target_plugin
                 for dep_d in target_plugin.depends_on:
                     check_cache(dep_d)
-
+            
+            if self.context_config['storage_converter']:
+                warnings.warn('The storage converter mode will be replaced by "copy_to_frontend" soon. '
+                              'It well be removed in one of the future releases. Please let us know if '
```
JoranAngevaare, line 834:
```suggestion
                              'It will be removed in one of the future releases. Please let us know if '
```

strax/context.py
```diff
@@ -948,6 +926,70 @@ def concat_loader(*args, **kwargs):
             loader_plugins=loader_plugins,
             savers=savers,
             targets=strax.to_str_tuple(final_plugin))
+    
+    def _add_saver(self, savers, run_id, d_to_save, target_plugin, _is_superrun, loading_this_data):
+        key = strax.DataKey(run_id, d_to_save, target_plugin.lineage)
```
JoranAngevaare, line 931: Could you make a docstring / type hints since you also confused yourself with the meaning of target, target_i and target plugin:
```python
key = strax.DataKey(run_id, d_to_save)
```

strax/plugin.py
```diff
@@ -116,6 +116,12 @@ def __init__(self):
             del compute_pars[compute_pars.index('start')]
             del compute_pars[compute_pars.index('end')]
 
+        if hasattr(self, 'provides') and not isinstance(self.save_when, dict):
```
JoranAngevaare, line 119: maybe also add immutabledict?
```python
>>> import immutabledict
>>> isinstance(immutabledict.immutabledict(), dict)
False
```

strax/utils.py
```diff
@@ -522,17 +522,20 @@ def multi_run(exec_function, run_ids, *args,
...
                 log.debug(f'Done with run_id: {_run_id} ' 
                           f'and {len(run_id_numpy)-tasks_done} are left.')
                 pbar.update(1)
+                if f.exception() is not None:
+                    raise f.exception()
+                
                 if throw_away_result:
                     continue
-
+
```
JoranAngevaare, line 536:
```suggestion

```

JoranAngevaare (commented):
Thanks Daniel, also for the much needed refactor!

I'd have one more request and that if you could maybe also add a test for the time-selection for multi-output plugins with several save_when levels?

strax/context.py
JoranAngevaare: I wonder why this is not covered by the tests 🤔 , do you see an easy way we would fail here?

JoranAngevaare: whoops, I see I was actually confused, but I think target_plugin is sometimes passed as a string

WenzDaniel: > Could you make a docstring / type hints since you also confused yourself with the meaning of target, target_i and target plugin:

Sure, but I cannot follow for your other two comments. 

> I wonder why this is not covered by the tests

According to coveralls it is is not it? 

> whoops, I see I was actually confused, but I think target_plugin is sometimes passed as a string

What do you mean? At least not during testing otherwise the function would fail.

JoranAngevaare: Hmm okay, yeah as I said I was confused, looking a bit closer I must have been messing up with `target_i` sorry for the jitter ( `get_components` must be one of the more confusing functions in my defense :joy:)

JoranAngevaare: you might by the way get rid of the `run_id` argument in  `_add_saver` by using `plugin.run_id` instead if you want to reduce the number of required arguments

WenzDaniel: > get_components must be one of the more confusing functions in my defense

Yes, absolutely agree on that :D

WenzDaniel:
Okay I addressed all your comments.

JoranAngevaare (approved):
Nice work Daniel!

PR #635, opened 2022-01-23
test straxen for coverage and backward compatibility

JoranAngevaare:
- We could increase coverage by running straxen tests on coveralls
 - We could check if changes in a PR would cause issues for running straxen. A fail here would mean we introduce breaking changes. We could either avoid doing that or allow it and fix it on straxens master 

In this PR, we don't use any of straxens secrets, so we don't have database access

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -61,6 +61,11 @@ jobs:
           TEST_MONGO_URI:  'mongodb://localhost:27017/'
         run: |
               coverage run --source=strax setup.py test -v
+              straxen_dir=../straxen/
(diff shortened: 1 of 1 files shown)


WenzDaniel (approved):
Nice idea, I like it. Is our testing with coveralls limited? In that case we might only run it under certain conditions. E.g. making a new release.

JoranAngevaare:
@WenzDaniel , no there is no time constrain. It's only the (possible) annoyance of having to iterate between strax <-> straxen. Could be a good motivation to make sure things are backward compatible at least one release but can also see this being annoying sometimes.

WenzDaniel:
In that case I would propose to just give it a try.

PR #636, opened 2022-01-24
deprecate py3.6 py3.7

JoranAngevaare:
# Deprecate python 3.6 and python 3.8
Let's make the transition to py>=3.8 such that we can start using things like the walrus operator. See https://docs.python.org/3/whatsnew/3.8.html

### The walrus operator
See this example, it's great
```python
if (n := len(a)) > 10:
    print(f"List is too long ({n} elements, expected <= 10)")
```


WenzDaniel:
Looks good, thanks a lot.

WenzDaniel (approved):

JoranAngevaare:
yeah it's on the agenda

PR #638, opened 2022-01-28
extend search field to also print occourances

JoranAngevaare:
Add a programmatic solution to https://github.com/XENONnT/straxen/issues/235, finding where datafields are used in the code

Inspect all of the plugin functions to see if a field (like `'time'`) is used in any line of the code. If so, return the function that uses it, e.g. `Records.compute`.

```python
python -c "from straxen.contexts import xenonnt; xenonnt().search_field('cs1')"

cs1 is part of corrected_areas (provided by CorrectedAreas)
cs1 is part of event_info (provided by EventInfo)

cs1 is used in CorrectedAreas.infer_dtype
cs1 is used in CorrectedAreas.compute
cs1 is used in EnergyEstimates.compute
cs1 is used in EnergyEstimates.cs1_to_e
```


strax/context.py
```diff
@@ -309,19 +312,79 @@ def deregister_plugins_with_missing_dependencies(self):
...
+        """
+        Find and return which plugin(s) use a given field.
+
+        :param search_string: a field that matches pattern exact
+        :param plugin: plugin where to look for a field
+        :return: list of code occurrences in the form of PLUGIN.FUNCTION
+        """
+        if plugin is None:
+            plugin = list(self._plugin_class_registry.values())
+        if not isinstance(plugin, list):
```
WenzDaniel, line 373:
```suggestion
        if not isinstance(plugin, (list, tuple)):
```

strax/context.py
```diff
@@ -309,19 +312,79 @@ def deregister_plugins_with_missing_dependencies(self):
...
+            plugin = [plugin]
+        result = []
+        for plug in plugin:
+            for attribute_name, class_attribute in plug.__dict__.items():
+                is_function = isinstance(class_attribute, types.FunctionType)
+                if is_function:
+                    for line in inspect.getsource(class_attribute).split('\n'):
+                        if search_string in line:
+                            if plug.__class__.__name__ == 'type':
+                                plug = plug()
```
WenzDaniel, line 383: Can you explain this line and maybe put a comment. I do not understand what you are doing here.

strax/context.py
```diff
@@ -309,19 +312,79 @@ def deregister_plugins_with_missing_dependencies(self):
...
+                     return_matches: bool = False,
+                     ) -> ty.Union[None, ty.Tuple[defaultdict, dict]]:
+        """
+        Find and print which plugin(s) provides a field that matches
+        pattern (fnmatch).
 
-            for field_name in p.dtype_for(d).fields:
+        :param pattern: pattern to match, e.g. 'time' or 'tim*'
+        :param include_code_usage:
+        :param return_matches:
```
WenzDaniel, line 326: You forgot the doc-string

WenzDaniel (approved):
Looks good just two minor comments. It is a nice feature.

strax/context.py
JoranAngevaare: right, so it's a check that you are getting the instance instead of the base class:

```python
class A:
   pass

A.__class__.__name__ # returns type
A().__class__.__name__ # returns A

strax/context.py
JoranAngevaare: ah thanks

JoranAngevaare:
Thanks for the review Daniel!

PR #649, opened 2022-02-12
Added lone hit area to area per channel

WenzDaniel:

Fix bug in merged S2s


zhut19 (approved):

WenzDaniel:
Yes, I was  thinking about a simple unittest too. However, I think it is not so straight forward as you have to trigger peaklet merging. We could check for some test data if `area` and `sim(area_per_channel)` gives the same.

PR #650, opened 2022-02-13
Fix testing from straxen

JoranAngevaare:
If we test with straxen's master, we should also run it's code


JoranAngevaare:
great thanks a lot!

PR #654, opened 2022-02-19
Buffering & rechunking plugin

JoranAngevaare:
This plugin solves several issues with the strax.OverLapWindowPlugin:
 - The OverLapWindowPlugin assumes data sorted by time
 - The OverLapWindowPlugin needs to forced into finding large windows if the data might be sparse and information distributed ammong multiple data entries (on/off of a logic signal)

I believe it can also be a solution to plugins where we don't really care about the chunking paradigm, such as WFsim. With this plugin, you could just instruct the plugin to keep buffering results until there is a suitable place to set the chunk boundary. We'd have to move the place where the input is gathered though (this is currently in [compute](https://github.com/XENONnT/WFSim/blob/7dc4dba9bce0ff6a2f3856fe46981456e35cf70a/wfsim/strax_interface.py))

Tests are still required, but first results show that it's reproducing the OverLapWindowPlugin's results. 

Rather than processing the input plugin straight away, buffer it for some time (if needed, can also be set to 0), then, find gaps in the input buffer, if those are large enough, start writing a chunk. If this is not the case, wait until there is one.


JoranAngevaare:
Stashing this work to a remote branch, needs some more thinking:
https://github.com/JoranAngevaare/strax/tree/stach_buffer_plugin

PR #655, opened 2022-02-21
Allow numpy types in configs for lineage hashing

JoranAngevaare:
- Allow numpy types in configs.
 - Allow immutibledicts and hash as an dict

Otherwise you get these uninformative errors:


JoranAngevaare:
I can't reproduce the problem this PR was supposed to solve

PR #656, opened 2022-02-22
Refactor in preparation for PyMongo 4.0

JoranAngevaare:
Maybe we should really do this, it's causing issues with python3.10:
https://pymongo.readthedocs.io/en/stable/examples/tls.html#python-3-10-incompatibilities-with-tls-ssl-on-mongodb-4-0

https://github.com/XENONnT/straxen/pull/932


jmosbacher:
I agree its probably time :)

PR #657, opened 2022-02-25
Change tempdirs in test

WenzDaniel:

Change the tempdir creation of some of our tests. The new procedure uses a tempdir created by `tempfile.TemporaryDirectory` this method has a weak  reference in its init which ensures that the directory gets deleted if its reference is garbage collected. In this way we can ensure permission denied erros etc. on shared systems like dali. 

See also: 

https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

and

https://stackoverflow.com/questions/50649701/shutil-rmtree-does-not-work-on-tempfile-temporarydirectory

### DIFF
--- a/tests/test_saving.py
+++ b/tests/test_saving.py
@@ -11,16 +11,15 @@ class TestPerRunDefaults(unittest.TestCase):
     def setUp(self):
         self.test_run_id = '0'
         self.target = 'records'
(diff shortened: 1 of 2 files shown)


JoranAngevaare (approved):
sure, looks good

PR #674, opened 2022-03-07
Add tests for lose ends

WenzDaniel:
Fixes a bug in our save_when extension introduced in https://github.com/AxFoundation/strax/pull/631. I forgot to check whether "lose-end" data_types like `pulse_counts` are stored properly after inlining plugins. The issue arises from the fact that we return early in [this line of get_components](https://github.com/AxFoundation/strax/blob/ef37118b1c3302b0af49179bd46629f619b3f98a/strax/context.py#L928) for records.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -920,12 +920,20 @@ class Context:
                 return
  
             # Now we should check whether we meet the saving requirements (Explicit, Target etc.)
+            # In case of the storage converter mode we copy already existing data. So we do not
+            # have to check for the saving requirements here.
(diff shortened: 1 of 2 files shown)


WenzDaniel:
Test failed before in https://github.com/AxFoundation/strax/runs/5449934414?check_suite_focus=true#step:8:92 But should now pass.

JoranAngevaare:
looks good, thanks!

WenzDaniel:
Thanks for spotting.

PR #687, opened 2022-04-11
Allow different chunk size

JoranAngevaare:
Allow different chunk target size from original

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1783,7 +1783,7 @@ class Context:
         if len(strax.to_str_tuple(target)) > 1:
             raise ValueError(
                 'copy_to_frontend only works for a single target at the time')
-        if target_frontend_id is not None or target_frontend_id >= len(self.storage):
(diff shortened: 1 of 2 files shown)


strax/context.py
```diff
@@ -1833,10 +1843,13 @@ def copy_to_frontend(self,
...
 
         if target_compressor is not None:
-            self.log.info(f'Changing compressor from {md["compressor"]} '
-                           f'to {target_compressor}.')
+            self.log.info(f'Changing compressor {md["compressor"]} -> {target_compressor}.')
             md.update({'compressor': target_compressor})
 
+        if rechunk and md["chunk_target_size_mb"] != rechunk_to_mb:
+            self.log.info(f'Changing chunk-size: {md["chunk_target_size_mb"]} -> {rechunk_to_mb}.')
+            md.update({'chunk_target_size_mb': rechunk_to_mb})
```
JoranAngevaare, line 1851: This is actually the only _real_ change, the rest is cosmetic

Issue #688, opened 2022-04-25
`Plugin.__version__` and `Plugin.version` are ambigous


JoranAngevaare:
`Plugin.__version__` is used where `Plugin.version` should be used. Examples:
- [L2021](https://github.com/AxFoundation/strax/blob/a0d51fdd3bea52c228c8f74c614fc77bb7cf1bc5/strax/context.py#L2021)
- [L549](https://github.com/AxFoundation/strax/blob/a0d51fdd3bea52c228c8f74c614fc77bb7cf1bc5/strax/context.py#L549)
- [L695](https://github.com/AxFoundation/strax/blob/a0d51fdd3bea52c228c8f74c614fc77bb7cf1bc5/strax/context.py#L695)

After writing #690, I'm starting to seriously doubt the usefulness of having a `version` that depends on the `run_id` (which isn't even generated from the `Plugin.run_id` attribute).

WenzDaniel:
Yes I think having run_id dependent versions does not make any sense anymore. To me https://github.com/AxFoundation/strax/blob/a0d51fdd3bea52c228c8f74c614fc77bb7cf1bc5/strax/plugin.py#L237-L243 looks like some legacy code and with CMT2.0 certainly not required anymore.

WenzDaniel:
But I am wondering if we should not try to be consistent in our code and either use always `__version__` or `.version()`. I would be in favor of the former. If we do the later we might want to make it a  property.

JoranAngevaare:
> But I am wondering if we should not try to be consistent in our code and either use always
Right, this issue is precisely about this 😉

I would also be in favor of dropping  the `version` class method (see my later addendum to this issue). I'm not super thrilled about this PR: https://github.com/AxFoundation/strax/pull/690. I can make https://github.com/AxFoundation/strax/pull/689 still work easily if we drop the `.version`-method by setting the `__getattr__` method appropriately.

My main issue is that the `version` method can now contain all kinds of weird code, which requires you to init the plugin in order to know what the version is going to be, which is slow. Making it a classmethod would already help.

PR #689, opened 2022-04-25
Auto version, fix #217

### LINKED ISSUE #217: Default version number
It would be helpful for plugin development if the default version number of a plugin was derived from an inspection of its own code. That way, changing the code would force reprocessing of the products of that plugin. A manually assigned version number would override that, so forced reprocessing remains in the user's control.

JoranAngevaare:
Add auto-inferring option to plugin by setting `__version__ = None`. Fixes #217

Generate a hash on the fly by hashing the plugin class.

Please see the appended test for a use case.

**Catches and pitfalls**
With https://github.com/AxFoundation/strax/issues/688 around, this would lead to some ambiguous results in the affected methods.

### DIFF
--- a/strax/plugin.py
+++ b/strax/plugin.py
@@ -242,13 +242,14 @@ class Plugin:
 
         Activate with setting __version__ to None
         """
-        attributes = [attr for attr in self.__dir__() if not attr.startswith('__')]
+        attributes = [attr for attr in self.__dir__()
+                      if not attr.startswith('__')]
 
         def _return_hashable(attr):
-            obj = getattr(self, attr)
-            if attr in ['takes_config']:
-                # handled by context
+            if attr in ['takes_config', '_auto_version']:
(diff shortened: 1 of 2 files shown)


WenzDaniel:
I do not like this change to be honest. It would make the version unreadable and things less comparable and harder to track in my opinion. Or do you envision this tool to be used only during plugin development? In that case I think it can be useful.

JoranAngevaare:
> I do not like this change to be honest. It would make the version unreadable and things less comparable and harder to track in my opinion. Or do you envision this tool to be used only during plugin development? In that case I think it can be useful.

@WenzDaniel , perhaps I should have stressed this. I indeed definitely would only argue using it for development. I would certainly not recommend using it in a mature framework like straxen.

WenzDaniel (approved):
Looks fine, thanks.

PR #690, opened 2022-04-25
Use plugin.version, fix #688

### LINKED ISSUE #688: `Plugin.__version__` and `Plugin.version` are ambigous

JoranAngevaare:
See #688 .

I did not change the _context_cache as I explained in the comment. It would be solving a problem that is so unlikely to which would potentially lead to very slow data availability checks that I'm not in favor of chaining it. When someone is really insisting, I could add a check that check that the `Plugin.version` method cannot be overwritten when `use_per_run_defaults` is deactivated.

Get a plugin instance to compute the version number based on a run-id.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -2031,7 +2031,7 @@ class Context:
         :return: dictionary of provided dtypes with their corresponding
             lineage hash, save_when, version
         """
-        dtypes = tuple(self._plugin_class_registry.key())
+        dtypes = tuple(self._plugin_class_registry.keys())
         plugins = self._get_plugins(targets=dtypes, run_id=run_id)
         hashes = set([
             (data_type,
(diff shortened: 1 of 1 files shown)


JoranAngevaare:
I don't really like this solution, there is too much arbitrariness, I'll close and keep the reference in the cited issue

PR #694, opened 2022-05-16
Prevent float/int funkyness in hitlet processing

JoranAngevaare:
The default method for create hitlets from hits returns bad results. Comparing ints and floats in numba is not recommendable, I suspect that the compare [between them did not work well on this line](https://github.com/AxFoundation/strax/blob/6aa0a4d0d33f73650139e2a2ebc162d0f0e27c89/strax/processing/hitlets.py#L107).

So now, lets just take the largest and smallest numbers in `np.int64`

See the issue here:
afbeelding
And the difference between ints and floats (the default was np.inf) for the `chunk_end`
afbeelding


WenzDaniel (approved):
Thanks a lot

PR #695, opened 2022-05-27
Fix overlflow bug in sort_by_time add little test

WenzDaniel:
Fixes an overflow bug in `sort_by_time`. Adds a little test which triggers the overflow.


JoranAngevaare (commented):
Nice work, could I make a few suggestions:
1. make the 10_000 configurable in `_sort_by_time` it should only be `max_channel +1`
2. start with checking that `channel` is in `x.dtype.names` if not, do you get the same `x10` improvement if w.r.t. `np.sort` (need a second `_sort_by_time` for this though)?

WenzDaniel:
Okay I made the changes for 1. Regarding 2. I do not see directly the use case for that. The only data_kind without any channel field is events and I do not think that a performance boost is needed here.

strax/processing/general.py
```diff
@@ -4,26 +4,39 @@
...
-        # Nothing to do, and .min() on empty array doesn't work, so:
         return x
+    with warnings.catch_warnings():
+        warnings.filterwarnings('error')
+        try:
+            max_channeL_plus_one = x['channel'].max() + 1
+            # Check if we trigger an overflow warning:
+            (x['time'].max() - x['time'].min()) * max_channeL_plus_one
+            x = _sort_by_time(x, max_channeL_plus_one)
+        except RuntimeWarning:
```
JoranAngevaare, line 25: this is rather crude, we know when we are going to have an overflow, you can easily calculate it, why not check that instead of trying and thereby doubling the work?

JoranAngevaare (commented):
@WenzDaniel there are many applications, e.g. in pema when sorting things like the truth et cetera, I don't see why we we wouldn't add some function like that to be honest. It's only a few lines - I could also add them if you want

strax/processing/general.py
WenzDaniel: Hmm, we could also do something like 
```
if not (x['time'].max() - x['time'].min())  > np.iinfo(np.int64).max / max_channel_plus_one
```
But I was a bit afraid of the division.

JoranAngevaare: right, but you can have an off by a few and make the switch earlier - that wouldn't matter in the end

By the way, any idea why the tests in straxen fail - it seems like something not working?

WenzDaniel: No not yet, I saw it but was not yet in the mood to check why :D

strax/processing/general.py
```diff
@@ -4,26 +4,48 @@
...
+    else:
+        channel = np.ones(len(x))
+        
+    max_time_difference = (np.iinfo(np.int64).max - 10) / np.abs(channel.max()+1)
+    # Subtract 10 to have some extra margin, just in case.
+    # Use absolute to account for peaks which are channel -1.
+    _time_range_too_large = (x['time'].max() - x['time'].min()) > max_time_difference
+    if not _time_range_too_large:
+        # Faster sorting:
+        x = _sort_by_time_and_channel(x, channel, channel.max()+1)
```
JoranAngevaare, line 31:
```suggestion
    max_time_difference = (np.iinfo(np.int64).max - 10) / np.abs(channel.max()+1)
    # Subtract 10 to have some extra margin, just in case.
    # Use absolute to account for peaks which are channel -1.
    _time_range_too_large = (x['time'].max() - x['time'].min()) > max_time_difference
    if not _time_range_too_large:
        # Faster sorting:
        if 'channel' in x.dtype.names:
            channel = x['channel']
        else:
            channel = np.ones(len(x))

        x = _sort_by_time_and_channel(x, channel, channel.max()+1)
```

JoranAngevaare (approved):
Looks good if the tests pass, I think it looks good. Maybe move making the channel buffer in the 
_time_range_too_large

strax/processing/general.py
WenzDaniel: I cannot because I need a channel already in the line before `max_time_difference = (np.iinfo(np.int64).max - 10) / np.abs(channel.max()+1)`.

WenzDaniel:
Now it should also work with peaks properly.

strax/processing/general.py
JoranAngevaare: ah yeah I see 👍

JoranAngevaare:
Great! Sorry for being so pedantic, but can't we just make all the sort channels positive? It's both faster and more generic, right?

afbeelding

JoranAngevaare:
@WenzDaniel I'm merging, I pushed two tiny changes: [936fc5d](https://github.com/AxFoundation/strax/pull/695/commits/936fc5d0f81bc6563fa7f488e99ab42fa7eaad3d), [7313d45](https://github.com/AxFoundation/strax/pull/695/commits/7313d45179194749f62a5c2c37fbc306c6dd6a4e)

PR #698, opened 2022-08-09
Numpy caching of data in online monitor storage

JoranAngevaare:
In https://github.com/AxFoundation/strax/issues/346, we uttered the prophetic words
```It's just the backend-buffer that is sitting otherwise idle in the background eating your RAM that needs clearing.```
We now learned that this is true when we have longer than usual runs and data with large volumes.

We also should have known that the `sys.getsizeof` wasn't taking the total size of the document but only of the dictionary.

This PR fixes both with using numpy for the buffers such that the max size of buffers is respected by the backend.

### DIFF
--- a/strax/storage/mongo.py
+++ b/strax/storage/mongo.py
@@ -114,7 +114,7 @@ class MongoBackend(StorageBackend):
             chunk_key = doc.get('chunk_i', None)
             if chunk_key is None:
                 # Should not happen because of the projection in find
-                # but let's double check:
+                # but let's double-check:
                 raise ValueError(
                     f'Projection failed, got doc with no "chunk_i":\n{doc}')
             # Update our registry with this chunks info. Use chunk_i as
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by mflierm — APPROVED** [review]

File: strax/storage/mongo.py [hunk 1]
```diff
@@ -123,9 +123,8 @@ def _build_chunk_registry(self, backend_key, dtype):
                     # Convert JSON to numpy
             chunk_len = len(doc.get('data', []))
             result = np.zeros(chunk_len, dtype=dtype)
-            for i in range(chunk_len):
-                for key in np.dtype(dtype).names:
-                    result[i][key] = doc['data'][i][key]
+            for key in np.dtype(dtype).names:
+                result[key] = [dd['data'][key] for dd in doc]
```
JoranAngevaare, line 127: For bookkeeping, this is about 10x faster:

```python
n=int(1e4)
bla = list(range(n))
n_bla = np.zeros(len(bla), dtype=[('a', np.int64), ('b', np.int64), ('c', np.int64)])
bla_d = [{'a': i, 'b': i, 'c':1} for i in bla]


%%timeit
for i in range(len(bla)):
    for key in ['a', 'b', 'c']:
        n_bla[i][key] = bla_d[i][key]
22.4 ms ± 97.1 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)


%%timeit
for key in ['a', 'b', 'c']:
    n_bla[key] = [dd[key] for dd in bla_d]
1.98 ms ± 50.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

PR #700, opened 2022-09-07
Restructure plugins

JoranAngevaare:
Small restructure of the plugin modules. This PR splits the large strax.plugin file into separate files for each plugin


WenzDaniel:
I think is fine although I am not a fan of these refactorings. Always makes it harder to back track who implemented something via git blame. But I agree it makes sense for the plugins file. Do we still need  the LoopPlugin?

JoranAngevaare:
> I think is fine although I am not a fan of these refactorings. Always makes it harder to back track who implemented something via git blame. But I agree it makes sense for the plugins file. 

Yeah, I would also not have started it if not for https://github.com/AxFoundation/strax/pull/654. I 100% see your argument. 

> Do we still need the LoopPlugin?

I think it does not hurt to keep, it has some nice and easy examples. Nevertheless, we could opt for adding a warning (that can be disabled) to use something like split by containment?

jmosbacher:
I think im for this. I think refactoring in a way that makes maintaining the code easier is becoming more important as the maintenance will have to pass to third-generation maintainers soon...

PR #702, opened 2022-10-14
fixed delta peak timestamp problem

FaroutYLq:
The (lone hits) delta peaks added to `merged_s2s` waveforms have wrong time stamp. For example, in the example `merged_s2s` waveforms below, the red shadow marks peaklets while the blue shadow marks where we expect the lone hits, given the time stamps directly from `lone_hits` datatype. You can see the lone hits show up in the wrong time.
image
image
image

This PR did a minimal fix to this issue.

```
index = (p['time'] - lh_i['time'])//p['dt']
```
is the problem. Apparently, lone hits' time should show up only after `merged_s2s` start. This implementation means that `index` is always negative, leading to a reversed ordering thus mirrored relative timestamp.

I think this fix is simple enough to skip this part...

### DIFF
--- a/tests/test_peak_merging.py
+++ b/tests/test_peak_merging.py
@@ -51,12 +51,18 @@ def test_replace_merged(intervals, merge_instructions):
 
 
 @hypothesis.given(fake_hits,
-                  hypothesis.strategies.integers(min_value=0, max_value=100))
+                  hypothesis.strategies.integers(min_value=0, max_value=int(1e18)),
+                  hypothesis.strategies.integers(min_value=0, max_value=100),
+                  hypothesis.strategies.integers(min_value=1, max_value=2),
+                  )
 @hypothesis.settings(deadline=None)
-def test_add_lone_hits(hits, peak_length):
+def test_add_lone_hits(hits, time_offset, peak_length, dt):
     peak = np.zeros(1, dtype=strax.peak_dtype())
+    peak['time'] = time_offset
+    hits['time'] += time_offset
     peak['length'] = peak_length
-    peak['dt'] = 1
+    hits['area'] = 1
+    peak['dt'] = dt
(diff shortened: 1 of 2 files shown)


JoranAngevaare:
Great catch, must have been introduced in https://github.com/AxFoundation/strax/pull/506!

Thanks a lot Lanqing!

One day we may try again to not merge lone hits in te first place, especially if they are after pulses.

JoranAngevaare (approved):

WenzDaniel:
Thanks for catching Lanqing. 

Rearding 

> One day we may try again to not merge lone hits in te first place, especially if they are after pulses.

One has to be a bit careful to not introduce in that case a z- and area dependent bias. This was one of the issues in the 1T Ar37 analysis as far as I remember.

WenzDaniel:
Can you update the test such that we do not run into this issue ever again https://github.com/AxFoundation/strax/blob/d6b3575df08f639cd4d054f116bb57f3cd2db92e/tests/test_peak_merging.py#L56 ?

tests/test_peak_merging.py
```diff
@@ -51,12 +51,18 @@ def test_replace_merged(intervals, merge_instructions):
...
+                  )
 @hypothesis.settings(deadline=None)
-def test_add_lone_hits(hits, peak_length):
+def test_add_lone_hits(hits, time_offset, peak_length, dt):
     peak = np.zeros(1, dtype=strax.peak_dtype())
+    peak['time'] = time_offset
+    hits['time'] += time_offset
     peak['length'] = peak_length
-    peak['dt'] = 1
+    hits['area'] = 1
```
JoranAngevaare, line 64: @WenzDaniel if we had realized this - the test would have worked. Now it always added zeros.

Issue #704, opened 2022-10-25
Peak waveform is truncated


terliuk:
**Describe the bug**

It looks like during peak building the peak waveform is being truncated by at least one sample, resulting in the last peaklet being truncated and ignored. This results in mismatch between data and peak area, as they are computed from peaklets and lone hits separately. 

Area is not expected to be affected, while peak timing properties can have a slight effect. 

**Screenshots**
effect on summed area
![image](https://user-images.githubusercontent.com/12826612/197827531-8011a6e1-6c66-485b-ab9e-03ad08b836a8.png)
And example of a waveform (zoomed out and zoomed in)
![image](https://user-images.githubusercontent.com/12826612/197828036-12b2054b-df43-4c12-b7dd-5e1082a4aede.png)
![image](https://user-images.githubusercontent.com/12826612/197828049-3effa44b-fc0e-414a-b2d2-ac841b532295.png)

**Versions**

I am using current (25 October 2022) development contaner with version
strax : 1.4.0 
straxen: 2.0.0

PR #705, opened 2022-10-26
Digitize full WF

JoranAngevaare:
See #704 

The sum WF is always conservative by truncating the peak when downsampling. While probably some loss is acceptable, having O(3%) differences between `p['data'].sum()` and `p['area']` might be a bit more confusing than what we'd 

Rather than being conservative with the clipping of the data-field, use the time if there is any after the peak. This is computed by checking if the next peak is far away enough that we can claim the bit of extra time that would be needed for the extra sample

There may still be a chance that there is another peak directly after another peak that would prevent us from filling the entire buffer. There are a couple of things we could do about that:
 - just ignore it - the chances that this happens is extremely small - especially for smaller peaks where this last sample might make any difference
 - add the data of that last sample that doesn't fit within the time interval to the second to last sample
 - rewrite tons of code to support overlapping peaks

### DIFF
--- a/strax/processing/peak_building.py
+++ b/strax/processing/peak_building.py
@@ -126,7 +126,10 @@ def find_peaks(hits, adc_to_pe,
 
 @export
 @numba.jit(nopython=True, nogil=True, cache=True)
-def store_downsampled_waveform(p, wv_buffer, store_in_data_top=False,
+def store_downsampled_waveform(p,
+                               wv_buffer,
+                               max_endtime=0,
+                               store_in_data_top=False,
                                wv_buffer_top=np.ones(1, dtype=np.float32)):
     """Downsample the waveform in buffer and store it in p['data'] and
     in p['data_top'] if indicated to do so.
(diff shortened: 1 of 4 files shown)


terliuk:
I would vote for option 2 with adding remaining part of the waveform to the last sample is the most reasonable option. As typical difference is below few %, this should preserve all the timing quantities that we typically use (50% width, 90% width, 10-to-50% area time etc). But we have to be sure that we document it clearly somewhere (at least as a very clear comment in code). 

On the long run, we can decouple the relation `endtime=length*dt + time` and to store dedicated end time or total duration of all samples, but this might require way more effort for testing, as it might "backfire" in some place.

JoranAngevaare:
> I would vote for option 2 with adding remaining part of the waveform to the last sample is the most reasonable option. As typical difference is below few %, this should preserve all the timing quantities that we typically use (50% width, 90% width, 10-to-50% area time etc). But we have to be sure that we document it clearly somewhere (at least as a very clear comment in code).


Yes agree, it's all quite hypothetical but we got to be mindfull of future PHD students banging their heads against the wall why things don't sum up. I just added the functionality. I'll try and add some more tests too to illustrate the point.


> On the long run, we can decouple the relation `endtime=length*dt + time` and to store dedicated end time or total duration of all samples, but this might require way more effort for testing, as it might "backfire" in some place.

Not sure if we would really gain much, We inherently loose information from downsampling, but that does not mean that the information lost is actually useful

JoranAngevaare:
FYI, if someone wants to pick this up, I did not figure out where but something is still a bit fishy about this PR as it creates very different Sum WFs (whereas it should be a very minor change). I'll close this for now, people might find it back in #704

Issue #707, opened 2022-12-30
Negative max_gap in peak_building


dachengx:
**Describe the bug**

In function `strax.processing.peak_building`, `max_gap` should be selected from positive time gaps, but in [code](https://github.com/AxFoundation/strax/blob/28553a74e6d56fc1dbb33b6c0d7cd381b0afb0aa/strax/processing/peak_building.py#L58)
```
p['max_gap'] = max(p['max_gap'], t0 - peak_endtime)
```
`t0 - peak_endtime` can be negative. It should be changed to 
```
p['max_gap'] = max(p['max_gap'], abs(t0 - peak_endtime))
```

**To Reproduce**
Insert the MWE of how to reproduce the error
```python
YOUR CODE GOES HERE
```

**Expected behavior**

**Screenshots**

**Versions**
Please add the version of strax and any related package

PR #708, opened 2022-12-30
Select max gaps from positive gaps

dachengx:

Fix: https://github.com/AxFoundation/strax/issues/707

The plugins in [straxen](https://github.com/XENONnT/straxen) might also need to be updated. 

Select max gaps from positive gaps, the gaps are differences between endtime of hits and time of hits. 

Change
```
p['max_gap'] = max(p['max_gap'], t0 - peak_endtime)
```
to
```
p['max_gap'] = max(p['max_gap'], abs(t0 - peak_endtime))
```


WenzDaniel:
Hej Dacheng thanks for spotting this, but I think this points to a different issue. `p['max_gap']` should be by default zero. So `max(p['max_gap'], t0 - peak_endtime)` can only be `>= 0` is not it? Could it be that we have an overflow of the buffer?

WenzDaniel (approved):

dachengx:
> Hej Dacheng thanks for spotting this, but I think this points to a different issue. `p['max_gap']` should be by default zero. So `max(p['max_gap'], t0 - peak_endtime)` can only be `>= 0` is not it? Could it be that we have an overflow of the buffer?
> 
> Okay I just noticed that I misunderstood your PR. You are absolutely right, there should be an absolute value as the next hit in a peak can be beyond the peak endtime as long as the this hit is within the max_gap range.

Hey @WenzDaniel . Yes, you are right. Also the default value of `p['max_gap']` is -1.

dachengx:
> Hey @WenzDaniel . Yes, you are right. Also the default value of `p['max_gap']` is -1. 

I just realized that the default value of `p['max_gap']` is indeed 0. We may see some `p['max_gap'] == -1` because we set it to -1 when [split the peak](https://github.com/AxFoundation/strax/blob/28553a74e6d56fc1dbb33b6c0d7cd381b0afb0aa/strax/processing/peak_splitting.py#L161).

dachengx:
Let me illustrate more here, only for reference: 

Before put into `find_peaks`, the hits are [sorted](https://github.com/XENONnT/straxen/blob/81a206bd14e300cb4f98f1396074c046f5875bcd/straxen/plugins/peaklets/peaklets.py#L181), but by `time` and `channel`. So there might be the case that the following peak starts before previous peak ends. Thus `t0 - peak_endtime` can be negative. 

image

In the figure, `time` is solid lines and `endtime` is dashed lines.

dachengx:
I guess the `max_gap` is meant to track the results of `find_peaks` because the function looks into the difference of `endtime` and `time`. But to make further use of hits properties, we might need some other features which I will add them in another PR.

JoranAngevaare (approved):

dachengx:
> You mean the following _hit_ hit right? "following _hit_ starts before previous _hit_ ends"

Yes. Sorry, that is a typo.

dachengx:
Shall we update the lineage of plugins in straxen?

PR #709, opened 2023-02-06
pbar for rechunker

JoranAngevaare:
Add an optional argument for displaying a pbar for the rechunker


JoranAngevaare:
surpassed in #710

PR #714, opened 2023-03-03
Raise compression errors if unable

JoranAngevaare:
Raise an informative error if `blosc` would discard data

```python
import numpy as np
import strax

bla = np.zeros(int(3e8), dtype=np.int64)
print(bla.nbytes/1e9)
_bytes = strax.io.COMPRESSORS['blosc']['compress'](bla)
print(len(_bytes))

# prints
2.4
0
```

for bookkeeping, this was `blosc '1.11.1'`


mflierm (approved):

dachengx:
Nice catch! @JoranAngevaare It took me some time to realize that the current development environment is not updated to strax v1.4.3 to debug this. Do you think this bug can affect previously processed data?

PR #715, opened 2023-04-10
Set start of further chunk to be the smallest start of dependencies

dachengx:

Set start of further chunk to be the smallest start of dependencies' chunk

This is an update to PR https://github.com/AxFoundation/strax/pull/345. We should also refer to issue https://github.com/AxFoundation/strax/issues/247. 

When loading `peaks` with `time_within`, an error shows here https://github.com/AxFoundation/strax/blob/dae09e59442562c552dc3e65a8184955db9c82a3/strax/chunk.py#L81. 
Because the chunks are split by https://github.com/AxFoundation/strax/blob/dae09e59442562c552dc3e65a8184955db9c82a3/strax/chunk.py#L377 (or maybe other reasons such as the algorithm of `merged_s2s`) so the `start` of `merged_s2` can be close to and smaller than the `start` of `peaklets`. The `start` of `peaklets` is also the `start` of the further chunk, because of https://github.com/AxFoundation/strax/blob/dae09e59442562c552dc3e65a8184955db9c82a3/strax/plugins/plugin.py#L570. So the 1st `time` of `peaks` chunk is smaller than the `start` of `peaklets`. 

  - Update the docstring(s)
  - Update the documentation


WenzDaniel (commented):
Thanks @dachengx looks fine to me.

PR #716, opened 2023-04-13
Save hits level information(hits time difference) in peaks

dachengx:

Save hits level information(hits height and time difference) in peaks. 

By adding some extra fields(`max_diff`, `min_diff`) in `dtypes.peak_dtype`. Then in `PeakSplitter._split_peaks` we will set the merged peaks' extra features to -1. 

  - Update the docstring(s)
  - Update the documentation


dachengx:
The two added fields will be then calculated by the straxen PR: https://github.com/XENONnT/straxen/pull/1155, and function https://github.com/XENONnT/straxen/blob/52a4b403a05d31dd13cf3e9931c4d6f733e7657c/straxen/plugins/peaklets/peaklets.py#L301.

jmosbacher (commented):
Just curious, is it absolutely necessary to change the existing dtype for hitlets? this will make this strax incompatible with all our existing data right?
why not for instance just add these two extra fields with a separate plugin?

dachengx:
> Just curious, is it absolutely necessary to change the existing dtype for hitlets? this will make this strax incompatible with all our existing data right? why not for instance just add these two extra fields with a separate plugin?

The PR will change the dtype of `peak_dtype` but not really hitlets. The change here

https://github.com/AxFoundation/strax/blob/059a0fe9248c1714e483669b45fee88343d77c99/strax/dtypes.py#L155

is only called in

https://github.com/AxFoundation/strax/blob/059a0fe9248c1714e483669b45fee88343d77c99/strax/processing/peak_splitting.py#L113. Those added fields to `hitlet_with_data_dtype` are designed to be compatible with the peak splitting function. 

The current implementation of this PR will make new strax incompatible with all our existing data. Since certain data is only compatible with a certain version of strax, I guess the only issue can be storage. Additionally, we already have bug in `peaklets`(fixed in https://github.com/AxFoundation/strax/pull/708). So maybe a thorough reprocessing is needed. 

These fields _**can**_ be added to another separated dtype, maybe `hits_diff_dtype`, and calculated in `Peaklets`. But the reprocessing is not cheaper because `records` were not stored so we anyway need to reprocess from `raw_records`.

dachengx:
Hey @jmosbacher. After a little bit of struggle in https://github.com/XENONnT/straxen/pull/1168, I get a dependency tree like https://straxen--1168.org.readthedocs.build/en/1168/reference/datastructure_nT.html#event-basics. But still, until https://github.com/XENONnT/straxen/pull/1168/commits/c68a5cbd16db7d72a2aa64c1a1cb10f7a67feec4 I can not find an easy way to avoid merge `max_diff` and `min_diff` into `peak_dtype`. So to avoid confusion among users, I insist on this PR and propose a minor version bump after this.

jmosbacher (approved):
Ok, as long as you are aware of the complexity this introduces and came to the conclusion that it is necessary then I have no issues with it.

PR #721, opened 2023-05-04
Fix argsort inside numba.jit using kind='mergesort'

dachengx:
As the title described. 

This PR is trying to prevent issues like https://github.com/XENONnT/straxen/issues/1175. 

Similar to https://github.com/XENONnT/straxen/pull/1176. 

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/run_selection.py
+++ b/strax/run_selection.py
@@ -306,7 +306,7 @@ def define_run(self: strax.Context,
     if isinstance(data, (pd.DataFrame, np.ndarray)):
         if isinstance(data, np.ndarray):
             data = pd.DataFrame.from_records(data)
-      
+
         # strax.endtime does not work with DataFrames due to numba
         if 'endtime' in data.columns:
             end = data['endtime']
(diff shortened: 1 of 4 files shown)


strax/run_selection.py
```diff
@@ -306,7 +306,7 @@ def define_run(self: strax.Context,
     if isinstance(data, (pd.DataFrame, np.ndarray)):
         if isinstance(data, np.ndarray):
             data = pd.DataFrame.from_records(data)
-
```
dachengx, line 309: This is just to remove an unused line.

jmosbacher (approved):
Nice catch.
Just wondering  if maybe it would be more future-proof if you had the `_sort_by_time_and_channel` function accept `sort_kind` as an argument so users could choose other options if they know what they are doing.

dachengx:
> Nice catch. Just wondering if maybe it would be more future-proof if you had the `_sort_by_time_and_channel` function accept `sort_kind` as an argument so users could choose other options if they know what they are doing.

Nice suggestion. Updated.

jmosbacher:
Maybe only slightly related to this PR, but would be nice at some point to introduce tests that check that all jit compiled functions produce the same results as their pure python implementations

PR #725, opened 2023-05-18
Check whether `things` and `containers` are sorted

dachengx:
in `split_by_containment` and `touching_windows`

`things` and `containers` should be sorted in `split_by_containment` and `touching_windows` functions.  This PR add a function `_check_sorted` to test whether they are sorted. 

This might be too strict and we might see errors in existed codes. 

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/processing/general.py
+++ b/strax/processing/general.py
@@ -176,6 +176,7 @@ def split_by_containment(things, containers):
         empty_list = empty_list[:0]
         return empty_list
 
+    _check_sorted(things, containers)
     return _split_by_containment(things, containers)
 
 
@@ -331,6 +332,7 @@ def touching_windows(things, containers, window=0):
          (i.e. container endtime equals the thing starttime, since strax
           endtimes are exclusive)
     """
+    _check_sorted(things, containers)
     return _touching_windows(
         things['time'], strax.endtime(things),
         containers['time'], strax.endtime(containers),
@@ -364,3 +366,14 @@ def _touching_windows(thing_start, thing_end,
         result[i] = left_i, right_i
 
     return result
+
+def _check_sorted(things, containers):
+    """
+    Check whether things and containers are sorted
+    :param things: Sorted array of interval-like data
+    :param containers: Sorted array of interval-like data
+    """
+    for array, name in zip([things, containers], ['things', 'containers']):
+        assert np.all(np.diff(array['time']) >= 0), f'{name} should be sorted'
+        assert np.all(array['endtime'] - array['time'] >= 0), f'{name} should'\
+            + ' have non-negative length'
--- a/strax/processing/general.py
+++ b/strax/processing/general.py
@@ -1,10 +1,10 @@
 import os
+from functools import wraps
 
 import strax
 import numba
 from numba.typed import List
 import numpy as np
-import warnings
 
 export, __all__ = strax.exporter()
 
@@ -127,13 +127,15 @@ def _find_break_i(data, safe_break, not_before):
 
 @export
 @numba.jit(nopython=True, nogil=True, cache=True)
-def fully_contained_in(things, containers):
+def fully_contained_in(things, containers, length_threshold=0, needs_check_timing=True):
     """Return array of len(things) with index of interval in containers
     for which things are fully contained in a container, or -1 if no such
     exists.
     We assume all intervals are sorted by time, and b_intervals
     nonoverlapping.
     """
+    if needs_check_timing:
+        _check_things_containers(things, containers, length_threshold, sorted_endtime=False)
     result = np.ones(len(things), dtype=np.int32) * -1
     a_starts = things['time']
     b_starts = containers['time']
@@ -161,13 +163,15 @@ def _fc_in(a_starts, b_starts, a_ends, b_ends, result):
 
 
 @export
-def split_by_containment(things, containers):
+def split_by_containment(things, containers, length_threshold=0, needs_check_timing=True):
     """
     Return list of thing-arrays contained in each container. Result is
     returned as a numba.typed.List or list if containers are empty.
 
     Assumes everything is sorted, and containers are non-overlapping.
     """
+    if needs_check_timing:
+        _check_things_containers(things, containers, length_threshold, sorted_endtime=False)
     if not len(containers):
         # No containers so return empty numba.typed.List
         empty_list = List()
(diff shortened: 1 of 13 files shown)


strax/processing/general.py
```diff
@@ -364,3 +366,14 @@ def _touching_windows(thing_start, thing_end,
...
     return result
+
+def _check_sorted(things, containers):
+    """
+    Check whether things and containers are sorted
+    :param things: Sorted array of interval-like data
+    :param containers: Sorted array of interval-like data
+    """
+    for array, name in zip([things, containers], ['things', 'containers']):
+        assert np.all(np.diff(array['time']) >= 0), f'{name} should be sorted'
```
WenzDaniel, line 377: Is this truly a requirement do all time values to be non-zero? I do not think so.

strax/processing/general.py
```diff
@@ -364,3 +366,14 @@ def _touching_windows(thing_start, thing_end,
...
+        assert np.all(array['endtime'] - array['time'] >= 0), f'{name} should'\
```
WenzDaniel, line 378: This does not check if things are sorted does it? It only checks if an interval is not negative. Things being sorted would be 
```python
np.diff(array['time']) >= 0
```
Some algorithms also require the intervals to be non-overlapping. In that case you could use:
```python 
    assert np.all(array['endtime'][:-1] - array['time'][1:] < 0)
```

WenzDaniel (commented):
Hje @dachengx thanks for adding this little check. However, I think you are not testing exactly what your description says, see the comments below.

strax/processing/general.py
dachengx: It is to say the differences between adjacent times are not negative.

strax/processing/general.py
dachengx: I intended to make another check that the length of `things` and `containers` are non-negative here.

dachengx: `touching_windows` also needs the `endtime` to be sorted. `split_by_containment` by definition is stricter and requires the intervals to be non-overlapping, because it returns only one container for each thing.

WenzDaniel: I must apologize, I just noticed that you are in fact computing the `diff` here... Somehow I have overlooked this yesterday o.O 

For split by containment we could add the second check I mentioned to test that intervals are indeed non-overlapping.

strax/processing/general.py
WenzDaniel: Yes, see  my comment below, I was being stupid.

dachengx:
As a summary, the most stringent check will be:

1. for `split_by_containment`, `_split_by_containment`, and `fully_contained_in`, we need to make sure that the `time` is sorted and that containers have non-negative lengths. 
2. for `touching_windows`, we need to make sure that the `time` and `endtime` are sorted and that containers have positive lengths.

FaroutYLq:
Hi @mflierm @cfuselli. Since Dacheng is pretty piled up right now, and we have a reprocessing coming very soon (last week we planned next week...), can you guys please help do the [2 checks](https://github.com/AxFoundation/strax/pull/725#issuecomment-1562464346) mentioned by @dachengx ?

WenzDaniel:
Since it is a change in strax I insist on adding an unittest before adding this function. We have become a bit sloppy about this. Even better would be a hypothesis test. See e.g..: https://github.com/AxFoundation/strax/pull/726

dachengx:
Before this commit 6d2e343e5021607e500d7918b2c1815d5b2ab506, we are not able to do tests with not disjoint overlapping boundaries because whatever we set to `disjoint`, the boundaries are always not overlapping.

dachengx:
@WenzDaniel To explain why we need the endtime to be sorted in `touching_window`, the following code will give `[[0, 3]]` which does not make sense. The 2nd things `(2,3)` and the container have no overlapping. 
```
things = np.array(
    [(0, 6), (2, 3), (4, 8)], dtype=strax.time_fields)
containers = np.array(
    [(5, 7)], dtype=strax.time_fields)

strax.touching_windows(things, containers
```

strax/processing/general.py
```diff
@@ -389,7 +389,7 @@ def _check_things_containers(things, containers, length_threshold=0, sorted_endt
     if sorted_endtime:
         _check_sorted_endtime(things, containers)
     _check_length(things, containers, length_threshold)
-    _check_overlapping(things, containers)
+    # _check_overlapping(things, containers)
```
dachengx, line 392: We should still check that the containers are not overlapping in `fully_contained_in`.

WenzDaniel:
Hej Dacheng, I am currently looking into this PR, and I think it needs a few changes. I doubt that we can have a single function which checks the sorting for things and containers as the requirements are different for `touching_windows` and `fully_contained_in`.  In addition, I think the issue is also a bit bigger than just adding a test as we use `touching_windows` in many situations where technically things and container are allowed to overlap, however touching windows does not work in that case.

strax/processing/general.py
```diff
@@ -1,10 +1,12 @@
 import os
+import warnings
+warnings.simplefilter('always', UserWarning)
+# for these fundamental functions, we throw warnings each time they are called
```
dachengx, line 4: Here I tried to throw error each time we see something considered as warning.

strax/processing/general.py
```diff
@@ -126,14 +128,41 @@ def _find_break_i(data, safe_break, not_before):
...
+        try:
+            _check_objects_non_negative_length(objects)
+        except Exception:
+            raise ValueError(f'{names} should have non-negative length!')
+
+    return _fully_contained_in(things, containers)
+
+
+@numba.jit(nopython=True, nogil=True, cache=True)
+def _fully_contained_in(things, containers):
```
dachengx, line 164: Here `_fully_contained_in` become the core function of `fully_contained_in`, in `fully_contained_in`, I added some error message.

strax/processing/general.py
```diff
@@ -347,6 +401,13 @@ def _touching_windows(thing_start, thing_end,
 
     for i, t0 in enumerate(container_start):
         t1 = container_end[i]
+        if t0 < thing_start[right_i]:
+            # Container overlapped with previous one so we have to
+            # go back to first left_i before current container, and
+            # reset right_i
+            while left_i > 0 and thing_end[left_i] > t0 - window:
```
dachengx, line 408: I modified a bit of the `while` statement, from `left_i >= 0 and thing_end[left_i] >= t0 - window` to `left_i > 0 and thing_end[left_i] > t0 - window`.

WenzDaniel (approved):
Ho Dacheng, thanks I like your idea of the sanity checks. We should monitor these warnings and

WenzDaniel (approved):
Thanks Dacheng I like your sanity check function. We should monitor these warning and see if we want to start raising errors instead at some point.

dachengx:
Just for the record, after this PR:

1. for both `fully_contained_in` and `touching_window`, `time` of `things` and `containers` must be sorted, otherwise, an error raises. 
2. for both `fully_contained_in` and `touching_window`, `things` and `containers` must have non-negative length, otherwise, an error raises. 
3. for `fully_contained_in`, a warning will show when containers are overlapping. 
4. for `touching_window`, a warning will show when `endtime` of things are not sorted.

dachengx:
I had worried about the versioning of the plugins in straxen. Currently for the TPC runs, `peak_shadow`, `peak_ambience`, `peaks`, `event_area_per_channel`, and `event_waveform` are heavily using `touching_windows`. But in these functions, `containers`' `endtime` are already sorted. So I think it is fine to not change the versions.

PR #726, opened 2023-05-23
Add new general fucntion which computes dt to some interval

WenzDaniel:
Adds general function which computes time difference between things and other intervals, e.g., events and DAQ veto intervals. 

Also added tests which check the general function behavior.

See: https://github.com/XENONnT/straxen/pull/1181 for a performance comparison.


cfuselli (approved):
Thanks Daniel. Looks all good to me, and the result is the same as what we had with the previous function used in [veto_proximity](https://github.com/XENONnT/straxen/blob/master/straxen/plugins/events/veto_proximity.py). Only difference is fixing how time_to_next is computed, now correctly fetching veto_time instead of veto_endtime. 

I tested this: 
```
Events: [(15,  20) (30,  35) (58, 105)]
Vetos [( 11,  12) ( 30,  38) ( 50, 100) (150, 900)]
Vetos lenghts [  1   8  50 750]

Times to previous interval (new): [ 3 18 20]
Times to previous interval (old): [ 3. 18. 20.]
Times to previous interval (diff): [0. 0. 0.]

Times to next interval (new): [10 15 45]
Times to next interval (old): [ 18.  65. 795.]
Times to prenextvious interval (diff): [  -8.  -50. -750.]
```

WenzDaniel:
@dachengx any additional comments? If not I will merge.

strax/processing/general.py
```diff
@@ -364,3 +364,69 @@ def _touching_windows(thing_start, thing_end,
...
+                times_to_prev[thing_ind] = dt
+                veto_intervals_seen += 1
+        
+        # Now check if current veto is still within event or already after it:
+        for veto_interval in intervals[veto_intervals_seen:]:
+            _current_interval_before_thing_ends = veto_interval['time'] < current_event_endtime
+            if _current_interval_before_thing_ends:
+                continue
+
+            # Now current event is after event so store time:
```
dachengx, line 428: Hey, @WenzDaniel . I guess here you mean 'Now current veto is after event so store time'?

WenzDaniel: Thanks for spotting yes.

PR #727, opened 2023-05-25
Accelerate select_runs

shenyangshi:
Select_runs is too slow since we check frontend availability for all runs in the run_db and then apply the tag mask. By moving the checking after the tag selection, we can accelerate run selection by 20 times for the runs we are interested.

### DIFF
--- a/strax/run_selection.py
+++ b/strax/run_selection.py
@@ -19,7 +19,11 @@ export, __all__ = strax.exporter()
 def list_available(self, target,
                    runs=None,
                    **kwargs) -> list:
-    """Return sorted list of run_id's for which target is available"""
+    """Return sorted list of run_id's for which target is available
+    :param target: Data type to check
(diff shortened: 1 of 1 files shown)


shenyangshi:
I think it’s due to the accumulation of runs, the size of run_db increases a lot.

dachengx:
> I think it’s due to the accumulation of runs, the size of run_db increases a lot.

Has someone recently manually changed RunDB? I still feel it is fast enough 1 week ago. We can discuss this in slack channel.

shenyangshi:
I guess so. But this function is always very slow (>1min) to run due to many redundant runs in the commissioning phase as well as unlinked runs not in the SR conditions.

dachengx:
Not bad. The codefactor complaining is not serious.

dachengx (approved):

PR #728, opened 2023-05-31
Loading plugins default to not from cache

shenyangshi:
Strax is not updating the plugin config when we register an updated plugin with new configs, since it defaults to load from cached plugin lists, the only option was to re-initialize the whole context.

This is causing confusion for straxen and cutax developers. Given the plugins are not taking many memories, maybe we can just load them again. Strax experts please comment @dachengx @WenzDaniel @JoranAngevaare.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -628,14 +628,16 @@ class Context:
 
     def _get_plugins(self,
                      targets: ty.Tuple[str],
-                     run_id: str) -> ty.Dict[str, strax.Plugin]:
+                     run_id: str,
+                     get_from_cache = False) -> ty.Dict[str, strax.Plugin]:
         """
         Return dictionary of plugin instances necessary to compute targets
         from scratch.
         For a plugin that produces multiple outputs, we make only a single
         instance, which is referenced under multiple keys in the output dict.
+        If get_from_cache is True, try to load the plugins from the cache.
         """
-        if self._plugins_are_cached(targets):
+        if self._plugins_are_cached(targets) and get_from_cache:
             return self.__get_plugins_from_cache(run_id)
 
         # Check all config options are taken by some registered plugin class


JoranAngevaare:
Hi @shenyangshi , setting up the caching functionality wasn't entirely trivial: https://github.com/AxFoundation/strax/pull/440, https://github.com/AxFoundation/strax/pull/485, https://github.com/AxFoundation/strax/pull/510, https://github.com/AxFoundation/strax/pull/516 https://github.com/AxFoundation/strax/pull/545, https://github.com/AxFoundation/strax/pull/561 so I can totally believe if there is some edge case not covered (yet).

1. Can you provide a MWE of the issue you are experiencing? 
2. This proposed change is not propagated to other functions, making it virtually impossible to use the cache after this change.
3. Disabling caching will hugely slow down data discovery (like select runs does), you can see the extreme performance loss by setting `use_per_run_defaults==True` in the context config and doing a `st.select_runs`. The cache was introduced precisely for this reason (otherwise, you build the plugin dependency tree for each time you call `st._get_single_plugin`, which is called from `st.select_runs` st.get_*` etc.).
4. I think you can only run into this issue when you don't change [the version of the plugin as well,](https://github.com/AxFoundation/strax/blob/ea28dae8443db2ca883d207e90ff7902723e00ea/strax/context.py#L551), which is probably bad practice since you'd much better use `st.set_config`. Yet, debugging this cache is quite a pain, so perhaps we should come up with something instead of saying "it's bad practice" 😉.
5. I think the easiest solution is to drop the current cache if one does an `st.register`, [maybe somewhere around this line](https://github.com/AxFoundation/strax/blob/ea28dae8443db2ca883d207e90ff7902723e00ea/strax/context.py#L263). you can drop the current cache by setting `st._fixed_plugin_cache=None` (or `self._fixed_plugin_cache=None`, if you add it to the `self.register`).

JoranAngevaare (changes requested):
Hi @shenyangshi , setting up the caching functionality wasn't entirely trivial: https://github.com/AxFoundation/strax/pull/440, https://github.com/AxFoundation/strax/pull/485, https://github.com/AxFoundation/strax/pull/510, https://github.com/AxFoundation/strax/pull/516 https://github.com/AxFoundation/strax/pull/545, https://github.com/AxFoundation/strax/pull/561 so I can totally believe if there is some edge case not covered (yet).

1. Can you provide a MWE of the issue you are experiencing? 
2. This proposed change is not propagated to other functions, making it virtually impossible to use the cache after this change.
3. Disabling caching will hugely slow down data discovery (like select runs does), you can see the extreme performance loss by setting `use_per_run_defaults==True` in the context config and doing a `st.select_runs`. The cache was introduced precisely for this reason (otherwise, you build the plugin dependency tree for each time you call `st._get_single_plugin`, which is called from `st.select_runs` st.get_*` etc.).
4. I think you can only run into this issue when you don't change [the version of the plugin as well,](https://github.com/AxFoundation/strax/blob/ea28dae8443db2ca883d207e90ff7902723e00ea/strax/context.py#L551), which is probably bad practice since you'd much better use `st.set_config`. Yet, debugging this cache is quite a pain, so perhaps we should come up with something instead of saying "it's bad practice" 😉.
5. I think the easiest solution is to drop the current cache if one does an `st.register`, [maybe somewhere around this line](https://github.com/AxFoundation/strax/blob/ea28dae8443db2ca883d207e90ff7902723e00ea/strax/context.py#L263). you can drop the current cache by setting `st._fixed_plugin_cache=None` (or `self._fixed_plugin_cache=None`, if you add it to the `self.register`).

WenzDaniel:
Hej Shenyang, thanks for spotting this issue. I second all of Jorans points. I developed a lot of plugins since we introduced the caching and never encountered such an issue. As Joran suggested can you provide an example? In addition, once you have an example can you make first a unittest based on that example which fails the test and then only afterwards add a solution which makes the test pass? I think we should in general enforce this order of making changes.

jmosbacher:
Hi, I would like to add to Joran and Daniels comments a more general comment. 
The percentage of time strax is used for development of new plugins is a tiny fraction of the time it is used for processing and loading data. 
It is a flawed approach to reduce the performance of every user's usage to save a few minutes during plugin development. The proper implementation of such a change (if indeed deemed to be necessary) would be to have a setting that disables the cache that developers can use during plugin development.
Removing existing features from software is not something to be lightly and especially not when the benefit is only seen by a tiny fraction of its usage.

shenyangshi:
Thanks everyone for the comments! Here is a MWE: https://github.com/XENONnT/analysiscode/blob/master/misc/example_strax.ipynb

But I realized this PR is not needed, as a much easier way to solve this problem as @JoranAngevaare mentioned, is to update plugin versions while making a change. I'll close this PR then.

JoranAngevaare:
Hi all, sounds good. Yet I just wanted to point out that option 5 is for free basically. If you set the cache to `None` (during st.register) you will rebuild it the next time anyway. So it won't impact any performance, just less poor PhD students that can't figure out why strax is so stubborn 😄.

But then again, updating the version of the plugin is almost always necessary for strax to prevent loading data with the same lineage. Unless you purposefully use `SaveWhen.NEVER` or something.

Issue #490, opened 2021-07-17
OverlapWindowPlugin causes strax.chunk.CannotSplit


JoranAngevaare:
**Describe the bug**
`OverlapWindowPlugin` causes un-splittable chunks. See e.g. `'020318'`

**To Reproduce**
Insert the MWE of how to reproduce the error
```bash
straxer     020318   --target event_info_double     --context xenonnt_v3     --package cutax     --notlazy     --timeout 1200     --context_kwargs '{"use_rucio": 1}'
```

**Expected behavior**
Since the OverlapWindowPlugin keeps track of what data is seen, it should never split results in such a way that the next time it splits this causes errors. This could potentially only happen if the overlap is too small and we are creating different splits because there is more data in the next iteration.

Additionally, we should implement a try: except block to provide a proper traceback since this is uninformative

**Versions**
```bash
Working on dali023.rcc.local with the following versions and installation paths:
python	v3.8.0	(default, Nov  6 2019, 21:49:08) [GCC 7.3.0]
strax	v0.16.1	/home/angevaare/software/dev_strax/strax/strax
straxen	v0.19.3	/home/angevaare/software/dev_strax/straxen/straxen
cutax	v0.1.1	/home/angevaare/software/dev_strax/cutax/cutax
```

**Traceback**
```Got 710 items. Now 337.3 sec / 18.7% into the run. Using 28501.6 MB RAM. ETA 1314.49 sec.
Exception in thread build:merged_s2s:
Traceback (most recent call last):
  File "/home/angevaare/software/Miniconda3/envs/strax_dev/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/home/angevaare/software/Miniconda3/envs/strax_dev/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/home/angevaare/software/dev_strax/strax/strax/mailbox.py", line 268, in _send_from
    self.kill_from_exception(e)
  File "/home/angevaare/software/dev_strax/strax/strax/mailbox.py", line 187, in kill_from_exception
    raise e
  File "/home/angevaare/software/dev_strax/strax/strax/mailbox.py", line 255, in _send_from
    x = next(iterable)
  File "/home/angevaare/software/dev_strax/strax/strax/plugin.py", line 616, in iter
    yield from super().iter(iters, executor=executor)
  File "/home/angevaare/software/dev_strax/strax/strax/plugin.py", line 427, in iter
[...]

PR #731, opened 2023-06-01
Improve chunks splitting in `OverlapWindowPlugin`

dachengx:

https://github.com/AxFoundation/strax/blob/b35db0a2f27be2746fb30c18d8f7d4b9d9e7722b/strax/plugins/overlap_window_plugin.py#L52-L53 set `allow_early_split=False` to make sure that the chunks are initialized correctly https://github.com/AxFoundation/strax/blob/b35db0a2f27be2746fb30c18d8f7d4b9d9e7722b/strax/chunk.py#L76-L85. 

In real data, we will sometime encounter overlapping peaks when the rate is high, so `result.split` will raise the error. 

The added codes search for splitting time `invalid_beyond` where no items happen in `provides`. 

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/plugins/overlap_window_plugin.py
+++ b/strax/plugins/overlap_window_plugin.py
@@ -88,7 +88,7 @@ class OverlapWindowPlugin(Plugin):
                 self.log.debug(
                     f'Inconsistent start times of the cashed chunks after'
                     f' {try_counter}/{max_trials} passes.')
-            invalid_beyond -= min(_potential_invalid_beyond) - 1  # ns
+            invalid_beyond = min(_potential_invalid_beyond) - 1  # ns
         else:
             raise ValueError(f'Buffer start time inconsistency cannot be '
                              f'resolved after {max_trials} tries')
--- a/strax/plugins/overlap_window_plugin.py
+++ b/strax/plugins/overlap_window_plugin.py
@@ -50,7 +50,7 @@ class OverlapWindowPlugin(Plugin):
 
         # Throw away results we already sent out
         _, result = result.split(t=self.sent_until,
-                                 allow_early_split=False)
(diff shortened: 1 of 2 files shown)


WenzDaniel:
Hej Dacheng, I am not sure what the current status of this PR is. Should I have a look? In addition, since it is a change for something so fundamental I think we should add a test.

dachengx:
> Hej Dacheng, I am not sure what the current status of this PR is. Should I have a look? In addition, since it is a change for something so fundamental I think we should add a test.

Hey Daniel. The codes are done. I will add a test. Sorry for making so many commits after marking the PR as ready.

WenzDaniel:
Since it is such a low level change can you provide maybe a test example where the old code fails and the new code passes? This would not only make reviewing much easier, but also provides some safety for the future.

dachengx:
@WenzDaniel Yes, I am trying to add that.

The current condition is like this:
1. the `sorted(x)` makes the result of `sorted_bounds` still disjoint even if the `disjoint=False` is argued https://github.com/AxFoundation/strax/blob/cb5664fc7cf4960fb10b47bdeca659a9d5dcfaec/strax/testutils.py#L36-L38.
2. so this makes us seldom(never) noticed that the `touching_window` is giving the wrong results even in the pytest action, .
3. https://github.com/AxFoundation/strax/pull/725 will be updated to not only raise errors when containers' `endtime`s are sorted, following @jmosbacher's great idea.
4. But the current `touching_window` will still not work when the things' `endtime`s are not sorted, in these cases, we need to use `split_by_containment`.

dachengx:
This PR is actually related to https://github.com/AxFoundation/strax/issues/490

PR #736, opened 2023-06-22
Fix touching window

dachengx:

In some insane cases, where both `endtime` of `things` and `container` are not sorted, like:

```
things = np.array([(0, 1, 1, 0),
                   (1, 5, 1, 0),
                   (2, 1, 1, 0)],
                   dtype=strax.interval_dtype)

containers = np.array([(0, 4, 1, 0), (0, 1, 1, 0)],
                       dtype=strax.interval_dtype)
```

The current algorithm will give wrong results:

```
[[0, 3], [0, 3]]
```

The correct answer is:

```
[[0, 3], [0, 1]]
```

The new algorithm only search for interval when the `time` and `endtime` of `things` and `containers` are sorted. If not, we first sort them and record the `argsort`, then later recover the order following `argsort` of input. 

When `things`' `endtime` are not sorted, we will return the indices of the first and last things which are touching the container. So we return the minimum index in `argsort`. 

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/processing/general.py
+++ b/strax/processing/general.py
@@ -410,30 +410,20 @@ def _touching_windows(thing_start, thing_end,
                       container_start, container_end,
                       window=0, endtime_sort_kind='mergesort'):
     n = len(thing_start)
-    thing_end_argsort = np.argsort(thing_end, kind=endtime_sort_kind)
-    thing_end_sort = thing_end[thing_end_argsort]
     container_end_argsort = np.argsort(container_end, kind=endtime_sort_kind)
 
     # we search twice, first for the beginning of the interval, then for the end
     left_i = right_i = 0
     result = np.zeros((len(container_start), 2), dtype=np.int32)
 
-    # the smallest index to the right of thing_end_argsort
-    # this is using space to save time complexity
-    min_thing_end_argsort = np.zeros(n + 1, dtype=np.int32)
-    min_thing_end_argsort[-1] = n
-    for i in range(n - 1, -1, -1):
-        min_seen = min_thing_end_argsort[i+1]
-        min_thing_end_argsort[i] = min(min_seen, thing_end_argsort[i])
-
     # first search for the beginning of the interval
     # containers' time is already sorted, but things' endtime is not
     for i, t0 in enumerate(container_start):
-        while left_i <= n - 1 and thing_end_sort[left_i] <= t0 - window:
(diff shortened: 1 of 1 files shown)


WenzDaniel:
Hej @dachengx I am a little bit hesitant in merging your PR as the code gets much more complex. I do not know if we really need to cover every single edge case which is actually forbidden based on the provided doc-string. Did not we just add the consistency check for this very same reason? 

In addition, I think one should try to avoid using `np.append` in a numba function as this makes things slow. At least it did in the past.

dachengx:
> Hej @dachengx I am a little bit hesitant in merging your PR as the code gets much more complex. I do not know if we really need to cover every single edge case which is actually forbidden based on the provided doc-string. Did not we just add the consistency check for this very same reason?
> 
> In addition, I think one should try to avoid using `np.append` in a numba function as this makes things slow. At least it did in the past.

The `append` is removed. Currently, the doc-string allows non-sorted `endtime` of `things` and `containers`. So I think we should cover all of these cases in the function. I do not concern too much about the speed, the new function is ~100% slower than the original one, tested by `%timeit`. But the original function is very fast.

strax/processing/general.py
```diff
@@ -408,34 +408,34 @@ def touching_windows(things, containers, window=0):
 @numba.njit(nogil=True, cache=True)
 def _touching_windows(thing_start, thing_end,
                       container_start, container_end,
-                      window=0):
-    result = np.zeros((len(container_start), 2), dtype=np.int32)
+                      window=0, endtime_sort_kind='mergesort'):
     n = len(thing_start)
+    thing_end_argsort = np.argsort(thing_end, kind=endtime_sort_kind)
+    thing_end_sort = thing_end[thing_end_argsort]
+    container_end_argsort = np.argsort(container_end, kind=endtime_sort_kind)
```
WenzDaniel, line 415: Why can we not use the default sorting algorithm here?

WenzDaniel (approved):
Thanks Dacheng, to me it looks fine. Although I am a little bit concerned performance wise. However, this should now lead to correct results.

strax/processing/general.py
dachengx: To make the sorting stable. We want to return the first index of `things` which is overlapping with `containers`.

PR #738, opened 2023-06-29
Use warning also in `abs_time_to_prev_next_interval`

dachengx:

Sometimes `things` in `abs_time_to_prev_next_interval` might be overlapping, or the `endtime` might be not sorted. So throw warning instead of error, following the logic of `touching_windows`. 

Just change the warning logic, the algorithm is not changed. 

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/processing/general.py
+++ b/strax/processing/general.py
@@ -384,7 +384,7 @@ def touching_windows(things, containers, window=0):
         warnings.warn(
             'endtime of things is not sorted! '
             'touching_windows will return the indices of the '
-            'first and last things which are touching the container')
+            'first and last things which are touching the container.')
     try:
         _check_time_is_sorted(containers['time'])
     except Exception:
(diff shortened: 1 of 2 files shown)


dachengx:
If my understanding is correct, the `abs_time_to_prev_next_interval` will return a valid result if `endtime` of `things` are sorted, but not `things` are not overlapping.

FaroutYLq (approved):
Looks good to me. How often do you expect this warning be triggered? I won't delete log files for this round of reprocessing so we can trace back anywise

dachengx:
> Looks good to me. How often do you expect this warning be triggered? I won't delete log files for this round of reprocessing so we can trace back anywise

For this function, I think you will not see warnings in straxen's plugins(because it is only used on the event level in straxen). But in some customized plugins outside strax(en), it is better to print out the warnings rather than errors just like `touching_windows` and `split_by_containments`.

FaroutYLq:
> For this function, I think you will not see warnings in straxen's plugins(because it is only used on the event level in straxen). 

Sorry I am confused, why I won't see these warning? For reprocessing we goes up to events

dachengx:
> Sorry I am confused, why I won't see these warning? For reprocessing we goes up to events

Events in straxen are not overlapping so that they will not trigger warnings.

dachengx:
> > Sorry I am confused, why I won't see these warning? For reprocessing we goes up to events
> 
> Events in straxen are not overlapping so that they will not trigger warnings.

So this PR is almost only for the users applying the function to their own data where the things might be overlapping.

FaroutYLq:
> Events in straxen are not overlapping so that they will not trigger warnings.

Wait then I think I need to reevaluate this PR: if this `abs_time_to_prev_next_interval` is only used in event level, and in event level we don't expect any overlap, should we keep the original assertion back there?

dachengx:
> Wait then I think I need to reevaluate this PR: if this `abs_time_to_prev_next_interval` is only used in event level, and in event level we don't expect any overlap, should we keep the original assertion back there?

`abs_time_to_prev_next_interval` is used in straxen event level. But it can also be applied to other levels or other data, outside straxen. strax is an independent package from any specified TPC so we would allow users to apply it to any kind of data?

FaroutYLq:
> `abs_time_to_prev_next_interval` is used in straxen event level. But it can also be applied to other levels or other data, outside straxen. strax is an independent package from any specified TPC so we would allow users to apply it to any kind of data?

It is true that `strax` is an independent package... Then should we add no-overlap assertion, correspondingly in `straxen` where `abs_time_to_prev_next_interval` is applied?

dachengx:
> Wait then I think I need to reevaluate this PR: if this `abs_time_to_prev_next_interval` is only used in event level, and in event level we don't expect any overlap, should we keep the original assertion back there?

If we want to test that the events are not overlapping in straxen, we should do it in straxen's `Event` plugin.

FaroutYLq (approved):

WenzDaniel:
Hej guys thanks. The only place we are using this function at the moment is the deadtime calculation in cutax. And here I think nothing can overlap (I hope at least).

FaroutYLq:
Hi @WenzDaniel are you sure? I didn't find anything in `cutax` using this function, but I do find only one in `straxen` [here](https://github.com/XENONnT/straxen/pull/1214)

PR #744, opened 2023-08-01
Refactor nv plugins

WenzDaniel:

Refactor NV/MV hitlets computation. 
1. Removes not needed parameters, to reduce required storage space.
2. Adds a new simple summed waveform function. The function will only compute a biased representation of the summed waveform as the floating point part of the baseline is only partially added. The bias is ~0.5 % and does not play a crucial role when estimating the signal shape


WenzDaniel:
Long time no update making it a draft.

WenzDaniel:
This PR goes together with: https://github.com/XENONnT/straxen/pull/1228

dachengx (approved):
Thank you @WenzDaniel

PR #745, opened 2023-08-04
Waveform attributes

ahiguera-mx:
this function computes quantiles and can downsample waveforms, either for records, peaklets, peak
Compute waveform attribures
Quantiles: represent the amount of time elapsed for a given fraction of the total waveform area to be observed in n_samples  i.e. n_samples = 10, then quantiles are equivalent deciles

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/processing/peak_properties.py
+++ b/strax/processing/peak_properties.py
@@ -102,7 +102,7 @@ def compute_widths(peaks, select_peaks_indices=None):
     peaks['area_decile_from_midpoint'][select_peaks_indices] = fr_times[:, ::2] - fr_times[:, i].reshape(-1,1)
 
 @export
-@numba.njit(nopython=True, cache=True)
+#@numba.njit(nopython=True, cache=True)
 def compute_wf_attributes(data, sample_length, n_samples: int, downsample_wf=False):
     """
     Compute waveform attribures
@@ -111,15 +111,18 @@ def compute_wf_attributes(data, sample_length, n_samples: int, downsample_wf=Fal
     i.e. n_samples = 10, then quantiles are equivalent deciles 
     Waveforms: downsampled waveform to n_samples
     :param data: waveform e.g. peaks or peaklets
-    :param n_samples: number of samples 
+    :param n_samples: compute quantiles for a given number of samples 
     :return: waveforms and quantiles of size n_samples
     """    
     assert len(data) == len(sample_length), "ararys must have same size"
 
+    num_samples = data.shape[1]
+
     waveforms = np.zeros((len(data), n_samples), dtype=np.float64)
     quantiles = np.zeros((len(data), n_samples), dtype=np.float64)
+    # Cannot compute with with more samples than actual waveform sample
+    assert num_samples > n_samples, "cannot compute with more samples than the actual waveform"
 
-    num_samples = data.shape[1]
     step_size = int(num_samples / n_samples)
     steps = np.arange(0, num_samples + 1, step_size)
     inter_points = np.linspace(0., 1. - (1. / n_samples), n_samples)
--- a/strax/processing/peak_properties.py
+++ b/strax/processing/peak_properties.py
@@ -102,7 +102,7 @@ def compute_widths(peaks, select_peaks_indices=None):
     peaks['area_decile_from_midpoint'][select_peaks_indices] = fr_times[:, ::2] - fr_times[:, i].reshape(-1,1)
 
 @export
-#@numba.njit(nopython=True, cache=True)
+@numba.jit(nopython=True, cache=True)
 def compute_wf_attributes(data, sample_length, n_samples: int, downsample_wf=False):
     """
     Compute waveform attribures
--- a/strax/processing/peak_properties.py
+++ b/strax/processing/peak_properties.py
@@ -100,3 +100,48 @@ def compute_widths(peaks, select_peaks_indices=None):
     i = len(desired_fr) // 2
     peaks['width'][select_peaks_indices] = fr_times[:, i:] - fr_times[:, ::-1][:, i:]
     peaks['area_decile_from_midpoint'][select_peaks_indices] = fr_times[:, ::2] - fr_times[:, i].reshape(-1,1)
+
+@export
+@numba.njit(nopython=True, cache=True)
+def compute_wf_attributes(data, sample_length, n_samples: int, downsample_wf=False):
+    """
+    Compute waveform attribures
+    Quantiles: represent the amount of time elapsed for
+    a given fraction of the total waveform area to be observed in n_samples
+    i.e. n_samples = 10, then quantiles are equivalent deciles 
+    Waveforms: downsampled waveform to n_samples
+    :param data: waveform e.g. peaks or peaklets
+    :param n_samples: number of samples 
+    :return: waveforms and quantiles of size n_samples
+    """    
+    assert len(data) == len(sample_length), "ararys must have same size"
+
+    waveforms = np.zeros((len(data), n_samples), dtype=np.float64)
+    quantiles = np.zeros((len(data), n_samples), dtype=np.float64)
+
+    num_samples = data.shape[1]
+    step_size = int(num_samples / n_samples)
+    steps = np.arange(0, num_samples + 1, step_size)
+    inter_points = np.linspace(0., 1. - (1. / n_samples), n_samples)
+    cumsum_steps = np.zeros(n_samples + 1, dtype=np.float64)
+    frac_of_cumsum = np.zeros(num_samples + 1)
+    sample_number_div_dt = np.arange(0, num_samples + 1, 1)
+
+    for i, (samples, dt) in enumerate(zip(data, sample_length)):
+        # reset buffers
+        frac_of_cumsum[:] = 0
+        cumsum_steps[:] = 0
+
+        frac_of_cumsum[1:] = np.cumsum(samples)
+        frac_of_cumsum[1:] = frac_of_cumsum[1:] / frac_of_cumsum[-1]
+
+        cumsum_steps[:-1] = np.interp(inter_points, frac_of_cumsum, sample_number_div_dt * dt)
+        cumsum_steps[-1] = sample_number_div_dt[-1] * dt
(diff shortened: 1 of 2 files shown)


strax/processing/peak_properties.py
```diff
@@ -100,3 +100,52 @@ def compute_widths(peaks, select_peaks_indices=None):
     i = len(desired_fr) // 2
     peaks['width'][select_peaks_indices] = fr_times[:, i:] - fr_times[:, ::-1][:, i:]
     peaks['area_decile_from_midpoint'][select_peaks_indices] = fr_times[:, ::2] - fr_times[:, i].reshape(-1,1)
+
+@export
+#@numba.njit(nopython=True, cache=True)
```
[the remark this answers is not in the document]
ahiguera-mx: Hi Daniel
Thanks for checking at this, the functions calculate different things
compute_index_of_fraction() return the index at which certain area has been achieved if the fraction is 1 it will return peak['length'] so this is in units of samples.

compute_wf_attributes() for a given set of quantiles (fractions) it return the the amount of time elapsed for a given fraction of the total waveform area to be observed

So in other words, compute_index_of_fraction() is in units of samples and compute_wf_attributes() is in units of ns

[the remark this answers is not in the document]
ahiguera-mx: also `compute_wf_attributes()` can handle an array of peaks whereas `compute_index_of_fraction()` can do only peak by peak
Do you think it would better to merge into one function?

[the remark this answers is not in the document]
dachengx: Looking at these codes, `compute_index_of_fraction` can also return floats. I think it is better to utilize `compute_index_of_fraction` or modify it.
https://github.com/AxFoundation/strax/blob/9b508f7f8d441bf1fe441695115d292c59ce631a/strax/processing/peak_properties.py#L48

LuisSanchez25:
@jmosbacher This is the strax pull request we need for the other pluggin, if you could help us with this that would be great, I will work on updating the private_repository and the straxen pull request in the mean time.

jmosbacher:
@LuisSanchez25 it doesnt look like you have addressed the comments from @dachengx and @WenzDaniel which seem valid. 
From glancing at the code, to me it also looks like the following holds
```python
compute_wf_attributes(...) == compute_index_of_fraction(...) * dt
```
So I think you need to demonstrate that this equality does not hold or give some other justification for adding this extra method before this can be merged.

LuisSanchez25:
Yes, I have not addressed this yet, Aaron and I discussed this a bit the other day so I was hoping we could talk to you about it at some point maybe tomorrow? I need a bit of time to check a few things here since I was not the main person working on this PR but wanted to make you aware of which PR we were talking about.

strax/processing/peak_properties.py
```diff
@@ -100,3 +100,52 @@ def compute_widths(peaks, select_peaks_indices=None):
...
+        # reset buffers
+        frac_of_cumsum[:] = 0
+        cumsum_steps[:] = 0
+        frac_of_cumsum[1:] = np.cumsum(samples)
+        frac_of_cumsum[1:] = frac_of_cumsum[1:] / frac_of_cumsum[-1]
+        cumsum_steps[:-1] = np.interp(inter_points, frac_of_cumsum, sample_number_div_dt * dt)
+        cumsum_steps[-1] = sample_number_div_dt[-1] * dt
+        quantiles[i] = cumsum_steps[1:] - cumsum_steps[:-1]
+   
+        if downsample_wf:
```
jmosbacher, line 146: this whole downsampling makes no sense to me. 

- If i pass `downsample_wf=False` to this function it will return an array of zeros as waveforms. in what case would that be useful?
- Why wouldnt you infer downsampling==True if num_samples > n_samples?
- What if (num_samples / n_samples) is not an interger? wouldnt the last bin sum over a different number of samples? so why is `waveforms[i] /= (step_size * dt)` correct?

LuisSanchez25:
> @LuisSanchez25 it doesnt look like you have addressed the comments from @dachengx and @WenzDaniel which seem valid. From glancing at the code, to me it also looks like the following holds
> 
> ```python
> compute_wf_attributes(...) == compute_index_of_fraction(...) * dt
> ```
> 
> So I think you need to demonstrate that this equality does not hold or give some other justification for adding this extra method before this can be merged.

So these two things are not the same as the compute_wf_attributes as this will just 

> @LuisSanchez25 it doesnt look like you have addressed the comments from @dachengx and @WenzDaniel which seem valid. From glancing at the code, to me it also looks like the following holds
> 
> ```python
> compute_wf_attributes(...) == compute_index_of_fraction(...) * dt
> ```
> 
> So I think you need to demonstrate that this equality does not hold or give some other justification for adding this extra method before this can be merged.

So those two things are not the same as the compute_wf_attributes gives back a particular index for a given percentage (like at 10% 20%, ect...) the function we have wants to compute the deciles so (between 0-10%, then 10-20%, ect) so there is a subtraction going on so doing a * dt wont cut it. We can still use the current function but the other key difference i

> @LuisSanchez25 it doesnt look like you have addressed the comments from @dachengx and @WenzDaniel which seem valid. From glancing at the code, to me it also looks like the following holds
> 
> ```python
> compute_wf_attributes(...) == compute_index_of_fraction(...) * dt
> ```
> 
> So I think you need to demonstrate that this equality does not hold or give some other justification for adding this extra method before this can be merged.

So these two functions are not the same, compute_index_of_fraction will give back an index for between 0 and 10% or 0 and 30% of the waveforms, the compute_wf_attributes function computes the index between deciles, (so 0-10%, 10-20%, ect...) so multiplying it by the dt wont cut it as we need to subtract subsequent values. We can still use the compute_index_of_fraction but the other difference is that the compute_index_of_fraction computes 1 waveform attribute at a time while the compute_wf_attributes does everything in one go. This is a functionality we could move to straxen when the plugin is called but we think it makes more sense to leave it here as this would be used by 2 plugins, the Bayes plugin and the SOM plugin

strax/processing/peak_properties.py
LuisSanchez25: This is because the bayes pluggin needs the waveform but the SOM plugin does not. So if we dont need it we can save computational time by not calculating this, but if you know of a better way to achive this goal I am more than happy to change this.

As to your last point I do not really understand what you are saying, can you elaborate a bit more?

jmosbacher: Lets take a simple example.

```python
num_samples = 10
n_samples = 6
step_size = int(num_samples / n_samples)
steps = np.arange(0, num_samples + 1, step_size)

for j in range(n_samples):
    print(steps[j], ":", steps[j + 1])

0 : 1
1 : 2
2 : 3
3 : 4
4 : 5
5 : 6
```
Does this look like correct downsampling to you?

LuisSanchez25: Ok I think I know what you are saying now. So this is dealing with waveforms so num_samples will always be 200, but there are no safeguards against non integer divisions of the data. We can make an assert statement to make sure num_samples is divisible by n_samples. I will make that change.

The waveforms[i] /= (step_size * dt) I am now a bit confused about, the objective is to get rid of the dt dependance so we can compare all wavefroms to each other, but I did some quick dimensional analysis and I think we should be multiplying by dt instead of dividing since waveform amplitude is in units of PE/sample = PE/(dt) so we should be multiplying the numerator by dt. @ahiguera-mx since you worked on this more with the Bayes classification is there something we are missing?

jmosbacher: - Like I explained when you originally asked me where to place this function. Strax is an open source framework built to be used by experiments other than XENON.  Function in strax cannot assume the data is exactly as it is in xenon.
- The required bar in terms of code quality to have a PR merged is also higher. 
- Allocating a massive array of zeros just to be used sometimes is not a good approach. You should downsample ina different function. There are many standard techniques for downsampling, such as windowed reductions. I would stick to one of those.

LuisSanchez25: Ok we dont need the downsampling for now so we can just remove that funcitonality, and make an assert statement to make sure data quantiles make sense, but the big question still is, given the difference between the functions, do we still want to try to merge it into one or should I leave them as separate things?

jmosbacher: I would try to either:
-  use the existing function in the plugin or adapt it to be able to produce what you need
- Provide some actual concrete example that demonstrates the need for the extra function. I find the hand wavy explanations very difficult to follow.

LuisSanchez25:
On a completely separate note, is there a reason there are no tests for the function compute_index_of_fraction in test_peak_properties?

WenzDaniel (changes requested):
I am sorry, I cannot reproduce what the function is supposed to do. At least my result is different from the expectation I get from the doc-string. Can you provide a full example with a full description of what I should expect to get by the function?

WenzDaniel:
Okay after some trial and error I think I figured out how to use the function, but the results looks off to me. If I run:
```python
res, _ = strax.compute_wf_attributes(peaklets[:2]['data'], peaklets[:2]['dt'], 10)
```
I get the following result:

example

where the blue lines are the cumulative of your returned deciles. The last one is pretty far off and even goes beyond the true peak.

I can achieve the very same with:

```python
res2 = strax.index_of_fraction(peaklets[:2], np.arange(10.1)/10)
res2 = (res2.T * peaklets[:2]['dt']).T
res2_compute_diff = np.diff(res2, axis=1)
```

which is 3 lines of code. The blue lines are the deciles 0 -> 0.1, 0 -> 0.2 etc. as computed by `index_of_fraction`. The red lines are the cumulative of the difference of the deciles, so the same as you function returns. 

example2

The result is also a bit off compared to the deciles themselves, but much closer than your function. So I honestly cannot follow when you are saying that we cannot use this already implemented function.  

In addition, the function is tested once here: https://github.com/AxFoundation/strax/blob/21cc96e011b0e4099138979791f34e8b1addedb7/tests/test_peak_properties.py#L46
and once indirectly here: https://github.com/AxFoundation/strax/blob/21cc96e011b0e4099138979791f34e8b1addedb7/tests/test_peak_properties.py#L66

WenzDaniel:
We should make a new PR with a revised waveform attributes function.

PR #746, opened 2023-08-07
Add small selection functions

WenzDaniel:
In this PR we extend the selection_str feature to allow to pass arbitrary small or more complex functions for the data selection, e.g., one could pass a function like:

```python
def s2_width_cut(events):
    ....
st.get_array(run, 'event_info', selection=s2_width_cut)
``` 

The selection function must take only a single argument which represents the data array itself. For example functions which take additional parameters must be reduced first:

```python
def s2_width_cut(events, para1, para2):
    ....
st.get_array(run, 'event_info', selection=lambda x: s2_width_cut(x, para1=1, para2=2))
```

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1205,6 +1205,7 @@ class Context:
                  seconds_range=None,
                  time_within=None,
                  time_selection='fully_contained',
+                 selection=None,
                  selection_str=None,
                  keep_columns=None,
                  drop_columns=None,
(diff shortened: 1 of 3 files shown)


WenzDaniel:
Maybe we should keep the option as selection_str. Otherwise we may break some analysts code.

dachengx (approved):
Looks good. Thanks @WenzDaniel .

Issue #747, opened 2023-08-10
show_config messes with plugin registry!


WenzDaniel:
**Describe the bug**
Somehow show config messes with the plugin registry such that `st._get_plugins` returns a wrong set of plugins. This is a pretty big bug.

MWE:

```python
st._get_plugins(('records'), '0')  # normal return
st.show_config('event_info')
st._get_plugins(('records'), '0')  # Returns all plugins needed for event info instead of records...
```

jmosbacher:
Nice catch, looks like the `Context._get_plugins` method ignores the targets when they are all found in the plugin cache and just returns all of the plugins given by  `Context.__get_plugins_from_cache` [here](https://github.com/AxFoundation/strax/blob/9b508f7f8d441bf1fe441695115d292c59ce631a/strax/context.py#L639)

WenzDaniel:
The issue was actually introduced with the plugins caching in: https://github.com/AxFoundation/strax/pull/483

jmosbacher:
I guess the places where it is used either select a specific plugin from the dict (like the key_for etc.) or further extract only the plugins based on what needs to be reprocessed (get_components) since some data can already exist. 
The only effect is probably a small performance hit because the `get_components` will call `Plugin.setup` on all the plugins before they are filtered out. 

Looks like an easy fix, are you already working on a PR or should I make one?

WenzDaniel:
Yes I also checked get_components, we are looping over all provided plugins and only take the required ones. 

I am working already on it thanks.

PR #748, opened 2023-08-10
Patch plugin cache

WenzDaniel:
Adds fix and test for https://github.com/AxFoundation/strax/issues/747

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -636,7 +636,19 @@ class Context:
         instance, which is referenced under multiple keys in the output dict.
         """
         if self._plugins_are_cached(targets):
-            return self.__get_plugins_from_cache(run_id)
+            cached_plugins = self.__get_plugins_from_cache(run_id)
+            plugins = {}
+            targets = list(targets)
+            for target in targets:
+                if target in plugins:
+                    continue
+                    
+                target_plugin = cached_plugins[target]
+                for provides in target_plugin.provides:
+                    plugins[provides] = target_plugin
+            
+                targets += list(target_plugin.depends_on)
+            return plugins
 
         # Check all config options are taken by some registered plugin class
         # (helps spot typos)
--- a/strax/context.py
+++ b/strax/context.py
@@ -639,14 +639,14 @@ class Context:
             cached_plugins = self.__get_plugins_from_cache(run_id)
             plugins = {}
             targets = list(targets)
-            for target in targets:
+            while targets:
+                target = targets.pop(0)
                 if target in plugins:
                     continue
-                    
+            
                 target_plugin = cached_plugins[target]
                 for provides in target_plugin.provides:
                     plugins[provides] = target_plugin
-            
                 targets += list(target_plugin.depends_on)
             return plugins
 
--- a/strax/testutils.py
+++ b/strax/testutils.py
@@ -208,6 +208,30 @@ class Peaks(strax.Plugin):
         return p
 
 
+@strax.takes_config(
+    strax.Option('base_area', type=int, default=0),
+    strax.Option('give_wrong_dtype', type=bool, default=False),
+    strax.Option('bonus_area', type=int, default=0))
+class PeaksWoPerRunDefault(strax.Plugin):
+    """Same as peak plugin but without per run default option
+    to allow for plugin caching.
+    """
+    provides = 'peaks'
+    data_kind = 'peaks'
+    depends_on = ('records',)
+    dtype = strax.peak_dtype()
+    parallel = True
+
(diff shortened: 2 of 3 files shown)


WenzDaniel:
Checked also with different straxen plugins that things are now working as expected.

jmosbacher (changes requested):
Looks good, thanks. 
The only thing that bothers me a bit is modifying `targets` inside the loop iterating over it. I dont think this is defined behavior by the python spec so a future implementation of the list object may break it. Especially since python will be removing the GIL soon, CPython core implementations are expected to change in the near future.
Maybe go with the standard way of doing this
```python
while targets:
    target = targets.pop(0)
    ...
```

WenzDaniel:
Sorry, based on your comment I am not sure if you whether you think the for-loop, or the appending-to-the-same-list is the problem. The former I agree, can be changed into a while loop. The latter cannot be changed as I do not know my full list of targets at the beginning of the loop. So using pop does not work ( At least I cannot see how this would work). The only other solution I could think about is defining some "global" dictionary and do some recursive function call like it is done in `_get_plugins` if we are not loading things form cache.

WenzDaniel:
So something like this:

```python
plugins = {}

def _get_depndent_plugins(targets):
    for target in targets:
        target_plugin = cached_plugins[target]

        for provides in target_plugin.provides:
            plugins[provides] = target_plugin
    
        [_get_depndent_plugins((target,)) for target in target_plugin.depends_on]

_get_depndent_plugins(targets)
```

jmosbacher:
Sorry if I wasnt explicit enough. for loops in python use iterators, the next value is returned from the iterator object and not the original object. In many list-like implementations including the current implementation of the builtin list, the iterator object returned is the original object so modifying it inside the loop can work but its undefined behavior that can change with the implementation of list in CPython.

The while loop does not create an iterator, and therefore should always work.
to be explicit in the proposed implementation:

```python
while targets:
    target = targets.pop(0)
    if target in plugins:
        continue

    target_plugin = cached_plugins[target]
    for provides in target_plugin.provides:
        plugins[provides] = target_plugin
    targets += list(target_plugin.depends_on)

```

WenzDaniel:
I am learning new things now, nice. How does the while loop then work if not with iterators? (It is hard to find any information about this by simply searching for it. You always get things like "how to write a while loop in python...") So in a while loop the iteration is not done on the original object but the modified one?

jmosbacher:
the while loop will re-evaluates the expression for thuthyness at every iteration and executes the code body if the expression is true, calling __bool__ method if its an object of a different type than bool. 
lists thruthyness has been defined in the python spec to return true if the list has non-zero length for a while now so no need to use `len(targets)` unless you want to support very old python versions.

WenzDaniel:
Ah sorry, no  I was more curious why

```python
while list:
    item = list.pop(0)
    list.append(...) #modify list
```
is okay but
```python
for item in list:
    list.append(...) #modify list
```
is not. For me both are iterators. I understood from your comment above (maybe wrongly), that a for-loop is only defined to loop over the original "list" without it being modifications, and in the while loop "list" is reevaluated after each iteration.

jmosbacher:
The python [spec](https://docs.python.org/3/reference/compound_stmts.html#the-while-statement)

jmosbacher:
Maybe the confusion is on what an iterator is in python?
The iterator type as defined [here](https://docs.python.org/3/library/stdtypes.html#iterator-types) has both `__iter__` and the `__next__` methods defined. These are the iteration mechanism calls. the `__iter__` method is called once at the beginning to set things up and then the `__next__` method is called at every iteration to fetch the next object.

The for loop first evaluates the expression to get an iterator (calling `__iter__` on it if its an object) and then begins the iteration. at every step of the iteration, the `__next__` method is called and the loop ends if the StopIteration exception is raised during this call.

WenzDaniel:
Yes this I knew, and due to this I would have though that modifying the list in a for loop would be save. Because the iterator only knows about the next object while iterating and only stops if `StopIteration` is raised. So I would have thought that adding things to my object I am iterating over would be fairly safe.

jmosbacher:
Ah yes, now i understand. The issue is that there is nothing in the spec that tells you to return the original object from `container.__iter__` and not a copy. So e.g. when the GIL is removed, CPython may switch to returning a copy to avoid write collisions from different threads etc. so better to be safe than sorry :)

WenzDaniel:
Thanks, was a nice discussion :D I can already go home now as I learned something today :D

jmosbacher:
oops, forgot to approve :)

PR #752, opened 2023-08-27
Check chunk size right after loading chunk

dachengx:

Check the size of chunk when loading data.

Add a decorator `check_chunk_n` for `strax.StorageBackend._read_chunk`.

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -442,7 +442,7 @@ def check_chunk_n(f):
         chunk_info = kwargs['chunk_info']
         chunk = f(self, *args, **kwargs)
         if len(chunk) != chunk_info['n']:
-            raise ValueError(
+            raise strax.DataCorrupted(
                 f"Chunk {chunk_info['filename']} of {chunk_info['run_id']} has {len(chunk)} items, "
                 f"but chunk_info {chunk_info} says {chunk_info['n']}")
         return chunk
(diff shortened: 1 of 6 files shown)


WenzDaniel:
Ah sorry, by size you mean number of chunks, sorry. Now I get it.

strax/chunk.py
```diff
@@ -425,3 +426,18 @@ def _update_subruns_in_chunk(chunks):
             else:
                 subruns[subrun_id] = subrun_start_end
     return subruns
+
+
+@export
+def check_chunk_n(f):
+    @wraps(f)
+    def wrapper(self, *args, **kwargs):
+        # assume chunk_info is the second argument
```
WenzDaniel, line 435: an we add a check that 'chunk_info' is in kwargs?

WenzDaniel (commented):
Hej Dacheng looks good to me. Just one small comment, and a bigger one. Can you please add a test which fails and raises your warning? You can use our Records test plugin for that, but you will need to manually manipulate the chunk ones it is made. It will suffice to test it only for the file system backend

dachengx:
Yes, @WenzDaniel . I added the test for `check_chunk_n`.

WenzDaniel (approved):
Thanks, looks good

PR #758, opened 2023-09-16
Mandatorily wrap `_read_chunk` in a `check_chunk_n` decorator

dachengx:

To ease the workflow of developing a new `StorageBackend`, the PR mandatorily wraps `_read_chunk` in a `check_chunk_n` decorator instead.

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/storage/common.py
+++ b/strax/storage/common.py
@@ -403,6 +403,13 @@ class StorageBackend:
     these have to be hardcoded (or made part of the key).
     """
 
+    def __new__(cls, *args, **kwargs):
+        """Mandatorily wrap _read_chunk in a check_chunk_n decorator"""
+        if '_read_chunk' in cls.__dict__:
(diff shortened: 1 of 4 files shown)


JYangQi00:
Thanks Dacheng,

I just want to make sure I'm understanding this correctly, so please correct me if I am wrong. The point of this PR is just to reduce code duplication by making the wrapper apply to _every_ instance of a `StorageBackend` via the `__new__` class, correct? If so, this PR looks good to me and I will accept

dachengx:
> Thanks Dacheng,
> 
> I just want to make sure I'm understanding this correctly, so please correct me if I am wrong. The point of this PR is just to reduce code duplication by making the wrapper apply to _every_ instance of a `StorageBackend` via the `__new__` class, correct? If so, this PR looks good to me and I will accept

Thanks. Yes, you are correct.

PR #762, opened 2023-09-30
Proposal to use pre-commit for continuous integration

dachengx:
## What does the code in this PR do / what does it improve?

Use `pre-commit`'s continuous integration.

The advantage of `pre-commit` is that:

1. It runs automatically when you `git commit`.
2. You will not upload codes that are not aligned with some styles.

The codes are changed only in code style, the requirement of hooks.

Note: `black` will force you to use double quotes, it actually auto-format for you, so do not worry.

You will see no reviewdog complaining, and pre-commit will do almost all formatting for you, and it is very fast.

## Can you briefly describe how it works?

To setup pre-commit:

```
cd strax
pip install pre-commit
pre-commit install
```

Besides basic whitespace and large files check,
1. `black` will **auto-format** your code, mainly about indent and line-length, and make auto-commit if necessary.
2. `docformatter` will **auto-format** your docstring, following google convention, and make auto-commit if necessary.
3. `flake8` will **check** the code-style, but in a more lite way than `wemake-python-styleguide`.

So there will be two pre-commit checks, one check of **changed files** when you `git commit`, and one check of **all files** by pre-commit CI.

## Can you give a minimal working example (or illustrate with a figure)?

When you `git commit`, it will automatically `pre-commit run` for the **changed** files, so actually you would not always run it manually.

If you want to check all the files, run `pre-commit run --all-files`. If you want to check the difference between the current branch and `main`, run `pre-commit run --from-ref main --to-ref .`. If you want to remove the color from the output, add `--color never`.

`black` and `docformatter` will change your code, you still need to make a commit by yourself on the changed part.

For more information, please go to https://pre-commit.com/ and https://pre-commit.ci/.

_Please include the following if applicable:_
  - [x] _Update the docstring(s)_
  - [ ] _Update the documentation_
  - [x] _Tests to check the (new) code is working as desired._
  - [ ] _Does it solve one of the open issues on github?_

Inspired by https://github.com/XENONnT/straxen/issues/1192

PR #769, opened 2023-10-24
Add chunk yielding plugin and tests

WenzDaniel:
Adds new type of plugin which allows compute method to return a generator instead of chunks. This allows to return sub-chunks early, in case the chunk of the new data types becomes too large. This is needed for example when simulating events and computing from a list of photon hits raw waveforms.


WenzDaniel:
Tested processing on a small run `025423` results are equivalent to strax master.

strax/plugins/down_chunking_plugin.py
```diff
@@ -0,0 +1,40 @@
...
+##
+
+@export
+class DownChunkingPlugin(Plugin):
+    """Plugin that merges data from its dependencies
+    """
+
+    def iter(self, iters, executor=None):
+
+        _plugin_uses_multi_threading = (self.parallel
```
JelleAalbers, line 21: Could we more simply assert/check that `not self.parallel` and `inspect.isgeneratorfunction(self.compute)` here? If compute is not a generator, you don't need this base class, and if it is, parallelization won't work and `parallel=False` seems right. 

If so, could do the checks in `__init__` (after `super.__init__` probably) since they don't depend on the executor anymore?

strax/plugins/down_chunking_plugin.py
```diff
@@ -0,0 +1,40 @@
...
+        if _plugin_uses_multi_threading:
+            raise NotImplementedError(
+                f'Plugin "{self.__class__.__name__}" uses an iterator as compute method. ' 
+                'This is not supported in multi-threading/processing.')
+        return super().iter(iters, executor=None)
+
+    def _iter_return(self, chunk_i, **inputs_merged):
+        return self.do_compute(chunk_i=chunk_i, **inputs_merged)
+
+    def _fix_output(self, result, start, end, _dtype=None):
```
JelleAalbers, line 34: Am I right that `_fix_output` is currently skipped by these plugins, and thus multi-output is not supported? That could be another check at the beginning (`assert not self.multi_output` or maybe a nicer `NotImplementedError`). You could also make `_fix_output` equal to `pass` -- or is there a reason you did it like this?

strax/plugins/plugin.py
```diff
@@ -523,6 +523,10 @@ class IterDone(Exception):
         finally:
             self.cleanup(wait_for=pending_futures)
 
+    def _iter_return(self, chunk_i, **inputs_merged):
```
JelleAalbers, line 526: Maybe add a docstring or comment to say this should either return or generate result chunks from the input? (Maybe _iter_compute is an alternative name)

strax/testutils.py
```diff
@@ -260,6 +260,74 @@ def compute(self, peaks):
         return dict(peak_classification=p,
                     lone_hits=lh)
 
+        
+# Plugins with time structure within chunks,
```
JelleAalbers, line 264: Am I right that this is identical to Records from testutils, apart from setting the record times differently? Could we inherit from that one (maybe after modifying it) to reduce duplication? 

(Maybe this is not possible of course; I vaguely recall something about inheritance messing up config options.)

tests/test_context.py
```diff
@@ -215,6 +215,34 @@ def tearDown(self):
         if os.path.exists(self.tempdir):
             shutil.rmtree(self.tempdir)
 
+    def test_down_chunking(self):
```
JelleAalbers, line 218: Not exactly sure how you would like to organize tests, but the comment on this test class says it's about testing per-run defaults of the context. I guess that's outdated, but it looks like it's least about texting context-related stuff, not particular plugin types.

Plugin things seem to be tested in test_core.py or in separate test files (test_overlap_plugin.py, test_multi_output.py), where they make dummy contexts when needed.

JelleAalbers (commented):
This implements a new `DownChunkingPlugin` base class, for plugins that produce multiple chunks iteratively from a single `compute` call. 

The main use case is plugins that create data that is much more itense in memory/time than the input, e.g. simulating waveforms from photon arrival times. Currently, you instead have to ensure the input is read in in small enough chunks. If memory isn't an issue but you want smaller files on disk, I suppose you could also do that with a modified Saver. But since memory is smaller than the disk, I guess that wouldn't help you much.

I left some comments but I definitely like the idea. This just makes a very small modification to the plugin base class so there should be no impact on other things.

There may be some ways to generalize this in the future, e.g. track the last produced chunk time in the base class so the user doesn't have to, or make a plugin type where compute _gets_ iterators in its arguments too (like the `iters` in `Plugin.iter`, but time-synchronized), so it can keep state in local variables rather than instance variables.

(I didn't look at the code style since #762 is likely incoming, but happy to do that if you like.)

dachengx:
I will not merge master into this now since that will mess up Jelle's review. Let's merge master into this PR later.

WenzDaniel:
Thank you very much @JelleAalbers for your thorough review. I agree with all your suggestions and I think I adopted all of them. Most of your points raised are originating from the fact that first I modified the strax.Plugin class in https://github.com/AxFoundation/strax/pull/751, before we decided to move it into a separate plugin class. (Except for the organization of our tests here I was simply lazy, but I moved the tests now in a separate file like we did with CutPlugins.)

JelleAalbers (approved):
Thanks Daniel -- Looks good!

PR #772, opened 2023-11-02
Guard hitlet entropy test from numerical errors

JelleAalbers:
... by avoiding waveforms for which normalization is not reliable in float32.

While installing strax I got a failure in one of the hitlet tests that uses hypothesis to generate test data. This is the same test as in #539 / #544. The PR makes sure the test is skipped on data that is very susceptible to floating point errors. This avoids failures from slight implementation differences between  the numba code and the numpy code we test against.

The waveform that triggered the failure was something like [1, -1, 1e-5] (with a few extra zeros). Note the amplitude is much larger than its sum, causing numerical errors in float32 (e.g. the first step in the conditional entropy computation is to divide by the waveform sum), and about ~1% differences between the numpy and numba computation for waveforms like the above.

I arbitrarily put the threshold for skipping the test at sum < 1e-4 * min-to-max amplitude. Now I get no more failures from hypothesis even with a max_examples of several thousand.


WenzDaniel (approved):
Thanks @JelleAalbers

dachengx (approved):
Thanks @JelleAalbers

PR #776, opened 2023-11-20
Add warning

WenzDaniel:
Add a small user warning if a plugin is deregistered as this is should be not a normal use case. Also converts list of plugins to deregistered into a set to avoid duplication. Small addition to: https://github.com/AxFoundation/strax/pull/775/files


JelleAalbers:
This is a good idea; I had this in #775 earlier but it generated quite a few messages when starting a cutax context. But maybe that's OK, or you want to merge this after fixing straxen/cutax first.

WenzDaniel:
For straxen I did: https://github.com/XENONnT/straxen/pull/1292
For cutax this is a bigger story because we are auto-registering cuts which we should not do in my opinion. But should be solved with: https://github.com/XENONnT/cutax/pull/317 I think.

JelleAalbers (approved):
Looks good! I assume you want to wait until at least https://github.com/XENONnT/cutax/pull/317 is merged to avoid a swarm of warning messages, but from the strax side I don't see any problems.

(Well, there is a typo in the message, `overlap dregister` should probably be `overlaps registered`, but I guess it reinforces the message that you should not be seeing this message ;-)

The `set` doesn't change any behavior I think due to the check `if old_plugin == currently_registered`, but I think it make things clearer. (The purpose is to deregister plugins once for each datatype.)

WenzDaniel:
Thanks @JelleAalbers, I fixed the typo. Regarding the `set`, I was under the impression that otherwise the error is raised multiple times, but I could be wrong now.

Issue #770, opened 2023-10-26
Extend compare_metadata


WenzDaniel:
We should add a fall back method to https://github.com/AxFoundation/strax/blob/5f9051bcb4fb4a8d46e5fd3da21e522959735352/strax/context.py#L1682 which only compares the lineages in case the current data-type not stored. Currently, we are loading the metadata via:
```python
# new metadata for the given runid + target; fetch from context
 new_metadata = self.get_metadata(run_id, target)
```
We should as fall back in case data is not stored:
```python
new_lineage = st.key_for(run_id, data-type).linegae
```
One needs to only cast all tuples into lists or vise-versa or deactivate the type checking as otherwise one cannot compare to metadata files loaded from disk.__

KaraMelih:
I started looking into this. It looks straightforward, and I think I just now understood what you meant by typecasting. Just to confirm;

```python
a = st.get_metadata(run_id, target)['lineage']
b = st.key_for(run_id, target).lineage
strax.utils.compare_dict(a, b)
```
returns the same information but one of them returns them in a list while the other returns them in a tuple.

Then I need to check how deep the lists and tuples are nested to convert them.

KaraMelih:
Should I make a PR branching from this commit 5f9051b ? 
The version on the master is different

KaraMelih:
I have refactored the code to check lineages in case the data is not stored. I also wrote the following utility function for converting tuples at varying-depth nested objects.

```python
import copy
def convert_tuple_to_list(init_func_input):
    func_input = copy.deepcopy(init_func_input)
    # if it is a tuple convert it and reiterate
    if isinstance(func_input, tuple):
        _func_input = list(func_input)
        return convert_tuple_to_list(_func_input)
        
    # if it is a list, go over all the elements until all tuples are lists
    elif isinstance(func_input, list):
        new_func_inp = []
        # check each element
        for i in func_input:
            new_func_inp.append(convert_tuple_to_list(i))
            # iterates until everything is all depths are exhausted
        return new_func_inp
        
    # if it is a dict iterate
    elif isinstance(func_input, dict):
        for k,v in func_input.items():
            func_input[k] = convert_tuple_to_list(v)
        return func_input
    else:
        # if not a container, return. i.e. int, float, bytes, str etc.
        return func_input
```

I tested it and seems to be working as intended. 

[...]

PR #778, opened 2023-11-27
Upgrade compare-metadata function

KaraMelih:

The previous version of the `st.compare_metadata()` was accepting one runid+target pair and one already existing metadata JSON file as was pointed here: #770 by @WenzDaniel 
This upgrade allows for two sets of //either// runid+target pairs or existing metadata in the form of a path to the JSON file or already loaded dictionary, to do the comparison. In case the data is not loaded anywhere (e.g. peak positions) instead of doing a full metadata comparison, it just compares the lineages.
It also allows to return of extracted metadata and lineage of both requested data for convenience, if `return_results=True` is passed.

You get your favorite context, and call the attribute;
```
context.compare_metadata(  ("053877", "peak_basics"), ("053877", "peak_positions"))
context.compare_metadata(  ("053875", "peak_basics"), ("053877", "peak_basics"))
context.compare_metadata(  ("053877", "peak_basics"),  "./my_path_to/JSONfile.json")
first_metadata = context.get_metadata(run_id, "events")
context.compare_metadata(  ("053877", "peak_basics"), first_metadata)
context.compare_metadata(  ("053877", "records"), ("053899", "records") )
```

```python
import cutax
st = cutax.contexts.xenonnt_online(_rucio_local_path='/project/lgrandi/rucio', include_rucio_local=True, include_rucio_remote=True)
st.compare_metadata(  ("053875", "peak_basics"), ("053877", "peak_basics") )
```

  -  Update the docstring(s) 
  - Update the documentation
  - Does it solve one of the open issues on github? Solves #770


KaraMelih:
@WenzDaniel could you help me understand this failing test? 
https://www.codefactor.io/repository/github/axfoundation/strax/pull/778
it refers to two definitions that do not come from this PR but already existing

WenzDaniel:
Hej @KaraMelih thank you very much for the update I think your PR goes in a very different direction than I had in my mind. The issue we had is, that if you compare the metadata of the currently used context with some old data stored somewhere that you can only perform this comparison if the metadata of your currently used context is stored somewhere. However, I think most of the time the use case is the other way around: I have a context and I cannot load any data anymore and I would like to understand why my context is different than the one used for the stored data. So only adding a simple try/except would have been already sufficient. Sorry for all the extra work. I will look into your PR now.

strax/context.py
```diff
@@ -1844,30 +1844,89 @@ def get_meta(self, run_id, target) -> dict:
 
     get_metadata = get_meta
 
-    def compare_metadata(self, run_id, target, old_metadata):
+    def compare_metadata(self, data1, data2, return_results=False):
         """Compare the metadata between two strax data.
```
WenzDaniel, line 1848: Can you add to the doc string which explains the pintout of the comparison. I was a little confused which direction things are compared is red the first and green the second file?

strax/context.py
```diff
@@ -1844,30 +1844,89 @@ def get_meta(self, run_id, target) -> dict:
...
 
-        :param run_id: run id to get
-        :param target: data type to get
-        :param old_metadata: path to metadata to compare, or a dictionary, or a tuple with another
-            run_id, target to compare against the metadata of the first id-target pair
+        :param data1, data2: either a list (tuple) of runid + target pair, or path to metadata to
```
WenzDaniel, line 1850: Can you add into your doc string the examples you provided here in github. Then I think it becomes more clea, like:

```python
"""
doc-string so far

    Examples:
    example code 1
    example code 2
"""
```

strax/context.py
```diff
@@ -1844,30 +1844,89 @@ def get_meta(self, run_id, target) -> dict:
...
+            available."""
+            # if the runid+target pair is given, check if stored
+            if metafile is None:
+                _is_stored = self.is_stored(run_id, target)
+                metadata = self.get_metadata(run_id, target) if _is_stored else None
+                lineage = (
+                    metadata["lineage"] if _is_stored else self.key_for(run_id, target).lineage
+                )
+                _lineage_hash = str(self.key_for(run_id, target))
+                print(_lineage_hash)
```
WenzDaniel, line 1890: Remove print statement?

strax/context.py
```diff
@@ -1844,30 +1844,89 @@ def get_meta(self, run_id, target) -> dict:
...
+                )
+            return run_id, target, metafile
+
+        def _extract_metadata_and_lineage(self, run_id, target, metafile):
+            """Extract the actual metadata and lineage based on given inputs and whether the data is
+            available."""
+            # if the runid+target pair is given, check if stored
+            if metafile is None:
+                _is_stored = self.is_stored(run_id, target)
+                metadata = self.get_metadata(run_id, target) if _is_stored else None
```
WenzDaniel, line 1885: Here you could have used a 

```python
try: 
   self.get_metadata(...)
except strax,DataNotStored:
   ....
```
(or what ever raise we are returning in case it is not stored.) Just as information for the future

WenzDaniel (approved):
Thanks @KaraMelih works like a charm. Maybe you can add the doc-string a little and then we can merge.

WenzDaniel:
Btw you can ignore the failing code factor.

strax/context.py
```diff
@@ -1823,30 +1823,96 @@ def get_meta(self, run_id, target) -> dict:
...
+        compare,     or a dictionary of the metadata
+        :param return_results: bool, if True, returns a dictionary with metadata and lineages that
+            are found for the inputs does not do the comparison
+
+        example usage:
+            context.compare_metadata( ("053877", "peak_basics"),  "./my_path_to/JSONfile.json")
+            first_metadata = context.get_metadata(run_id, "events")
+            context.compare_metadata( ("053877", "peak_basics"), first_metadata)
+            context.compare_metadata( ("053877", "records"), ("053899", "records") )
+            results_dict = context.compare_metadata( ("053877", "peak_basics"), ("053877", "events_info"), return_results=True)
```
WenzDaniel, line 1839:
```suggestion
            context.compare_metadata( ("053877", "peak_basics"),  "./my_path_to/JSONfile.json")
            first_metadata = context.get_metadata(run_id, "events")
            context.compare_metadata(
                 ("053877", "peak_basics"), first_metadata)
            context.compare_metadata( 
                ("053877", "records"), ("053899", "records") )
            results_dict = context.compare_metadata( 
                ("053877", "peak_basics"), ("053877", "events_info"),
                 return_results=True)
```

WenzDaniel: Make git pre commit happy

WenzDaniel (approved):

PR #784, opened 2023-12-05
Avoid deprecated generated_jit

JelleAalbers:
Avoid a `NumbaDeprecationWarning` and a future crash from using the deprecated function `numba.generated_jit`.

`numba.generated_jit` is deprecated: https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-generated-jit. I think most people don't see the warning message for this yet because our environments use numba versions that are more than a year old.

We use `generated_jit` in `strax.endtime` to condition its behavior on the dtype of its input. This PR replaces `generated_jit` with `numba.extending.overload`. An additional benefit is that we don't need to condition on the environment variable `NUMBA_DISABLE_JIT` anymore, this was one of the main reasons `generated_jit` was deprecated.

### DIFF
--- a/strax/processing/general.py
+++ b/strax/processing/general.py
@@ -1,4 +1,3 @@
-import os
 import warnings
 
 warnings.simplefilter("always", UserWarning)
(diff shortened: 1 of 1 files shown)


dachengx:
Thanks @JelleAalbers . Just to check my understanding. Now outside numba decorated functions, like barely call `strax.endtime` will not trigger the usage of `_overload_endtime`. This is different from before but is exactly the functionality of `numba.extending.overload`. Am I correct? I test with some toy functions and it seems yes.

dachengx (approved):
I think we should be fine with the performance if we always put complex calculations into numba jit. Thanks again!

Issue #781, opened 2023-11-30
Copy to buffer breaks if dtype changes.


WenzDaniel:
**Describe the bug**
Not necessarily a bug, but more of a missing feature? If the dtype used in [copy_to_buffer](https://github.com/AxFoundation/strax/blob/21cc96e011b0e4099138979791f34e8b1addedb7/strax/dtypes.py#L266) is reduced by a field, but using the same function name it fails. This is problematic for example when adding removing the data_top field in straxen peaks. 

See screenshot below for an example.

<img width="216" alt="example" src="https://github.com/AxFoundation/strax/assets/43881800/dbd1688a-7267-44da-8d2c-769c1ff7fc54">

dachengx:
With MWE copied from @WenzDaniel linked image:

```
import numpy as np
import strax

a = np.ones (1, strax.peak_dtype(digitize_top=True))
b = np.zeros(1, strax.peak_dtype(digitize_top=True))
strax.copy_to_buffer(a, b, '_my_copy_function_name')
a = np.ones(1, strax.peak_dtype(digitize_top=False))
b = np.zeros(1, strax.peak_dtype(digitize_top=False))
strax.copy_to_buffer(a, b, '_my_copy_function_name')
```

There is no error anymore, thanks to https://github.com/AxFoundation/strax/pull/785.

PR #785, opened 2023-12-13
Add hot fix for copy_to_buffer

WenzDaniel:
In this PR we add a hotfix for the issue described in https://github.com/AxFoundation/strax/issues/781. Currently, we raise an error in `copy_to_buffer` if the dtypes of input and buffer changes, but the same function name is used. This also leads to failing tests in https://github.com/XENONnT/straxen/pull/1299 which is needed for https://github.com/XENONnT/straxen/pull/1300.

The PR is only a hotfix as it only adds a random string to the copy function name if the typing error is raised. However, this also means that the copy function is not cached after a dtype change unless the cache is cleared. 

For the future: 

A better approach would be to evaluate the cached function first and define the copy to buffer function name based on the dtype difference.

### DIFF
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -264,7 +264,7 @@ def copy_to_buffer(
         globals()[func_name](source, buffer)
     except numba.TypingError:
         rnd_string = _get_random_lower_case_string()
-        globals()[func_name + rnd_string](source, buffer)
+        copy_to_buffer(source, buffer, func_name + rnd_string, field_names)
 
 
 def _get_random_lower_case_string(n: int = 8):
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -273,6 +273,7 @@ def _create_copy_function(res_dtype, field_names, func_name):
     """Write out a numba-njitted function to copy data."""
     # Cannot cache = True since we are creating the function dynamically
     code_dtype_cache = f"{func_name + '_dtype_cache'} = {field_names}"
+    # pylint: disable=exec-use
     exec(code_dtype_cache, globals())
 
     code = f"""
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -258,7 +258,7 @@ def copy_to_buffer(
 
     # Add first two letters of all dtype fields to the function name as a
     # sort of hash-key.
-    field_names_hash = "".join([name[:2] for name in field_names])
+    field_names_hash = strax.deterministic_hash(repr(buffer.dtype))
     func_name += field_names_hash
     if func_name not in globals():
         # Create a numba function for us
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -256,26 +256,20 @@ def copy_to_buffer(
     if not func_name.startswith("_"):
         raise ValueError('Start function with "_"')
 
+    # Add first two letters of all dtype fields to the function name as a
+    # sort of hash-key.
+    field_names_hash = "".join([name[:2] for name in field_names])
+    func_name += field_names_hash
     if func_name not in globals():
         # Create a numba function for us
         _create_copy_function(buffer.dtype, field_names, func_name)
-    try:
-        globals()[func_name](source, buffer)
-    except numba.TypingError:
-        dtype_cache = globals()[func_name + "_dtype_cache"]
-        extra_fields = [name for name in dtype_cache if name not in field_names]
-        extra_fields = "_".join(extra_fields)
-        func_name += "_wo_" + extra_fields
-        copy_to_buffer(source, buffer, func_name, field_names)
+
+    globals()[func_name](source, buffer)
 
 
 def _create_copy_function(res_dtype, field_names, func_name):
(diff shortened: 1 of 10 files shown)


JoranAngevaare:
Caching a random string might induce a memory leak as each time the function is called, you cache a different string. Can you not just change the name of the cache-key when you call the function? `strax.copy_to_buffer(peaks, result, "bla")` -> `strax.copy_to_buffer(peaks, result, f"bla_{self.some_option")`

WenzDaniel:
Hej @JoranAngevaare nice to hear from you. Yes I have the same concern, and I also thought about adding to every plugin where we are calling this function an extra option, but this is a bit too cumbersome in my opinion. Right now I have no idea for a good solution. I do not know how we can get the dtype in the chached function easily. I am open for suggestions.

WenzDaniel:
Looks like that this hotfix does not work anyhow. For me locally it was working though.

JoranAngevaare:
Hi Daniel, likewise! Hope all is well. I just stumbled on this PR as I'm still subscribed to `strax`.

**Something simple**
The caching system is not very sophisticated, so you can just dump whatever string. I doubt that people are ever looking into the actual cached names.

Something blunt, that would also erase the need of the `func_name` is just adding the `repr` of the dtype:
```python
>>> a=np.ones([1,2], dtype=[('a', np.int64), ('b', np.int64)])
>>> repr(a.dtype)
"dtype([('a', '<i8'), ('b', '<i8')])"
```
You can just add that to the `func_name`, or even just make the `func_name` optional and just cache the `repr` of the `dtype`. You can make it more fancy buy throwing the `repr` in the `strax.deterministic_dtype`. 

**Something a bit more advanced**
Some other things worth considering would be making the cache explicit, e.g. following the `URLConfig` [`_CACHE`](https://github.com/XENONnT/straxen/blob/f23319dd37a23dddcc37156e62d112e36c8deafc/straxen/misc.py#L319C7-L319C16#L319C7-L319C16). I don't remember why I added it to `globals` at the time rather than some global dictionary.

Another, additional option might be using an LRU cache to make sure that the number of cached functions doesn't blow up, maybe something like [`straxen.CacheDict`](https://github.com/XENONnT/straxen/blob/f23319dd37a23dddcc37156e62d112e36c8deafc/straxen/misc.py#L319C7-L319C16#L319C7-L319C16), a minimal implementation is also used in [CMT1](https://github.com/XENONnT/straxen/blob/f23319dd37a23dddcc37156e62d112e36c8deafc/straxen/misc.py#L319C7-L319C16#L319C7-L319C16).

Both are probably overkill, I don't really believe that you can end up in a case where somehow the dtype is changed every time someone calls this function. But we have to keep that poor PhD in mind that will one day face the memory leak that my limited imagination had discarded 😉. 

The minimal amount of work is probably adding a `_COPY_CACHE` global dictionary instead of `globals()` and raising an error if more than a 1e6 keys are cached.

WenzDaniel:
> The caching system is not very sophisticated, so you can just dump whatever string. I doubt that people are ever looking into the actual cached names

Yes, I also decided now to make it very simple. I just cache in addition the dtype when calling the function. In this way the number of cached functions should be well less than 10.

WenzDaniel:
I do not get why it is failing here... locally it works...

JoranAngevaare:
It simply fails because coveralls is testing without numba, but why would you want to use a `try-except` block when you can just always add the fields to the cached function. Also much better performance wise?

WenzDaniel:
Haha thanks @JoranAngevaare for pointing it out. I also noticed now that it is a different raise condition. I thought just the same about making it simple now :D. You should come back to us you are missed :-)  I only use the first two letters of each field in the dtype to let the name not explode. 

Being ACs comes with very little time for these things.... this sucks a bit...

strax/dtypes.py
```diff
@@ -256,6 +256,10 @@ def copy_to_buffer(
     if not func_name.startswith("_"):
         raise ValueError('Start function with "_"')
 
+    # Add first two letters of all dtype fields to the function name as a
+    # sort of hash-key.
+    field_names_hash = "".join([name[:2] for name in field_names])
+    func_name += field_names_hash
```
JoranAngevaare, line 262:
```suggestion
    # Add first two letters of all dtype fields to the function name as a
    # sort of hash-key.
    field_names_hash = strax.deterministic_hash(repr(buffer.dtype))
    func_name += field_names_hash
```

JoranAngevaare (dismissed):
Hi Daniel, I'm sure there is still loads of fun to be had in XENON 😉.

Changes look good! Seems simple and effective. Only one suggestion that would also cache the `dtype` of the field might be something as below, the `field_names_hash` will be anyway unreadable so you might consider including the `np.int64` `np.int16` etc. as well.

Additionally, you can now make the `func_name` arguement optional, I don't really think there is a use case where you'd use the same buffer fields but for some reason need a new `func_name`.

Good to have the test in place 👍 !

WenzDaniel:
I merge since Joran basically approved.

PR #789, opened 2024-01-14
`np.float` is deprecated

dachengx:

> Using the aliases of builtin types like np.int is deprecated

reference: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

In principle the auto tests in https://github.com/XENONnT/base_environment/pull/1483 will not pass, but the line changed in this PR was never checked:

image

  - Update the docstring(s)
  - Update the documentation


WenzDaniel:
Thanks Dacheng

PR #794, opened 2024-01-17
Remove NotImplementedError for multi_output in DownChunkingPlugin

HenningSE:
Right now it is not possible to have a DownChunkingPlugin with multiple outputs. I would like to add truth information handling in fuse and for that I need a second output in the PMTResponseAndDAQ plugin. 

The PR just removes the `NotImplementedError`.

It was necessary to change how the PMTResponseAndDAQ yields its results a bit but I found no problems running the fuse simulations with the implemented changes. The corresponding fuse PR can be found here: https://github.com/XENONnT/fuse/pull/99.

### DIFF
--- a/strax/plugins/down_chunking_plugin.py
+++ b/strax/plugins/down_chunking_plugin.py
@@ -26,13 +26,6 @@ class DownChunkingPlugin(Plugin):
                 "currently does not support parallel processing."
             )
 
-        if self.multi_output:
-            raise NotImplementedError(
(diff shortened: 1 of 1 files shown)


dachengx:
There is some discussion about this. As a reference: https://github.com/AxFoundation/strax/pull/769#discussion_r1386797972

dachengx:
Hey @HenningSE . Is this still needed?

HenningSE:
If we go with the Truth plugins as they are right now in the PR: no.

dachengx:
We are still able to reopen this later.

PR #796, opened 2024-01-26
Warn if user checks is_stored for plugin not always saved

cfuselli:

This PR addresses the issue of users potentially being unaware when querying for stored data of a plugin in strax that is configured with save_when < ALWAYS (3).  In such cases, even if the is_stored function returns False, it might not reflect a lack of data due to processing or storage issues but rather due to the plugin's configuration. This update aims to improve user awareness by providing a warning in such scenarios.

Tested quickly with some single-output and multi-output plugins and seems to work.


strax/context.py
```diff
@@ -1988,6 +1988,22 @@ def is_stored(self, run_id, target, **kwargs):
...
+        # Before returning False, check if the data can be made trivially
+        plugin = self._plugin_class_registry[target]
+        save_when = plugin.save_when
+
+        # Mutli-target plugins provide a save_when per target
+        if isinstance(save_when, immutabledict):
+            save_when = save_when[target]
+
+        if save_when < 3:  # 3 is SaveWhen.ALWAYS
+            warnings.warn(
```
WenzDaniel, line 2001: Hej @cfuselli you could add within this if-statement something like:
```python
try:
    components = st.get_components('058518', target)
    targets_plugin_made_from = components.loaders.keys()
    # Warn user:
    print('Can be made from: ', targets_plugin_made_from)
except strax.DataNotAvailable:
    # Warn that data cannot be made
    print('cannot be made')
```

cfuselli: Nice but I think it goes way beyond the scope of is_stored, no? 
The warning is just intended to inform user of the existence of plugins that are not stored, so that they won't use these plugins to check if something is stored.

strax/context.py
```diff
@@ -1988,6 +1988,22 @@ def is_stored(self, run_id, target, **kwargs):
...
+
+        # Before returning False, check if the data can be made trivially
+        plugin = self._plugin_class_registry[target]
+        save_when = plugin.save_when
+
+        # Mutli-target plugins provide a save_when per target
+        if isinstance(save_when, immutabledict):
+            save_when = save_when[target]
+
+        if save_when < 3:  # 3 is SaveWhen.ALWAYS
```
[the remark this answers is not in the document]
cfuselli: 👍

strax/context.py
dachengx: Hey @WenzDaniel. Shal I still mark this as an unsolved thread?

WenzDaniel: For me it is fine to move ahead. I take what I can get.

yuema137 (approved):
I also tested with a few plugins and it looks fine to me. Regarding Daniel's comment, I'm thinking that we could probably add a `is_makable` function to tell the users the nearest stored parents that the plugin could be made of.

yuema137:
@dachengx should we merge this PR soon? I think it's a good warning message for the users.

PR #798, opened 2024-01-31
Update tags handling, added comment field. Allows to define superuns …

WenzDaniel:
Tiny update which adds comments to superrun rundocs, and allows to make superruns of superruns.


yuema137 (approved):
Looks good to me. Thanks Daniel

PR #803, opened 2024-02-07
Simpler chunk length check, avoid recursion limit crash

JelleAalbers:

This fixes a crash @cfuselli always sees after restrax for some time.

In #758 we made StorageBackend check that the chunk length matches the length specified in the metadata. We wrap the class attribute _read_chunk in `__new__`. But `__new__` runs every time we make a new instance. Applications that make backends many times (like a running restrax session) end up with many layers of wrappings, and eventually hit python's recursion limit.

This keeps the check (and its unit test) but implements it more simply. Since `_read_chunk` is only used in `StorageBackend._read_and_format_chunk`, we can just add the check there. It will still apply to every storage backend.

@dachengx please let me know if I missed a specific reason behind the implementation in #758!

### DIFF
--- a/strax/storage/common.py
+++ b/strax/storage/common.py
@@ -429,13 +429,6 @@ class StorageBackend:
 
     """
 
-    def __new__(cls, *args, **kwargs):
-        """Mandatorily wrap _read_chunk in a check_chunk_n decorator."""
-        if "_read_chunk" in cls.__dict__:
-            method = getattr(cls, "_read_chunk")
-            setattr(cls, "_read_chunk", strax.check_chunk_n(method))
-        return super(StorageBackend, cls).__new__(cls)
-
     def loader(self, backend_key, time_range=None, chunk_number=None, executor=None):
         """Iterates over strax data in backend_key.
 
@@ -531,6 +524,12 @@ class StorageBackend:
         else:
             data = self._read_chunk(
                 backend_key, chunk_info=chunk_info, dtype=dtype, compressor=metadata["compressor"]
+            )
+
+        if len(data) != chunk_info["n"]:
+            raise strax.DataCorrupted(
+                f"Chunk {chunk_info['filename']} of {chunk_info['run_id']} has {len(data)} items, "
+                f"but chunk_info {chunk_info} says {chunk_info['n']}"
             )
 
         _is_superrun = chunk_info["run_id"].startswith("_")
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -1,5 +1,4 @@
 import typing as ty
-from functools import wraps
 
 import numpy as np
 import numba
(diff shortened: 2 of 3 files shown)


strax/storage/common.py
```diff
@@ -429,13 +429,6 @@ class StorageBackend:
 
     """
 
-    def __new__(cls, *args, **kwargs):
-        """Mandatorily wrap _read_chunk in a check_chunk_n decorator."""
-        if "_read_chunk" in cls.__dict__:
-            method = getattr(cls, "_read_chunk")
```
dachengx, line 435: Thanks @JelleAalbers for the improvement. I understand the request from Carlo and your new implementation. I implemented #758 to force-check the chunk size. Can we do something like `hasattr(cls._read_chunk, '__closure__') and cls._read_chunk.__closure__ is not None` to test whether `_read_chunk` is decorated?

Or use some method like https://stackoverflow.com/a/68583930.

JelleAalbers: Sure, I understand the check is important, and the storage class hierarchy is sufficiently confusing that I understand you reached for decorators. And there will be a way to fix the decorator solution with more metaprogramming (at the risk of making more hard-to-spot bugs).

I think `_read_and_format_chunk` is the better place to do this check. The `_read_chunk` method returns a numpy array, not a strax Chunk: it's `_read_and_format_chunk`'s job to turn that into a chunk. Creating a chunk triggers checks, like whether the data matches the time range, the start and end time are integers, etc. So it makes sense to check the chunk length matches the metadata at the same time.

There is no code that calls  `_read_chunk` directly -- at least  not in the AxFoundation or XENONnT GitHub. The only reason I'd write something like that is to _avoid_ all the checks that come with making a strax chunk, like when trying to read data whose metadata is corrupted. Then you don't want the chunk length check either.

dachengx:
@FaroutYLq did you check the chunk size manually but not using `_read_chunk`?

FaroutYLq:
> @FaroutYLq did you check the chunk size manually in outsource but not using `_read_chunk`?

No I [was using](https://github.com/XENONnT/outsource/blob/a220e2cc004ff2d66ff61c4306240cb308ac2e6e/outsource/workflow/runstrax.py#L295) `strax.load_file(files[i], compressor=compressor, dtype=dtype)` for each existing file. Is it more recommended to use `_read_chunk` now?

dachengx:
> No I [was using](https://github.com/XENONnT/outsource/blob/a220e2cc004ff2d66ff61c4306240cb308ac2e6e/outsource/workflow/runstrax.py#L295) `strax.load_file(files[i], compressor=compressor, dtype=dtype)` for each existing file. Is it more recommended to use `_read_chunk` now?

No not really.

dachengx (approved):

PR #816, opened 2024-04-09
Fix 'division by zero' bug when updating the progress bar for a very small chunk

zaitsev136:
**What is the problem?**

When developing plugins, some fake test data can be used, with a very small size of the resulting data chunk. When strax loads a very small chunk, it calls time.time() twice to measure how much time it took to load (or process). On some systems, time.time() can return exactly the same values if called twice within a small time interval. This causes a "division by zero" error in the _update_progress_bar function.

**How is it fixed?**

time.perf_counter() is used instead of time.time()


dachengx:
If you have multiple chunks in a run then the `tqdm` will update its `last_print_t` by `time.time` so that in that case you will see negative time because `pbar` used both `time.time` and `time.perf_counter`.

dachengx (changes requested):
If you have multiple chunks in a run then the `tqdm` will update its `last_print_t` by `time.time` so that in that case you will see negative time because `pbar` used both `time.time` and `time.perf_counter`.
image

zaitsev136:
Here is the minimal code example needed to reproduce the error:

```
import numpy as np
import strax

class SimplestPlugin(strax.Plugin):

    provides = 'simple_data'
    depends_on = tuple()
    __version__ = None
    dtype = strax.time_dt_fields + [
                (('Simple data field', 'data'), np.int16),
            ]

    def compute(self, chunk_i):
        start = 0
        dt = 1000
        length = 1
        end = start + dt*length
        result = np.empty(1, dtype=self.dtype)
        result['time'] = start
        result['length'] = length
        result['dt'] = dt
        result['data'] = chunk_i

        return self.chunk(start=start, end=end, data=result)

    def is_ready(self, chunk_i):
        return chunk_i < 1

    def source_finished(self):
        return True


if __name__ == '__main__':

    context = strax.Context(storage=[strax.DataDirectory('test_helix_data'), ],
                            register=[SimplestPlugin, ])
    data = context.get_array('run0', 'simple_data')
    exit()
```

Below is the error that I get in ~50% of the cases when I run the above code.


```
Source finished!
Could not estimate run start and end time from run metadata: assuming it is 0 and inf
Target Mailbox (simple_data) killed, exception <class 'ZeroDivisionError'>, message float division by zero
Exception in thread build:simple_data:
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.3056.0_x64__qbz5n2kfra8p0\lib\threading.py", line 1016, in _bootstrap_inner
    self.run()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.3056.0_x64__qbz5n2kfra8p0\lib\threading.py", line 953, in run
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\mailbox.py", line 290, in _send_from
    self.close()
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\mailbox.py", line 354, in close
    self.send(StopIteration)
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\mailbox.py", line 307, in send
    raise MailboxKilled(self.killed_because)
strax.mailbox.MailboxKilled: (<class 'ZeroDivisionError'>, ZeroDivisionError('float division by zero'), <traceback object at 0x00000295304E89C0>)
  File "C:\Users\zaits\PycharmProjects\helix\helix\test.py", line 109, in <module>
    data = context.get_array('run0', 'simple_data')
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\context.py", line 1606, in get_array
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\context.py", line 1606, in <listcomp>
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\context.py", line 1472, in get_iter
    generator.throw(e)
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\processor.py", line 312, in iter
    raise exc.with_traceback(traceback)
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\processor.py", line 265, in iter
    yield from final_generator
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\mailbox.py", line 438, in _read
    self.kill_from_exception(e)
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\mailbox.py", line 203, in kill_from_exception
    raise e
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\mailbox.py", line 435, in _read
    yield res
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\context.py", line 1457, in get_iter
    self._update_progress_bar(
  File "C:\Users\zaits\PycharmProjects\helix\venv\lib\site-packages\strax\context.py", line 1531, in _update_progress_bar
    pbar.mbs.append((nbytes / 1e6) / seconds_per_chunk)
ZeroDivisionError: float division by zero

Process finished with exit code 1
```


It is not 100% guaranteed that you can reproduce this error on your machine, because the behavior of time.time() function is system-dependent. Here is what [time module documentation](https://docs.python.org/3/library/time.html#time.time) says about it:

`Note that even though the time is always returned as a floating point number, not all systems provide time with a better precision than 1 second. `

Which in principle means that if processing of a chunk takes less than a second, it is possible that on some systems `time.time() - pbar.last_print_t` in context.py/_update_progress_bar will be zero.

zaitsev136:
oh, I see, I didn't realize that `last_print_t` is automatically updated with time.time()

In that case I suggest to continue using time.time(), but to add a check before dividing by `seconds_per_chunk` in `context.py/_update_progress_bar()`

E.g.,
```
if seconds_per_chunk == 0:
   seconds_per_chunk = 1e-6 # using 1 microsecond when time.time() yields the exact same value twice
```

The estimated rate would be wrong, of course, but I think this is such a niche problem, it really does not matter too much what the processing rate is in this case.

Would that be ok?

zaitsev136:
Or maybe a more involved solution:

```
        seconds_per_chunk = time.time() - pbar.last_print_t
        if seconds_per_chunk == 0:
            pbar.mbs.append(np.inf)
        else:
            pbar.mbs.append((nbytes / 1e6) / seconds_per_chunk)
        mbs = np.mean(np.array(pbar.mbs)[np.isfinite(pbar.mbs)]) if np.any(np.isfinite(pbar.mbs)) else np.inf
        if np.isinf(mbs):
            rate = ''
        elif mbs < 1:
            rate = f"{mbs*1000:.1f} kB/s"
        else:
            rate = f"{mbs:.1f} MB/s"
        postfix = f"#{n_chunks} ({seconds_per_chunk:.2f} s). {rate}"
```

Here, if seconds_per_chunk is 0, np.inf is appended to pbar.mbs.
When the mean rate is calculated, inf values are ignored. If mbs only contains inf values, the rate information is not posted in the postfix. Users would get an output like this, without the rate:
`Loading record_basics: |██████████| 100.00 % [00:00<00:00], #1 (0.00 s). `

dachengx:
@zaitsev136  How about this: https://github.com/AxFoundation/strax/pull/819?

dachengx:
So we close this? @zaitsev136

PR #819, opened 2024-04-09
Increase the timing precision of progress bar

dachengx:

To solve https://github.com/AxFoundation/strax/pull/816

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1447,7 +1447,7 @@ class Context:
                 run_id, targets=targets, progress_bar=progress_bar
             )
             with _p as pbar:
-                pbar.last_print_t = time.time()
+                pbar.strax_print_time = time.perf_counter()
                 pbar.mbs = []
                 for n_chunks, result in enumerate(strax.continuity_check(generator), 1):
                     seen_a_chunk = True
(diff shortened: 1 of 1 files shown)


yuema137 (approved):
@dachengx Thanks for this PR and thanks for catching this issue @zaitsev136
Looking at #816 and this PR, I agree that it's a better solution to use an independent attribute.  The `last_print_t` is an attribute provided by `tqdm` which is updated with `time.time` in `__iter__`. So when we do 
https://github.com/AxFoundation/strax/blob/a1e921fce2519db5e58ac710487a6690475f0438/strax/context.py#L1473 or
https://github.com/AxFoundation/strax/blob/a1e921fce2519db5e58ac710487a6690475f0438/strax/context.py#L1552
The `pbar.last_print_t` will be overwritten by `time.time`. As we don't want to touch the inner side of `tqdm` it's good to separate the high-resolution time calculation with a different one.

dachengx:
Let's wait for @zaitsev136 's review.

zaitsev136:
Looking good, thanks a lot for looking into this

dachengx:
> Looking good, thanks a lot for looking into this

Will this PR solve the error you saw?

zaitsev136:
yes, it will

PR #821, opened 2024-04-11
Update context.py

WenzDaniel:
Propagates multi_run doc string to get_XXX functions such that people are aware of that feature.


dachengx (approved):
Thanks @WenzDaniel

PR #838, opened 2024-05-02
Implement Hyperrun

dachengx:

Hyperruns. Hyperruns are superruns, also processing of hyperruns depends on `run_metadata` defined similarly to https://straxen.readthedocs.io/en/latest/tutorials/SuperrunsExample.html#Define-a-superrun:.

If a hyperrun has run_id `__000000`, the `plugin.run_id` of the plugins used in processing will be `000000`, and the subruns of the hyperrun will be mixed(concatenated) while processing.

A new attribute `allow_hyperrun` of `strax.Plugin` class is added. The `allow_hyperrun` of `data_type` depends on the `data_type` whose `allow_hyperrun` is `True` can only be `True`. So this means, your father is `True`, so you must be `True`.

The key feature of hyperrun is that the chunks of subruns can be loaded together.

For superruns, the subruns are still made and loaded separately. For hyperrun, we can really treat the combination of subruns as a single run in the context of processing.

For example, in the added test., the data_type `sum` deepens on `ranges`. The `"data"` in `ranges` are just sequence of numbers in order with an offset, like `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]` and `[10, 11, 12, 13, 14, 15, 16, 17, 18, 19]`. The `sum` will sum the `"data"` in `ranges` by `ranges["data"] + ranges["data"][::-1]`, so you will get `[0] * 10` or `[29] * 10`.

Frist, note that we will have 3 runs. And the `"data"` in the `ranges` of 3 subruns are assigned to be `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`, `[10, 11, 12, 13, 14, 15, 16, 17, 18, 19]`, and `[20, 21, 22, 23, 24, 25, 26, 27, 28, 29]`

1. If the loaded `sum` is from a superrun, the `sum` will be `[9] * 10 + [29] * 10 + [49] * 10`. Just because the subruns will still be calculated separately.
2. If the loaded `sum` is from a hyperrun, the `sum` will be `[29] * 30`. Because the subruns will still be first combined into `"data"` whose length is 30, then `ranges["data"] + ranges["data"][::-1]` run. So the subruns are really loaded as if they are the same run.

The example is in the added test.

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -245,7 +245,7 @@ class Chunk:
         )
 
     @classmethod
-    def concatenate(cls, chunks):
+    def concatenate(cls, chunks, allow_hyperrun=False):
         """Create chunk by concatenating chunks of same data type You can pass None's, they will be
         ignored."""
         chunks = [c for c in chunks if c is not None]
@@ -261,7 +261,7 @@ class Chunk:
 
         run_ids = [c.run_id for c in chunks]
 
-        if len(set(run_ids)) != 1:
+        if len(set(run_ids)) != 1 and not allow_hyperrun:
             raise ValueError(
                 f"Cannot concatenate {data_type} chunks with different run ids: {run_ids}"
             )
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -277,6 +277,12 @@ class Chunk:
                 )
             prev_end = c.end
 
+        # try:
+        #     np.concatenate([c.data for c in chunks])
+        # except:
+        #     import pdb
+        #     pdb.set_trace()
+
         return cls(
             start=chunks[0].start,
             end=chunks[-1].end,
--- a/strax/context.py
+++ b/strax/context.py
@@ -2432,6 +2432,48 @@ class Context:
             for data_type, _hash, save_when, version in hashes
         }
 
+    def check_hyperrun(self):
+        # root data_type
+        root_data_types = set()
+        for k, v in self._plugin_class_registry.items():
+            if not v.depends_on:
+                root_data_types |= set((k,))
+
+        # inverse tree whose key is depends_on and value is provides
+        inverse_tree = dict()
+        for k, v in self._plugin_class_registry.items():
+            for d in v().depends_on:
+                inverse_tree.setdefault(d, [])
+                inverse_tree[d] += v().provides
+
+        # define a recursive function to check if all the dependencies support hyperruns
+        def check_support_hyperrun(data_type, checked=set(), seen_allow=None):
+            if data_type in checked:
+                return checked
+            if self._plugin_class_registry[data_type].allow_hyperrun:
+                seen_allow = data_type
+            if seen_allow and not self._plugin_class_registry[data_type].allow_hyperrun:
+                raise ValueError(
+                    f"Already support hyperruns in {seen_allow}, "
+                    f"but it's dependency {data_type} does not support hyperruns."
+                )
+            checked |= set((data_type,))
+            if data_type not in inverse_tree:
+                return checked
+            for d in inverse_tree[data_type]:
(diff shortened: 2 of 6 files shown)


strax/chunk.py
```diff
@@ -277,6 +277,12 @@ def concatenate(cls, chunks):
                 )
             prev_end = c.end
 
+        # try:
```
yuema137, line 280: Should we remove the debug comments?

strax/context.py
```diff
@@ -731,7 +740,7 @@ def __get_requested_plugins_from_cache(
                 continue
 
             requested_p = plugin.__copy__()
-            requested_p.run_id = run_id
+            requested_p.run_id = self.hyperrun_id(run_id)
```
yuema137, line 743: Isn't it just `_run_id`? Probably better to be consistent but maybe I'm too nitpicking

yuema137:
Hi @dachengx this PR looks good to me, but I'm a little bit confused:
- Why do we add hyperrun on top of superrun? It seems to me that their duties are kind of duplicated and the restrictions & checkings for hyperrun could be set for superrun as well. 
It will be great if you could provide a scenario when we need this feature

strax/context.py
dachengx: What do you mean by `_run_id`?
`self.hyperrun_id` is used several times so I write a function.

strax/chunk.py
dachengx: Yes, sorry I forgot it. Removed at https://github.com/AxFoundation/strax/pull/838/commits/467205547c18fa235be7d54dae50748b31247349

strax/context.py
yuema137: I meant that `_run_id = self.hyperrun_id(run_id)` was already calculated previously, so it doesn't harm to just use it?

dachengx: GitHub website can wrap the unchanged block of codes.

dachengx: The already calculated `_run_id` is from `def _set_plugin_config`. But the `requested_p.run_id = self.hyperrun_id(run_id)` is in `def 
 __get_requested_plugins_from_cache`.

dachengx:
@yuema137 the test is added

yuema137 (approved):
Hi @dachengx, I read the changes carefully, and now I think I understand this much better:
- In `get_components`, the treatment for hyperrun is actually similar to normal runs, which means there is only a single loader. For superruns a loader is defined for each subrun.
- And the difference is that for hyperruns more checks are done to guarantee that the targets to load satisfy the requirements for hyperruns.
- Therefore the hyperrun is precisely equivalent to a single run in the view of strax/straxen. So when you use the exhaust plugins, the processor will regard several runs as a single one and give you the whole chunk. But for normal plugins, there should not be a difference.

The test works fine for me. Please merge it if my understanding is correct. Otherwise please let me know.

zihaoxu98:
Thanks for this hyper interesting PR. Although I cannot follow the code, based on my understanding in the test case, we always need `ExhaustPlugin` when we need to concatenate the chunks and do the computing right? I am asking because I found the document of `ExhaustPlugin` is lost and it would give the wrong result if I changed it to the normal `Plugin`.

yuema137:
@zihaoxu98 The purpose of Hyperrun is to let strax regard the runlist as a single run (for super run it's not the case as strax still treats them as separate runs). It doesn't make observable changes for normal plugins because the data is still chunked, and the computation is done chunk by chunk. 
Only with ExhaustPlugin, which tries to combine whatever chunk it can find within a single run, is the computation performed for the combined big chunk as we expect. So basically, `ExhaustPlugin` merges all possible chunks in a single run, and Hyperrun breaks the boundary between runs.

PR #840, opened 2024-05-15
Update register_all

cfuselli:
If you do register_all on a module that contains a plugin that inherits with another, the effect of registering the plugin will be cancelled. 

Example: 

Module1: 
```
class ParentPlugin(strax.Plugin):
    provides = "some_type"
    ...
    ...
```

Module 2:
```
class MyPlugin(ParentPlugin):
    ...
    ...

```

Running register_all on Module2 will register: 

- ParentPlugin
- MyPlugin
- ParentPlugin

Adding this check, this will only register MyPlugin.


cfuselli:
I see that some tests fail. Maybe I am missing something..

JoranAngevaare:
Seems like this "feature" was utilized in straxen in the past, so probably in other places too. Actually, in straxen we were particularly lazy as we register_all `straxen.plugins`: https://github.com/XENONnT/straxen/blob/fe9b9fae529862c6dc1e3e273f441d662bc7e355/straxen/contexts.py#L95. This means that probably close to none of the plugins will be registered after this change. You would have to change this to `register_all([straxen.plugins.afterpulses.afterpulse_processing, straxen.plugins.aqmon_hits.aqmon_hits, ...])` etc.

Maybe instead for now you could add a warning that these additional plugins will not be registered anymore in a feature release unless explicitly imported? Or you could add a "strict/sloppy" argument for register all if yo want to be able to turn on/off this behavior. Otherwise things start breaking all at once (as you see with these tests now).

WenzDaniel:
Given the discussion above and the absence of any activity on this I will close the PR for now.

PR #844, opened 2024-06-09
Only replace `"_temp"` in folder name by `""` if saved successfully

dachengx:

Before this PR, when an error happens when computing a data_type, the folder of that data_type will be renamed from `0-data_type-lineage_temp` to `0-data_type-lineage`. This will confuse users about whether the data_type is successfully made if users can not access their error log.

In this PR, only if all data are successfully saved, the rename will be performed.

Add a keyword argument `successful` to `_close` method of backends. So later when not successfully exits, the `successful` will be False.

  - Update the docstring(s)
  - Update the documentation


dachengx:
I can not solve the errors easily. So I close this.

PR #856, opened 2024-07-12
Allow `_chunk_number` to be list or tuple

dachengx:

This is helpful when you just want to process several chunks for testing.

`_chunk_number` will be propagated from `get_iter`: https://github.com/AxFoundation/strax/blob/bb23240c199276656fe9bd1dda7e9f5cf90bb127/strax/context.py#L1481

  - Update the docstring(s)
  - Update the documentation


yuema137 (approved):
It's straightforward and looks good. Thanks.

Issue #829, opened 2024-04-26
No safe guard for mimicking plugin


dachengx:
If plugins A and B have the same `__version__`, `depends_on`, `provides`, and `self.__class__.__name__`, they will have the same lineage. We currently have to method to prevent this.

yuema137:
See #857

PR #857, opened 2024-07-16
Add Source Hash to Lineage

yuema137:
To solve the problem mentioned in #829 , we add the hash for the source code to the plugin lineage. A unit test is also added as an minimal working example.
We need to check really carefully if this change of lineage format impact any functions in strax/straxen. So I label it as draft for now.


yuema137:
The gain is limited, considering that we don't have adequate manpower to maintain. This idea should be kept in mind and may be helpful in the future, though.

Issue #846, opened 2024-06-16
Zero length array returned when targets duplicated


dachengx:
**Describe the bug**

When there are multiple targets and some of them duplicate, an error will show.

```
ValueError: Cannot merge chunks with different number of items: [[012882.event_info: 0sec 124900000 ns - 4sec 876677420 ns, 20 items, 0.0 MB/s], [012882.event_basics: 4sec 876677420 ns - 4sec 876677420 ns, 0 items, -1.0 MB/s], [012882.event_basics: 4sec 876677420 ns - 4sec 876677420 ns, 0 items, -1.0 MB/s]]
```

**To Reproduce**

```python
import straxen
from straxen.test_utils import nt_test_run_id


run_id = nt_test_run_id

st = straxen.test_utils.nt_test_context()
st.make(nt_test_run_id, 'event_basics')
st.make(nt_test_run_id, 'event_info')

data = st.get_df(
    run_id,
    ('event_info', 'event_basics', 'event_basics'),
)

print(len(data))
```

**Expected behavior**

**Screenshots**

**Versions**
Please add the version of strax and any related package

yuema137:
see #859

PR #859, opened 2024-07-26
Check duplicate `depends_on` of `Plugin`

yuema137:
### Explanation
Basically what happened was:
- In `plugin.py`, the chunks are fetched for each `self.depends_on` and stored in a dictionary: https://github.com/AxFoundation/strax/blob/2218e78722cb64cdedb17e255752b7857c5f229b/strax/plugins/plugin.py#L458-L493
- Therefore, when there are duplicate items in `self.depends_on` , the later one overwrites the previous one
- The later one cannot get anything, because the fetch is done in this way: https://github.com/AxFoundation/strax/blob/2218e78722cb64cdedb17e255752b7857c5f229b/strax/plugins/plugin.py#L467-L468 As the data is already read and split by the previous fetch, the later one can only get an empty result
- So finally both ones are empty

### Solution
The fix is simple, which is to remove the duplicate items in `self.depends_on` of the `Plugin` class.


strax/plugins/plugin.py
```diff
@@ -110,6 +110,8 @@ def __init__(self):
             raise ValueError(f"depends_on not provided for {self.__class__.__name__}")
 
         self.depends_on = strax.to_str_tuple(self.depends_on)
+        # Remove duplicates
+        self.depends_on = tuple(dict.fromkeys(self.depends_on))
```
dachengx, line 114: Better use `tuple(set(self.depends_on))`.

dachengx (changes requested):

strax/plugins/plugin.py
yuema137: I believe it's the same functionally, but yes, let's make it easier.

yuema137:
@dachengx In order to pass the pre-commit check I had to format it with black & flake. I think it happens now and then when we make PR for the files that were not checked previously. Should we do a through cleaning for strax?

strax/plugins/plugin.py
```diff
@@ -607,7 +620,8 @@ def do_compute(self, chunk_i=None, **kwargs):
                 # This block will be deleted </end>
             elif len(set(tranges.values())) != 1:
                 message = (
-                    f"{self.__class__.__name__} got inconsistent time ranges of inputs: {tranges}"
+                    f"{self.__class__.__name__} "
```
dachengx, line 623: Why this change is needed?

strax/plugins/plugin.py
```diff
@@ -459,14 +461,25 @@ class IterDone(Exception):
                     # Fetch other inputs (when needed)
                     for d in self.depends_on:
                         if d != pacemaker:
+                            print(
```
dachengx, line 464: Please remove these prompt.

strax/plugins/plugin.py
```diff
@@ -110,6 +110,8 @@ def __init__(self):
             raise ValueError(f"depends_on not provided for {self.__class__.__name__}")
 
         self.depends_on = strax.to_str_tuple(self.depends_on)
+        # Remove duplicates
+        self.depends_on = tuple(self.depends_on)
```
dachengx, line 114: `self.depends_on = tuple(set(self.depends_on))`

dachengx (changes requested):

strax/plugins/plugin.py
dachengx: But will this really kill the issue? https://github.com/AxFoundation/strax/issues/846

strax/plugins/plugin.py
yuema137: the line is to long and black complains

dachengx (approved):

PR #863, opened 2024-08-04
Include `chunk_number` in lineage: Per chunk storage

dachengx:

The `chunk_number` was passed to `Context.get_iter` to only load a specific chunk. But the previous implementation has weaknesses:

1. The `chunk_number` is not tracked by lineage so if you process twice with a different `chunk_number`, though the lineage is the same, the results are not.
2. Can not assign which `data_type` to load by chunk number.

This PR makes sure that the `chunk_number` is tracked by lineage, and you can assign which chunk to load which data_type.

This PR is designed following plan: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:analysis:analysis_tools_team:sr2_processing. We will later make [outsource](https://github.com/XENONnT/outsource) more compatible with strax via this PR.

Depends on https://github.com/AxFoundation/strax/pull/856

Functionality not implemented:

1. The `chunk_number` will not be passed to run selection, because we will not send the metadata including `chunk_number` to DB.
2. Change `Context.__add_lineage_to_plugin` to add `chunk_number` as a configuration of `Plugin`.
3. `chunk_number` is set to be `chunk_number: ty.Optional[ty.Dict[str, ty.List[int]]] = None` as argument in functions, e.g. it can be `{"raw_records": [0, 1]}` or `{"peaklets": [1], "lone_hits": [0]}`. The latter example is means this PR adds the functionality.
4. Add a function `Context.merge_per_chunk_storage` to combine the per chunk storage into normal storage(where `chunk_number` is not a configuration of `Plugin`).

Vulnerability of this PR:

1. When the chunking of dependencies changes, the result will also change. But this change can not be reflected in lineage.

```
st.make("0", "peaklets", chunk_number={"raw_records": [0]})
st.make("0", "peak_basics", chunk_number={"peaklets": [1], "lone_hits": [0]})
```

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -906,6 +906,22 @@ class Context:
 
         return plugin
 
+    @staticmethod
+    def _check_chunk_number(chunk_number: ty.List[int]):
+        """Check if the chunk_number is a list of consecutive integers."""
+        mask = isinstance(chunk_number, list)
+        mask &= all([isinstance(x, int) for x in chunk_number])
+        if not mask:
+            raise ValueError(f"chunk_number should be a list of integers, but got {chunk_number}")
+
+        # Check if the difference between adjacent elements is exactly one
+        for i in range(len(chunk_number) - 1):
+            if chunk_number[i + 1] - chunk_number[i] != 1:
+                raise ValueError(
+                    "chunk_number should be a list of consecutive integers, "
+                    f"but got {chunk_number}"
+                )
+
     def __add_lineage_to_plugin(
         self,
         run_id,
@@ -962,8 +978,7 @@ class Context:
                         raise ValueError(
                             f"Chunk number for {d_depends} is already set in the lineage"
                         )
-                    if not isinstance(chunk_number[d_depends], list):
-                        raise ValueError("chunk_number should be a list of integers")
+                    self._check_chunk_number(chunk_number[d_depends])
                     configs["chunk_number"][d_depends] = chunk_number[d_depends]
 
         plugin.lineage = {
--- a/strax/context.py
+++ b/strax/context.py
@@ -1,20 +1,21 @@
-import datetime
+import time
 import logging
 import warnings
 import fnmatch
-from functools import partial
-import typing as ty
-import time
+import itertools
 import json
+from functools import partial
+import types
+import typing as ty
+from enum import IntEnum
+import datetime
+import inspect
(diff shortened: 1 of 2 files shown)


dachengx:
@FaroutYLq no hurry, I need to test it further. Maybe you are able to hire more people to review.

dachengx:
Maybe we can add the hash (maybe sha256) of metadata of `raw_records` to lineage to fix the venerability of this PR.

FaroutYLq (commented):
Hi thanks for the efforts. I need a bit more time to digest but please let me start asking dumb questions:
- Do we have any plugin that is supposed have no chunk at all (only with metadata) when computation is finished? I vaguely remember seeing something like this before, but not sure if it is just because that it failed somewhere. I don't have an example on hand unfortunately.
  - If so will this PR breaks things?
- In this PR, is it expected to have different hash for different chunks, in the same plugin?

I am happy to review this and hope to learn more about the core, but I want to figure out these questions first before diving into it.

dachengx:
> Hi thanks for the efforts. I need a bit more time to digest but please let me start asking dumb questions:
> 
> * Do we have any plugin that is supposed have no chunk at all (only with metadata) when computation is finished? I vaguely remember seeing something like this before, but not sure if it is just because that it failed somewhere. I don't have an example on hand unfortunately.
>   
>   * If so will this PR breaks things?
> * In this PR, is it expected to have different hash for different chunks, in the same plugin?
> 
> I am happy to review this and hope to learn more about the core, but I want to figure out these questions first before diving into it.

-  `chunk_number` tell us which chunk to load but not which chunk to save. So if a data_type has zero chunk, there will be an error shows.
- Yes.

WenzDaniel:
Is this feature really needed? I have the feeling it will cause us more trouble than we gain. Can you give a few use cases why this feature is required?

dachengx:
> Is this feature really needed? I have the feeling it will cause us more trouble than we gain. Can you give a few use cases why this feature is required?

It is needed in reprocessing when we want to process a run but do not want to wait for all chunks to be processed in sequence. A homemade and similar feature is already in [outsource](https://github.com/XENONnT/outsource) but according to @FaroutYLq , it is not so compatible with strax. So this feature is needed.

PR #866, opened 2024-08-11
Add subruns information in `DataKey` of superruns to track metadata

dachengx:

The subruns of a superrun is not checked when loading a saved superrun after redefining the superrun. You will find the example in the MWE.

This PR does the following things:

1. Add an argument of `DataKey`: `subruns`, which is loaded from the metadata of superrun. Of course, if there is no metadata, an error will be raised.
2. Add `Context.get_datakey` for the context to get the decorated `DataKey`. Change all direct call of `strax.DataKey` in `Context` to `Context.get_datakey`, including `Context.key_for`.
3. Include the deterministic hash of `subruns` to the `run_id` in `DataKey`, so if everything is the same except `subruns`, the `DataKey.__repr__` will still change.

The motivation that we do not include `subruns` in `lineage` is that we would like to keep the lineage hash of same `data_type` of subruns and superrun the same, like `0-records-j3nd2fjbiq` and `_superrun_test_uzafu3c2wk-records-j3nd2fjbiq`.

```
#!/usr/bin/env python
# coding: utf-8

import json
import datetime
import pytz
from bson import json_util

import numpy as np

import strax
from strax.testutils import Records, Peaks
import straxen


superrun_name = "_superrun_test"

# Prepare for a context
st = strax.Context(
    storage=[
        strax.DataDirectory(
            "./strax_data", provide_run_metadata=True, readonly=False, deep_scan=True
        )
    ],
    register=[Records, Peaks],
    config={"bonus_area": 42},
)
st.set_context_config({"use_per_run_defaults": False, "write_superruns": True})


def _write_run_doc(context, run_id, time, endtime):
    """Function which writes a dummy run document."""
    run_doc = {"name": run_id, "start": time, "end": endtime}
    with open(context.storage[0]._run_meta_path(str(run_id)), "w") as fp:
        json.dump(run_doc, fp, sort_keys=True, indent=4, default=json_util.default)


# Define run metadata of each subrun
offset_between_subruns = 10
now = datetime.datetime.now()
now.replace(tzinfo=pytz.utc)
subrun_ids = [str(r) for r in range(3)]
for run_id in subrun_ids:
    rr = st.get_array(run_id, "records")
    time = np.min(rr["time"])
    endtime = np.max(strax.endtime(rr))

    _write_run_doc(
        st,
        run_id,
        now + datetime.timedelta(0, int(time)),
        now + datetime.timedelta(0, int(endtime)),
    )

    st.set_config({"secret_time_offset": endtime + offset_between_subruns})  # untracked option
    assert st.is_stored(run_id, "records")

st.define_run(superrun_name, subrun_ids)
print(f"When subruns are: {list(st.run_metadata(superrun_name)['sub_run_spec'].keys())}:")
print(f"Old DataKey is {st.key_for(superrun_name, 'records')}")
st.make(superrun_name, "records")
assert st.is_stored(superrun_name, "records")

# But if you change the metadata definition...
st.define_run(superrun_name, subrun_ids[:-1])
print(f"When subruns are: {list(st.run_metadata(superrun_name)['sub_run_spec'].keys())}:")
print(f"New DataKey is {st.key_for(superrun_name, 'records')}")
assert not st.is_stored(superrun_name, "records")
```

The output will be

```
When subruns are: ['0', '1', '2']:
Old DataKey is _superrun_test_uzafu3c2wk-records-j3nd2fjbiq
When subruns are: ['0', '1']:
New DataKey is _superrun_test_zc44ingvpo-records-j3nd2fjbiq
```

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1069,9 +1069,9 @@ class Context:
                 key, chunk_number=chunk_number, time_range=time_range
             )
 
-            _is_temp = not target_plugin.provides[0].startswith("_temp")
-            _is_superrun = run_id.startswith("_") and _is_temp
-            _is_hyperrun = run_id.startswith("__") and _is_temp
+            _is_not_temp = not target_plugin.provides[0].startswith("_temp")
+            _is_superrun = run_id.startswith("_") and _is_not_temp
+            _is_hyperrun = run_id.startswith("__") and _is_not_temp
             # hyperrun is always superrun
             allow_hyperrun = plugins[target_i].allow_hyperrun
             if not loader and (
@@ -1080,7 +1080,7 @@ class Context:
                 if time_range is not None:
                     raise NotImplementedError("time range loading not yet supported for superruns")
 
-                sub_run_spec = self.run_metadata(run_id, "sub_run_spec")["sub_run_spec"]
+                sub_run_spec = self.run_metadata(run_id, projection="sub_run_spec")["sub_run_spec"]
 
                 # Make subruns if they do not exist.
                 self.make(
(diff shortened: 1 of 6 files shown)


dachengx:
The error in the GitHub action will be killed after updating the test of straxen.

PR #867, opened 2024-08-15
Add option to save first samples of peak(lets) waveform

HenningSE:

This PR adds the option to save the first n (default to 200) samples of the merged peak(lets) waveform without downsampling. In case of very wide S1s or S1s with a lot of afterpulses strax will downsample the peak waveforms to fit into the 200 sample summed waveform that we save with peaklets and peaks. This process removes important information of the S1 waveform. With this PR the first n samples of the waveforms are saved for peaklets that 1. get downsampled and 2. have a downsampling factor smaller 7. This should cover basically all S1s without introducing a large file size overhead.


strax/processing/peak_building.py
```diff
@@ -170,6 +174,14 @@ def store_downsampled_waveform(
             wv_buffer[: p["length"] * downsample_factor].reshape(-1, downsample_factor).sum(axis=1)
         )
         p["dt"] *= downsample_factor
+
+        # If the waveform is downsampled, we can store the first samples of the waveform
+        if store_waveform_start & (downsample_factor <= 6):
```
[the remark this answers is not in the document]
HenningSE: Thanks @dachengx, you are right. Is https://github.com/AxFoundation/strax/pull/867/commits/7f4000ba091afe3ac41ba2a3ae4c571a1d261778 fine for you?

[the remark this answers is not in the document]
dachengx: Thanks. Would you make the `max_downsample_factor=6` an argument of `store_downsampled_waveform`?

strax/processing/peak_merging.py
```diff
@@ -85,7 +85,7 @@ def merge_peaks(peaks, start_merge_at, end_merge_at, max_buffer=int(1e5)):
 
         # Downsample the buffers into new_p['data'], new_p['data_top'],
         # and new_p['data_bot']
-        strax.store_downsampled_waveform(new_p, buffer, True, buffer_top)
+        strax.store_downsampled_waveform(new_p, buffer, True, True, 6, buffer_top)
```
dachengx, line 88: This 6 should be an argument as well. You should look for possible calls of `merge_peaks` and use `URLConfig` in these places.

dachengx (changes requested):
Thanks @HenningSE . There is one missing argument.

dachengx (approved):
Thanks @HenningSE . If you are sure this PR works well with your proposed change in straxen, please merge it after autotest finishes.

HenningSE:
Thanks @dachengx, it looks like I'm not an authorized user and can not merge.

strax/processing/peak_building.py
```diff
@@ -170,6 +180,14 @@ def store_downsampled_waveform(
...
         p["dt"] *= downsample_factor
+
+        # If the waveform is downsampled, we can store the first samples of the waveform
+        if store_waveform_start & (downsample_factor <= max_downsample_factor_waveform_start):
+            if p["length"] > len(p["data_start"]):
+                p["data_start"] = wv_buffer[: len(p["data_start"])]
+            else:
+                p["data_start"][: p["length"]] = wv_buffer[: p["length"]]
+
     else:
```
dachengx, line 191: Maybe here you should also assign `data_start`.

HenningSE: Hi @dachengx, I think this case is not needed as we do not downsample the waveform. The `data` field contains all needed information.

PR #895, opened 2024-10-02
S3 protocol

LuisSanchez25:

Adds S3 functionality to strax

We can now use an S3 storage system to store data

`import strax`
`st = strax.Context()`
`s3_storage = strax.S3Frontent()`
`st.storage = [s3_storage]`

  - Update the docstring(s)
  - Update the documentation


dachengx:
Thanks, @LuisSanchez25. Would it be more appropriate to put these new `StorageFrontend` and `StorageBackend` to straxen like `RucioRemoteFrontend` https://github.com/XENONnT/straxen/blob/a412b4382eed3277fd599c105d69855df38bc7f8/straxen/storage/rucio_remote.py#L29?

dachengx:
And I generally think this is a super good idea! We should benefit from the market resources!

LuisSanchez25:
Hey @dachengx I am a bit unfamiliar with why the decision was made to put the Rucio frontend in straxen rather than in strax, that feels like a tool that would be beneficial for others outside of XENON right? So to me it might make more sense to move the Rucio storage to straxen but maybe I am just missing something. 

Thanks! I am still in the process of fully testing this, I think it still needs some tweaks but I should have a working prototype soon!

dachengx:
@LuisSanchez25 I think by design, strax is for the prototypes of all classes, like plugins and storage. straxen will inherit the classes and make the functions more specific. I think this is why they put the `RucioRemoteFrontend` to straxen.

strax/io.py
```diff
@@ -73,21 +73,42 @@ def _load_file(f, compressor, dtype):
...
     """
+
     if isinstance(f, str):
         final_fn = f
         temp_fn = f + "_temp"
-        with open(temp_fn, mode="wb") as write_file:
-            result = _save_file(write_file, data, compressor)
-        os.rename(temp_fn, final_fn)
-        return result
+        if is_s3_path is False:
```
dachengx, line 88: `if is_s3_path is False:` -> `if not is_s3_path:`

dachengx (changes requested):
Thanks @LuisSanchez25 . I did not look into `s3.py` deeply because I see there are commented-out codes so maybe it is not finished?

I still would insist on moving the S3 functionality to straxen. strax should be just basic usage and processor etc.

You should not change `save_file` but do what should be done only in `_save_chunk` function.

WenzDaniel:
There was not any activity for quite some time and I see that there was a change request by @dachengx. So I will close this PR for now.

PR #898, opened 2024-10-12
Set single thread of zstd and blosc

dachengx:

Only https://github.com/XENONnT/outsource/pull/213 can not limit the usage of CPU because `zstd` and `blosc` by default will use as many as CPU on the machine.

This PR replaces `zstd` with `zstandard` and sets a one-thread constraint of `zstd` and `blosc`.

The reason for replacing `zstd` with `zstandard` is listed here: https://python-zstandard.readthedocs.io/en/latest/projectinfo.html#comparison-to-other-python-bindings and https://github.com/indygreg/python-zstandard/blob/e27f2f49f187f5f698e0a6004bee2aed80fe39e3/docs/projectinfo.rst?plain=1#L30.

I hope this will not make the previously compressed data unable to be decompressed.

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ pymongo = "*"
 scipy = "*"
 tqdm = ">=4.46.0"
 zarr = "*"
-zstd = "*"
+zstandard = "*"
 sharedarray = { url = "https://xenon.isi.edu/python/SharedArray-3.2.3.tar.gz", optional = true }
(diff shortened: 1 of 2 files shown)


dachengx:
Please @rsriya and @ghusheng remove zstd and add standard in https://github.com/XENONnT/base_environment/blob/36b01dbce38ca0362c52cc2f44c78a129d2ce759/extra_requirements/requirements-tests.txt.

PR #901, opened 2024-10-13
Do not add producer that has been added by saver

dachengx:

This is to solve the error when `processor = "single_thread"` in https://github.com/AxFoundation/strax/issues/903.

In the example of https://github.com/AxFoundation/strax/issues/903, after `st.make(nt_test_run_id, "peaklets", processor="single_thread")`, the `pulse_counts` is saved. So the loader will load this plugin because `peaklets` need `records` and `pulse_counts` are produced along with `records`. So `pulse_counts` is in `components.loaders`.

But before this PR, since `records` is also in `components.plugins`, https://github.com/AxFoundation/strax/blob/dca35457bea07b278e331b07fefb98cea842e60a/strax/processors/single_thread.py#L38 will register `pulse_counts` in the producer so this caused the bug.

This PR prevents the producer registration if a `data_type` is already in `components.loaders`.

Please check https://github.com/AxFoundation/strax/issues/903.

### DIFF
--- a/strax/context.py
+++ b/strax/context.py
@@ -1006,7 +1006,7 @@ class Context:
         """Get the datatype that is provided by a plugin but not depended on by any other plugin."""
         provides = [prov for p in plugins.values() for prov in strax.to_str_tuple(p.provides)]
         depends_on = [dep for p in plugins.values() for dep in strax.to_str_tuple(p.depends_on)]
-        uniques = list(set(provides) ^ set(depends_on))
+        uniques = list(set(provides) - set(depends_on))
         return strax.to_str_tuple(uniques)
 
     @property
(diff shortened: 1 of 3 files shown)


strax/context.py
```diff
@@ -1006,7 +1006,7 @@ def _get_end_targets(plugins: dict) -> ty.Tuple[str]:
         """Get the datatype that is provided by a plugin but not depended on by any other plugin."""
         provides = [prov for p in plugins.values() for prov in strax.to_str_tuple(p.provides)]
         depends_on = [dep for p in plugins.values() for dep in strax.to_str_tuple(p.depends_on)]
-        uniques = list(set(provides) ^ set(depends_on))
+        uniques = list(set(provides) - set(depends_on))
```
dachengx, line 1009: Get rid of dependencies, do not keep used dependencies.

strax/processors/single_thread.py
```diff
@@ -35,9 +35,12 @@ def __init__(
                 continue
             plugins_seen.append(p)
 
+            # Some data_types might be already saved and can be loaded;
+            # remove them from the list of provides
             self.post_office.register_producer(
                 p.iter(iters={dep: self.post_office.get_iter(dep, d) for dep in p.depends_on}),
                 topic=strax.to_str_tuple(p.provides),
+                registered=tuple(components.loaders),
```
dachengx, line 43: This is the core of this PR: do not register `loaders`.

PR #904, opened 2024-10-13
Set `SingleThreadProcessor` as the default processor

dachengx:

After https://github.com/AxFoundation/strax/pull/773, we can set `SingleThreadProcessor` as the default one to reduce the complexity of processing.

  - Update the docstring(s)
  - Update the documentation


dachengx:
GREAT!

Issue #906, opened 2024-10-15
n_top_channels always set to 0


HenningSE:
In `add_lone_hits` the numba function `_add_lone_hits` is called with `n_top_channels` always set to 0. 
https://github.com/AxFoundation/strax/blob/88432479892850fb90c93240a53db0f7938ab604/strax/processing/peak_merging.py#L196
This line should be changed to 
```
_add_lone_hits(peaks, lone_hits, to_pe, n_top_channels=n_top_channels)
```
so that the value is properly passed.

PR #907, opened 2024-10-15
Propagate n_top_channels to _add_lone_hits

HenningSE:

Solve https://github.com/AxFoundation/strax/issues/906


dachengx (approved):
Thanks @HenningSE

PR #909, opened 2024-10-16
Save `data_start` even there is no downsampling

dachengx:

Further, fix https://github.com/AxFoundation/strax/pull/867#discussion_r1797770537. As said in that conversation, People might misunderstand the non-downsampled `data_start` if they are zero.

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/processing/peak_building.py
+++ b/strax/processing/peak_building.py
@@ -180,18 +180,17 @@ def store_downsampled_waveform(
             wv_buffer[: p["length"] * downsample_factor].reshape(-1, downsample_factor).sum(axis=1)
         )
         p["dt"] *= downsample_factor
-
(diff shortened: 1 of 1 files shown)


HenningSE:
Thanks, @dachengx we can do it like this! Do you still want to keep the upper limit for the downsampling factor? It was introduced to save disk space and keep data_start for S2s at 0. To avoid further confusion, we could also remove this limit if you want.

dachengx:
> Thanks, @dachengx we can do it like this! Do you still want to keep the upper limit for the downsampling factor? It was introduced to save disk space and keep data_start for S2s at 0. To avoid further confusion, we could also remove this limit if you want.

Maybe yes. I will update the PR.

PR #910, opened 2024-10-16
Remove `max_downsample_factor_waveform_start`, simplify the saving of `data_start`

dachengx:

`max_downsample_factor_waveform_start` is not strictly needed. We can use the following logic: if `store_in_data_start` is `False`, do not save the `data_start`.

Also, change `digitize_top` to `store_data_top`, `store_in_data_top` to `store_data_top` to unify the variable names.

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/dtypes.py
+++ b/strax/dtypes.py
@@ -166,7 +166,7 @@ def peak_dtype(
     n_widths=11,
     digitize_top=True,
     hits_timing=True,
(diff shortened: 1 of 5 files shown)


dachengx:
The incompatible to straxen, shown by the failure of autotests, is because of the change from `digitize_top ` to `store_data_top `.

Issue #916, opened 2024-10-19
Check if all `argsort` use `mergesort`


dachengx:
We must ensure that the `numpy.argsort` gives deterministic results, as well as in https://github.com/XENONnT/straxen.

This issue is closely related to https://github.com/XENONnT/straxen/pull/1176.

yuema137:
I checked the source code of `numpy` and confirmed that `argsort` is a more fundamental method while `sort` is a higher-level API based on it. The algorithms are defined in cpp files like https://github.com/numpy/numpy/blob/main/numpy/_core/src/npysort/mergesort.cpp

With the `mergesort` method, the relative order is preserved:
https://github.com/numpy/numpy/blob/020e78cc3a4c7283bbe83290f0be5de3bc842f07/numpy/_core/src/npysort/mergesort.cpp#L67

So as long as we use `mergesort` I think we are good here.

dachengx:
> So as long as we use `mergesort` I think we are good here.

Yes! So would you check whether we always use `mergesort`?

yuema137:
Solved in #918

PR #918, opened 2024-10-23
Add enforcement for `np.sort` and `np.argsort`

yuema137:
This PR enforces deterministic sorting behavior by implementing the `sort_enforcement` module. Specifically, it:

1. Makes `mergesort` the default sorting algorithm
2. Disables non-deterministic sorting algorithms (`quicksort` and `heapsort`) to prevent their usage
3. Wraps `np.sort` and `np.argsort` to enforce the behaviors mentioned above

The implementation is robust to import order - the enforcement takes effect whenever `strax` is imported, regardless of whether `numpy` is imported before or after. This ensures consistent sorting behavior across all strax operations, addressing the non-deterministic sorting issues reported in #916.

A unit test is also added accordingly.


strax/sort_enforcement.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def enable_safe_sorting():
+    """Patches all NumPy sorting methods to enforce mergesort.
+
+    Returns a function to restore original behavior if needed.
+
+    """
+    # Replace all sorting methods
+    np.sort = sort_wrapper
```
dachengx, line 67: 1. This line will enforce overwriting any usage of `np.sort` and `np.argsort` in any other packages. I do not know whether this is want we want.
2. I am not sure on this point, but is `=` essentially the assignment or just the reference? Similar comment to https://github.com/AxFoundation/strax/pull/918/files#diff-ef513ebbd93a519d9d3ec5d64f437701a1784d501a1b05f1a323e160db63c8b7R6. if it is just reference, we will have n implicit change on `original_sort`.

yuema137:
Unfortunately, the `kind` for `np.sort` is hard-coded as `quicksort` in `numba` now:
https://github.com/numba/numba/blob/0f363d1b2dd19f2aa1a8cec5f0a99c3dd95512f8/numba/np/arrayobj.py#L6524

For `np.argsort`, `mergesort` and `quicksort` are both supported though.

So we have two options: 

1. Make a PR to `numba` to enable `mergesort` for `np.sort` (need to check if it's easy)
2. Move all the sorting in `strax` outside of the `numba` decorators (need to check performance)

I'm checking the possibility for both ways.

strax/processing/statistics.py
```diff
@@ -1,13 +1,14 @@
...
 
 import strax
+from strax.sort_enforcement import stable_argsort, stable_sort
 
 export, __all__ = strax.exporter()
 
 
 @export
-@numba.njit(cache=True)
+@register_jitable
```
dachengx, line 11: Why the change here is needed?

strax/sort_enforcement.py
```diff
@@ -0,0 +1,46 @@
...
+    Returns:
+        Sorted array using mergesort algorithm
+
+    """
+    if kind != "mergesort":
+        raise SortingError(UNSTABLE_SORT_MESSAGE)
+    return np.sort(arr, kind="mergesort", **kwargs)
+
+
+@register_jitable
```
dachengx, line 32: Why `register_jitable` is needed?

tests/test_sort.py
```diff
@@ -0,0 +1,76 @@
...
+import warnings
+from strax.sort_enforcement import SortingError, stable_sort, stable_argsort
+
+
+class TestSortEnforcement(unittest.TestCase):
+    def setUp(self):
+        """Set up test fixtures before each test method."""
+        self.arr = np.array([3, 1, 4, 1, 5, 9, 2, 6])
+        # Store expected sorted array and indices for comparison
+        self.expected_sorted = np.array([1, 1, 2, 3, 4, 5, 6, 9])
```
dachengx, line 12: These two variables can be generated using `np.sort(*, kind='merge_sort')`.

tests/test_sort.py
```diff
@@ -0,0 +1,76 @@
...
+        self.expected_argsort = np.array([1, 3, 6, 0, 2, 4, 7, 5])
+
+    def test_explicit_stable_sort(self):
+        """Test explicit stable_sort function (should not warn)"""
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")  # Turn warnings into errors
+            sorted_arr = stable_sort(self.arr)
+            np.testing.assert_array_equal(sorted_arr, self.expected_sorted)
+
+    def test_explicit_stable_sort_argsort(self):
```
dachengx, line 22: What is `stable_sort_argsort`? Should be `stable_argsort`?

tests/test_sort.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def test_wrapped_quicksort_rejection(self):
+        """Test that quicksort and heapsort raise errors in wrapped functions."""
+        # Test stable_sort wrapper
+        with self.assertRaises(SortingError):
+            stable_sort(self.arr, kind="quicksort")
+        with self.assertRaises(SortingError):
+            stable_sort(self.arr, kind="heapsort")
+
+        # Test stable_sort_argsort wrapper
```
dachengx, line 37: Also here `stable_sort_argsort` -> `stable_argsort`?

tests/test_sort.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def test_sort_stability(self):
+        """Test that wrapped sorting is stable (stable_sort property)"""
+        # Create array with duplicate values
+        arr = np.array(
+            [(1, "a"), (2, "b"), (1, "c"), (2, "d")], dtype=[("num", int), ("letter", "U1")]
+        )
+        sorted_arr = stable_sort(arr, order="num")
+        # Check that relative order of equal elements is preserved
+        self.assertEqual(sorted_arr[0]["letter"], "a")
```
dachengx, line 69: I believe these can be written more cleverly in for loop.

dachengx (commented):
In general, I guess you can use `hypothesis` package so do not need to assign `self.arr` in the test by hand.

strax/sort_enforcement.py
yuema137: with `register_jitable`, the function would be called as a regular python function when not used in a `numba` decorated function, and will be compiled as `numba` function if it's called within a decorated function

strax/processing/statistics.py
yuema137: Same as below

tests/test_sort.py
yuema137: fixed

strax/sort_enforcement.py
dachengx: Why not use numba jit as default for `stable_argsort` and `stable_sort `?

strax/processing/statistics.py
```diff
@@ -42,89 +28,96 @@ def highest_density_region(data, fractions_desired, only_upper_part=False, _buff
...
+    # Last interval
+    interval = ind[g0:]
+    res[fi, 0, g_ind + 1] = interval[0]
+    res[fi, 1, g_ind + 1] = interval[-1] + 1
+    return fi + 1, res
+
+
+@export
+@register_jitable
+def highest_density_region(data, fractions_desired, only_upper_part=False, _buffer_size=10):
```
dachengx, line 63: Are we sure that the behavior of `highest_density_region` is not changed? It is used at https://github.com/XENONnT/cutax/blob/195b6922cf5dc27da23fb2896a252f557be80752/cutax/cut_efficiency.py#L194.

yuema137: Originally it was using the `numba` implemented `sort`, which uses `quicksort`. But now we avoid using the `numba` one, only call the pure Python function, and enforce `mergesort`. So I would suppose the behavior changes.

strax/sort_enforcement.py
yuema137: For `stable_sort` the only supported method is `quicksort`, so we need to avoid it. For `stable_argsort` it's ok to use the `numba` implemented `mergesort`.

strax/processing/statistics.py
```diff
@@ -41,90 +23,153 @@ def highest_density_region(data, fractions_desired, only_upper_part=False, _buff
...
-    # If we end up here this might be due to an offset
-    # of the distribution with respect to zero. In that case it can
-    # happen that we do not find all desired fractions.
-    # Hence we have to enforce to compute the last step from the last
-    # lowest hight we have seen to zero.
-    # Left and right edge is by definition 0 and len(data):
+            # Process intervals with numba
+            fi, res = _process_intervals_numba(ind, gaps, fi, res, 0, _buffer_size)
+
+            if fi == len(fractions_desired):
```
dachengx, line 166: Here you might have added 1 unnecessary indent.

dachengx (changes requested):

strax/processing/statistics.py
yuema137: i removed it

dachengx: `            if fi == len(fractions_desired):` -> `        if fi == len(fractions_desired):`

dachengx (changes requested):
https://github.com/AxFoundation/strax/pull/918/files#r1841529802

dachengx:
I should mention that this PR is following up https://github.com/XENONnT/straxen/pull/1176, for future reference.

dachengx (approved):

PR #920, opened 2024-11-05
Fixing store_data_start in recursive peak splitter calls and peaklet["length'] fix in store_downsampled_waveform

HenningSE:

This PR fixes some smaller problems related to the `data_start` handling in the `peaklets` building/ splitting. 

- The flag `store_data_start` was not passed to recursive calls of the peak splitter. As a result some `peaklets` had `data_start` always set to 0. 
- `peaklet["length']` is used in `store_downsampled_waveform` to select the part of the waveform that is stored in `data_start`. If the waveform is downsampled, `peaklet["length']` is overwritten. When saving  `data_start` we need to take the initial length of the peaklet not the downsampled length.

### DIFF
--- a/strax/processing/peak_building.py
+++ b/strax/processing/peak_building.py
@@ -190,7 +190,7 @@ def store_downsampled_waveform(
         if p_length > len(p["data_start"]):
             p["data_start"] = waveform_buffer[: len(p["data_start"])]
         else:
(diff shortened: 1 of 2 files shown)


dachengx (approved):

dachengx:
@HenningSE Please make the title of this PR more detailed.

PR #934, opened 2024-11-26
Add Safeguard for Superrun

MerzJohannes:
It adds a safeguard for the superrun. If the flag is enabbled than the config files of subruns are compared and if they differ an error is raised.
Get a list of all the configs that are taken by the plugins and iterate over them. If a plugin takes the config it's added to a list. Then a set of the list is taken and if there is more than one element in there an error is raised since two of the subruns are taking different configs.

  - Update the docstring(s)
  - Update the documentation


yuema137:
This idea sounds good to me. Thanks for the effort! I looked into it a little bit and found a critical issue that might make the implementation hard, and we probably need another PR regarding this, depending on whether the behavior is actually needed:
The `__get_plugin` function always calls the `_plugins_are_cached` function to load the plugin object:https://github.com/AxFoundation/strax/blob/574661b61004a954bc810bad864d378313203b6e/strax/context.py#L861-L909

And the plugins are cached without `run_id` information: https://github.com/AxFoundation/strax/blob/574661b61004a954bc810bad864d378313203b6e/strax/context.py#L742-L781

Although the `run_id` info is added later after the cache is loaded in `__get_requested_plugins_from_cache`, the `run_id` information is actually not propagated. That means no matter whatever plugin configs you set for the individual runs, they will never be updated as the program is always loading the original one.  

https://github.com/AxFoundation/strax/blob/574661b61004a954bc810bad864d378313203b6e/strax/context.py#L783-L814

That's probably why the `test_only_combining_subruns_with_same_config` could never work. I have two questions regarding this situation:

1. It seems that with the current setup, the plugins under different `run_id` share the same object (I have tested locally and they really do) - is that what we want?

2. If that is not a bug but already an enforcement for plugin consistency, do we still need this safeguard?

@dachengx It would be great to hear your opinions

MerzJohannes:
Thank you Sue, I also think it would be good to hear @dachengx  opinion

dachengx (changes requested):
I agree with @yuema137 .

The solution has two parts:

1. for corrections that are values, compare values using `straxen.URLConfig.fetch` or `straxen.URLConfig.evaluate_dry`.
2. for corrections that are files, compare the file name.

But we have to find a way to extract the file name, which requires understanding xedocs.

dachengx:
> for corrections that are files, compare the file name.

This requirement of functionality is correlated with CMT -> xedocs transition, so we can come back after CMT is completely removed.

dachengx:
Sorry, this topic is complex now and got replaced by https://github.com/XENONnT/straxen/pull/1546.

PR #937, opened 2024-12-17
Propagate min/max diff in merged S2s

WenzDaniel:
Propagate min/max diff into merged S2s. For max_diff this is just a proxy within the respective peaklets, but min diff is correct.

Tested with run 025423.


strax/processing/peak_merging.py
```diff
@@ -69,6 +70,14 @@ def merge_peaks(
             new_p["n_hits"] += p["n_hits"]
             new_p["saturated_channel"][p["saturated_channel"] == 1] = 1
 
+            # Propagate min/max diff for sub-peaklets, for max diff this
+            # is just an approximation since peaklets can be farther apart.
+            # However, the value of the individual peaklets might os more
+            # informative
+            # For min diff the value is correct.
+            new_p["max_diff"] = max(new_p["max_diff"], p["max_diff"])
+            new_p["min_diff"] = min(new_p["min_diff"], p["min_diff"])
```
dachengx, line 79: If a peaklet is split, its `min_diff` and `max_diff` will be -1: https://github.com/AxFoundation/strax/blob/956031e150c00f0ce207c8435d7ee016422cf840/strax/processing/peak_splitting.py#L222. So we should exclude `-1` when getting the `min_diff`? This is also why we did not add the fields, besides your comments above.

A typo maybe: `might os more`.

dachengx (changes requested):
Thank you @WenzDaniel !

strax/processing/peak_merging.py
dachengx: It is OK to not handle the case where they are -1.

PR #942, opened 2024-12-22
Delete input chunks after compute method to save memory

dachengx:
This will lower the memory usage of `st.make(run_id, 'peaklets')` of XENONnT by 1 / 6, if set `clean_chunk_after_compute = True`.

Note: https://xe1t-wiki.lngs.infn.it/doku.php?id=xenon:xenonnt:straxen:optimize_outsource_resources#optimized_memory_usage

### DIFF
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -206,9 +206,9 @@ class Chunk:
         """
         t = max(min(t, self.end), self.start)  # type: ignore
         if t == self.end:
-            data1, data2 = self.data, self.data[:0]
+            data1, data2 = self.data, self.data[:0].copy()
         elif t == self.start:
-            data1, data2 = self.data[:0], self.data
+            data1, data2 = self.data[:0].copy(), self.data
         else:
             data1, data2, t = split_array(data=self.data, t=t, allow_early_split=allow_early_split)
 
@@ -249,9 +249,6 @@ class Chunk:
             **{**common_kwargs, "run_id": run_id_second_chunk},
         )
         return c1, c2
-
-    def __del__(self):
-        del self.data
 
     @classmethod
     def merge(cls, chunks, data_type="<UNKNOWN>"):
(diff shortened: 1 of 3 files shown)


strax/chunk.py
```diff
@@ -206,9 +206,9 @@ def split(self, t: ty.Union[int, None], allow_early_split=False):
         """
         t = max(min(t, self.end), self.start)  # type: ignore
         if t == self.end:
-            data1, data2 = self.data, self.data[:0]
+            data1, data2 = self.data, self.data[:0].copy()
```
dachengx, line 209: This is to prevent the reference build between `data2` and `self.data` to make sure that the `sys.getrefcount(data)` is low enough.

strax/plugins/plugin.py
```diff
@@ -667,13 +668,27 @@ def do_compute(self, chunk_i=None, **kwargs):
...
-            kwargs["chunk_i"] = chunk_i
+            _kwargs["chunk_i"] = chunk_i
         if self.compute_takes_start_end:
-            kwargs["start"] = start
-            kwargs["end"] = end
-        result = self.compute(**kwargs)
+            _kwargs["start"] = start
+            _kwargs["end"] = end
+        result = self.compute(**_kwargs)
+        del _kwargs
```
dachengx, line 678: Delete `_kwargs` because it will keep a reference between `_kwargs` and `kwargs[k].data`.

strax/plugins/plugin.py
```diff
@@ -667,13 +668,27 @@ def do_compute(self, chunk_i=None, **kwargs):
...
+            keys = list(kwargs.keys())
+            for k in keys:
+                # Minus one accounts for reference created by sys.getrefcount itself
+                n = sys.getrefcount(kwargs[k].data) - 1
+                if n != 1:
+                    raise ValueError(
+                        f"Reference count of input {k} is {n} "
+                        "and should be 1. This is a memory leak."
+                    )
+                del kwargs[k].data
```
dachengx, line 691: You might see `AttributeError` when calling `kwargs[k]` which is a `strax.Chunk`.

PR #953, opened 2025-01-15
When loader is available, do not save anything

dachengx:
This is a bug fix for https://github.com/AxFoundation/strax/pull/871.

Do not allow `save` in `combining` mode.

### DIFF
--- a/strax/chunk.py
+++ b/strax/chunk.py
@@ -297,6 +297,7 @@ class Chunk:
             data_kind=data_kind,
             run_id=run_id,
             data=data,
+            subruns=_merge_subruns_in_chunk(chunks, merge=True),
             superrun=_merge_superrun_in_chunk(chunks, merge=True),
             target_size_mb=max([c.target_size_mb for c in chunks]),
         )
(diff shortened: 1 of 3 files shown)


strax/context.py
```diff
@@ -1248,8 +1248,16 @@ def concat_loader(*args, **kwargs):
                 for dep_d in target_plugin.depends_on:
                     check_cache(dep_d)
 
-            # In case we can load the data already we want make a new superrun.
-            if loader and not (is_superrun and self.context_config["write_superruns"]):
```
dachengx, line 1252: The most important bug is here. The exception of superrun is not needed. Because if the subruns are not available, the function will iteratively recursively call `check_cache`. So we do not have to worry about resaving superruns.

strax/chunk.py
```diff
@@ -219,7 +220,17 @@ def split(self, t: ty.Union[int, None], allow_early_split=False):
             target_size_mb=self.target_size_mb,
         )
 
-        subruns_first_chunk, subruns_second_chunk = _split_runs_in_chunk(self.subruns, t)
+        if self.is_superrun and (
```
dachengx, line 223: In normal cases, the time interval of superrun exactly covers the data. If it does not, it means we might be resetting the start and end time in metadata.

PR #957, opened 2025-01-17
Force memory allocation in `merge_arrs`

dachengx:
Otherwise, weirdly you will see

`peaklets = st.get_array(run_id, targets=('peaklets', 'peaklet_classification'))` but `peaklets['type']` is always 0 (from `"peaklets"`).

This error might happen at https://github.com/AxFoundation/strax/blob/4e13c7a55e033758071f9092dd83c57868fa36c3/strax/context.py#L1823.


dachengx:
Sorry, I have to tell you this PR is wrong. The true solution is at https://github.com/AxFoundation/strax/pull/958.

PR #990, opened 2025-03-19
Add arguments of `numpy.argsort`, like `order`

dachengx:
Like `stable_sort`.


WenzDaniel:
Too many tests fail make this a draft again.

PR #1000, opened 2025-08-14
Better performance lineage `chunk_number` assignment

dachengx:

```
import straxen  # e9fe5dd7fb61fb537d35732730b41639ebc39980
import saltax  # e100810ad106351ef1180499889257d59c97ed90

st = saltax.contexts.sxenonnt(
    saltax_mode="salt",
    generator_name="ybe",
    recoil=0,
    rate=10,
    simu_mode="all",
    simulation_config="sr2_dev",
    corrections_version="global_v18",
    output_folder="strax_data",
)

run_id = "047967"
st.is_stored(run_id, "peaklets", chunk_number={"raw_records": [0, 1, 2]})
```

First get the efficiency of the above codes:

```
python -m cProfile -o output.prof test_chunks.py
gprof2dot -f pstats output.prof | dot -Tsvg -o profile.svg
```

The most significant part is

image

This is because when `chunk_number` is not None, strax before this PR will not cache plugins, so every time building the lineage of a higher plugin, the `__get_plugin` will be called.

This PR makes sure that whenever `chunk_number` is None or not, the plugin (whose lineage does not include `chunk_number`) is cached. Then I designed a dedicated function to assign the `chunk_number` in lineage afterwards.

  - Update the docstring(s)
  - Update the documentation

### DIFF
--- a/strax/plugins/plugin.py
+++ b/strax/plugins/plugin.py
@@ -196,9 +196,6 @@ class Plugin:
                 plugin_copy.__setattr__(attribute, copy(source_value))
         return plugin_copy
 
-    def __deepcopy__(self):
-        return self.__copy__(_deep_copy=True)
-
     def __getattr__(self, name):
         """Allow access to config parameters as attributes this allows backwards compatibility in
         cases where a descriptor style config depends on a non descriptor style config."""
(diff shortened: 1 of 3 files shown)


strax/plugins/plugin.py
```diff
@@ -196,9 +196,6 @@ def __copy__(self, _deep_copy=False):
                 plugin_copy.__setattr__(attribute, copy(source_value))
         return plugin_copy
 
-    def __deepcopy__(self):
```
dachengx, line 199: This is essentially a bug. `__deepcopy__` never works because `deepcopy(plugin.deps)` will also call `__deepcopy__`. `plugin.deps` is a dictionary of plugins.

PR #1005, opened 2026-01-29
Implement time delay plugin

mhliu0001:

In simulation plugins (e.g., S2 simulations in `fuse`), operations like sampling electron drift time add variable delays to output timestamps. This causes output data to exceed input chunk boundaries, breaking strax's chunking assumptions. This PR adds `TimeDelayPlugin`, a base class that handles buffering and chunking for plugins with time-delayed output.

`TimeDelayPlugin` buffers output data and releases it when safe:
- Data with `endtime <= input_chunk.end` is output immediately
- Data extending beyond the boundary is buffered until subsequent chunks arrive
- Output is re-sorted by time (since variable delays may reorder data)
- Chunk boundaries are constrained to ensure buffered data fits in subsequent chunks
- Supports both single and multi-output plugins

Subclasses implement `compute_with_delay(**kwargs)` (returns delayed output arrays).

```python
import numpy as np
import strax
from strax.testutils import Records, run_id

class VariableDelayPlugin(strax.TimeDelayPlugin):
    """Plugin that adds random delays, simulating e.g. drift time."""
    depends_on = ("records",)
    provides = "delayed_records"
    data_kind = "delayed_records"
    dtype = strax.record_dtype()

    def compute_with_delay(self, records):
        result = records.copy()
        # Simulate variable drift time
        delays = np.random.randint(0, self.max_delay, size=len(result))
        result["time"] = result["time"] + delays
        return result

st = strax.Context(storage=[], register=[Records, VariableDelayPlugin])
result = st.get_array(run_id, "delayed_records")
```

Output times may exceed input chunk boundaries, and the plugin handles buffering and re-sorting automatically.

**Checklist:**
- [x] Docstrings included in `TimeDelayPlugin` class
- [x] Documentation updated: `docs/source/advanced/plugin_dev.rst`
- [x] Tests added: `tests/test_time_delay_plugin.py` (6 tests covering constant delay, variable delay, empty chunks, multi-output, chunk continuity, and straddling data)

### DIFF
--- a/docs/source/advanced/plugin_dev.rst
+++ b/docs/source/advanced/plugin_dev.rst
@@ -35,6 +35,7 @@ There are several plugin types:
    * ``CutPlugin``: Plugin type where using ``def cut_by(self, <datakind>)`` inside the plugin a user can return a boolean array that can be used to select data.
    * ``MergeOnlyPlugin``: This is for internal use and only merges two plugins into a new one. See as an example in straxen the ``EventInfo`` plugin where the following datatypes are merged ``'events', 'event_basics', 'event_positions', 'corrected_areas', 'energy_estimates'``.
    * ``ParallelSourcePlugin``: For internal use only to parallelize the processing of low level plugins. This can be activated using stating ``parallel = 'process'`` in a plugin.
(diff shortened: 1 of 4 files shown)


WenzDaniel:
As discussed with @mhliu0001, converting this to a draft for now.

